PR #26, opened 2025-06-11
Implement TargetIR intermediate representation for multi-backend support

steveklabnik:
## Summary
- Replace direct AST → x86-64 compilation with AST → TargetIR → x86-64 pipeline
- Add platform-independent instruction set with virtual registers for future backend targets
- Fix register corruption in recursive calls via stack-based register spilling

## Test Results
- ✅ All 44 unit tests passing across all crates
- ✅ All 4 integration tests passing (simple: 42, factorial: 120, while loop: 42, all samples compile)
- ✅ factorial(5) now correctly returns 120 instead of 1

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/rue-codegen/src/lib.rs
+++ b/crates/rue-codegen/src/lib.rs
@@ -2,53 +2,121 @@ use rue_ast::{CstRoot, ExpressionNode, FunctionNode, StatementNode};
 use rue_semantic::Scope;
 use std::collections::HashMap;
 
+mod regalloc;
+pub use regalloc::RegisterAllocator;
+
 #[derive(Debug, Clone, PartialEq)]
 pub struct CodegenError {
     pub message: String,
 }
 
-// x86-64 instruction representation
+/// Virtual register - will be allocated to a physical register or stack slot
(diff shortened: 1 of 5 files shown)


steveklabnik:
Note: this took about four hours in total. Claude's time estimates of the steps was that it was going to take more like 10. A lot of time was spent debugging an issue with recursive calls as part of a multiplication... but it did figure it out in the end.

crates/rue-codegen/src/lib.rs
```diff
@@ -570,296 +742,776 @@ impl Assembler {
...
+            Instruction::Pop { dest } => {
+                regalloc.allocate(*dest);
+            }
+            // TODO: Handle other instruction types
+            _ => {}
         }
     }
 
-    fn emit_instruction(&mut self, instr: &Instruction) -> Result<(), CodegenError> {
+    // Convert instructions to machine code (old method, keep for compatibility)
```
steveklabnik: will want to remove this, we don't need compatibility

PR #54, opened 2025-07-21
Implement comprehensive error handling improvements

steveklabnik:
1. Implementing Display and Error traits for all error types (CodegenError,
  LexError, ParseError, SemanticError)
2. Adding error formatting with source context in the lexer
3. Implementing parser error recovery with a recover_missing_semicolon function
4. Adding semicolon recovery to all statement types that need semicolons

### DIFF
--- a/crates/rue-parser/src/lib.rs
+++ b/crates/rue-parser/src/lib.rs
@@ -20,17 +20,68 @@ impl ParseError {
     }
 }
 
+impl std::error::Error for ParseError {}
+
+impl std::fmt::Display for ParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.message)
+    }
+}
+
 impl Parser {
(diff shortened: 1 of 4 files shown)


crates/rue-parser/src/lib.rs
```diff
@@ -20,17 +20,69 @@ impl ParseError {
...
 
         while !self.is_at_end() {
-            items.push(self.parse_item()?);
+            match self.parse_item() {
+                Ok(item) => items.push(item),
+                Err(e) => {
+                    // For now, return the first error immediately
+                    // In a full implementation, we'd collect errors and continue
+                    return Err(e);
+                    // TODO: Implement full error recovery
```
steveklabnik, line 81: I am okay with leaving this for the future.

steveklabnik:
On second thought, I only want to do this as part of greater work, right now, this makes semicolons effectively optional, which isn't what I want, I want to treat that as an error but continue going to see other errors, and so I need to do that before I'd consider merging this.

PR #71, opened 2025-07-22
Fix function argument ordering bug in codegen

steveklabnik:
When preparing arguments for function calls, arguments were being placed in  registers in reversed order due to the register allocator's preference order combined with how arguments were being collected.

The fix properly tracks where arguments currently are and moves them to  their correct ABI positions, handling potential register conflicts using R15  as a temporary register when needed.

This ensures function arguments are always passed according to the System V  ABI calling convention (first arg in RDI, second in RSI, etc).

### DIFF
--- a/crates/rue-codegen/src/lowering.rs
+++ b/crates/rue-codegen/src/lowering.rs
@@ -758,16 +758,80 @@ impl<'a> Lowering<'a> {
             }
         }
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
I have a test for this coming in a future PR, I manually verified that this is fixed with the test case from then, but I didn't want to backport it.

PR #86, opened 2025-08-01
Enhance MIR design and refactor to use helper methods

steveklabnik:
Major improvements to the Mid-level Intermediate Representation (MIR):

- Add source spans to all MirTerminator variants for precise error reporting
- Add Switch and Unreachable terminator variants for advanced control flow
- Add CallKind enum to track function purity (Pure vs Impure) for optimization
- Add Copy trait to MirConst for better performance
- Remove redundant return_type field from MirValue::Call

Refactor HIR to MIR lowering to use cleaner helper methods:
- Use emit_const, emit_binop, emit_unop throughout for less boilerplate
- Consolidate temporary creation and assignment patterns
- All unit constant assignments now use emit_const helper

Fix MIR validation to handle the new Call structure:
- Add special handling for Call assignments in verifier
- Add function signature inference for common functions
- Fix input() function return type mapping

This also re-writes register allocation to be a simple "spill all" allocator. I want correctness, not performance, at this stage of the game.

We also moved some crates around: we now have rue-lowering and rue-optimize. This separates out some stuff that was in rue-ir or rue-codegen that isn't really about either of them directly.

Because I am bad at commit discipline, this *also* redoes our buck integration; I wanted to include some dev-dependencies, which I hadn't done before, and so this is best done by some stuff that I got from @adamhjk and the SI repo, so thank you for that, Adam!


crates/rue-ir/src/mir_verifier.rs
```diff
@@ -353,14 +442,58 @@ impl MirVerifier {
...
+                    MirTerminator::Return { .. } | MirTerminator::Unreachable { .. } => {}
                 }
             }
         }
 
         reachable
     }
 
+    /// Infer the return type of a function call based on its name
+    /// This is a temporary solution until we have proper function signature tracking
```
steveklabnik, line 466: this is very dumb, let's eliminate it later https://github.com/steveklabnik/rue/issues/88

PR #154, opened 2025-08-19
Phase 5: Create standalone rue-runtime crate & buffered/io

steveklabnik:
- Move buffered I/O to separate no_std crate as originally planned
- Build as static library (.a) with panic=abort profile
- Provide C ABI functions for assembly interop
- Add example assembly programs showing integration
- Successfully links and runs with assembly runtime


crates/rue-runtime/src/buffered_io.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+impl BufferedStdout {
+    /// Get a reference to the global buffered stdout instance
+    ///
+    /// # Safety
+    /// This function is unsafe because it provides mutable access to global state.
+    /// The caller must ensure this is only called from a single thread or properly
+    /// synchronized.
+    unsafe fn get_instance() -> Self {
```
steveklabnik, line 42: i don't have threads and won't for a long time so i'm just doing this this way for now

PR #169, opened 2025-10-26
docs: update outdated struct layout comment in tests

steveklabnik:
The comment claimed struct layout used heuristics, but this was outdated. The proper TypeContext-based layout calculation infrastructure was implemented on the same day (Aug 6, 2025) and is fully functional.

The infrastructure includes:
- TypeContext::compute_layout() for size/alignment calculation
- StructDef::compute_layout() for field offset computation
- Proper integration through scope_to_type_context()

This was just a stale comment from when the work was in progress.

### DIFF
--- a/crates/rue/tests/aggregate_runtime_integration_tests.rs
+++ b/crates/rue/tests/aggregate_runtime_integration_tests.rs
@@ -27,8 +27,7 @@
 //! - ✅ Memory allocation strategies
 //!
 //! ### Known Limitations:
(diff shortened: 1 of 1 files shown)


steveklabnik:
@claude can you update this to not explain the thing you just removed? This should simply say “none” and not describe things that are working already.

steveklabnik:
Seems like Claude can’t do that, at least how it’s configured right now.

PR #303, opened 2025-12-23
Fix /plan workflow: iterate on ADR before creating beads

steveklabnik:
The /plan skill was creating bd issues too early in the process. The correct workflow is:

1. Ask clarifying questions
2. Draft the ADR
3. Iterate with user until they approve
4. Only then create the epic + subtask beads

This ensures we don't create tracking issues for plans that may change significantly during review.


steveklabnik:
My first feature attempt with the new infrastructure ran into this issue, I think this is better.

PR #343, opened 2025-12-24
ignore beads directory

steveklabnik:
moving to a beads-metadata branch


steveklabnik:
https://github.com/rue-language/rue/tree/beads-metadata

PR #575, opened 2025-12-28
Expand allocatable register set to include caller-saved registers

steveklabnik:
This change improves register allocation by adding caller-saved registers to the allocatable set for both x86_64 and aarch64 backends:

- x86_64: Added RCX, RSI, RDI, R8, R9, R11 (6 caller-saved) to existing 5 callee-saved registers (RBX, R12-R15), for 11 total
- aarch64: Added X13, X14, X15 (3 caller-saved) to existing 10 callee-saved registers (X19-X28), for 13 total

The allocator now uses liveness clobber information to avoid allocating a vreg to a register that would be destroyed during its live range (e.g., by function calls). Caller-saved registers are listed first in the allocation order, so leaf functions avoid callee-save overhead entirely.

Key changes:
- linear_scan_impl: Added is_clobbered_during() check when finding free regs
- x86_64/mir.rs: Added is_callee_saved() method to Reg enum
- Both backends: Updated assign_registers() to filter used_callee_saved
- Tests: Made robust using ALLOCATABLE_REGS.len() instead of magic numbers

### DIFF
--- a/crates/rue-codegen/src/aarch64/regalloc.rs
+++ b/crates/rue-codegen/src/aarch64/regalloc.rs
@@ -13,19 +13,30 @@ use crate::regalloc::{Allocation, RegAllocDebugInfo, linear_scan, linear_scan_wi
 
 /// Available registers for allocation.
 ///
-/// We use callee-saved registers (X19-X28) for general allocation.
-/// This ensures values survive across function calls.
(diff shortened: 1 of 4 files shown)


steveklabnik:
It's weirding me out that these two runners are just dying every time, let's not do this.

PR #751, opened 2026-01-03
Fix typo in `@handle` spec

### LINKED ISSUE #750: Clarify `@handle` spec rule 3.8:45
https://github.com/rue-language/rue/blob/7c6652a813a09ca25139b80047778bcab91f7f4e/docs/spec/src/03-types/08-move-semantics.md?plain=1#L226

Should this read "does not consume the receiver", since the original remains valid after the call, or am I misunderstanding something?

DorianListens:
Fixes #750


steveklabnik:
Thank you! 🙏

PR #880, opened 2026-01-17
Optimize VariableMoveState, update ADR-0029, add examples

seth-schultz:
## Summary

- **Optimize**: Convert `VariableMoveState.partial_moves` from `HashMap` to `Vec` per ADR-0032
  - Single-pass prefix checking in `is_path_moved` (O(n) vs O(n×depth))
  - 3-6x faster lookups for typical workloads (0-5 partial moves)
- **Update ADR-0029**: Mark anonymous struct methods as implemented
- **Add examples**: `prime_factors.rue` and `vector2d.rue`

## Changes

| Commit | Description |
|--------|-------------|
| 1 | HashMap→Vec optimization with single-pass algorithm |
| 2 | ADR-0029 status: proposal → implemented |
| 3 | Two working example programs |

## Testing

- All 1341 spec tests pass
- All unit tests pass
- Examples compile and run correctly

### DIFF
--- a/crates/rue-air/src/sema/context.rs
+++ b/crates/rue-air/src/sema/context.rs
@@ -39,23 +39,23 @@ pub(crate) type FieldPath = Vec<Spur>;
 pub(crate) struct VariableMoveState {
     /// If Some, the entire variable has been fully moved at this span.
     pub full_move: Option<Span>,
-    /// Partial moves: maps field paths to the span where they were moved.
+    /// Partial moves: field paths paired with the span where they were moved.
     /// For example, if `s.a` was moved, this contains ([sym("a")], span).
-    pub partial_moves: HashMap<FieldPath, Span>,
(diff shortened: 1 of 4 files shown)


steveklabnik:
Thanks for this — closing as stale (5 months on it's now CONFLICTING), but the good parts are being salvaged rather than lost. It bundled three unrelated changes that have since diverged: (1) the `VariableMoveState` HashMap→Vec optimization is sound and still applies — re-landing it fresh (tracked as RUE-124), since it matches the ADR-0032 Vec precedent already on trunk; (2) the ADR-0029 status edit unfortunately regresses trunk, which already documents that feature as implemented-and-stabilized more accurately, so we're dropping that hunk; (3) `prime_factors.rue` is great and comes along with the re-land, but `vector2d.rue` no longer compiles — it tripped a real bug (`Self` in inline-method return position → E0204), now filed as RUE-123. Net: closing here, re-landing the optimization + working example as a focused PR.

PR #882, opened 2026-01-17
Improve hello.rue example and remove linker debug output

seth-schultz:
## Summary

Two quality-of-life improvements:

### 1. Update hello.rue to print `Hello, World!`

The classic first program should actually print something! Updated from just returning `42` to using `@dbg` to print the greeting.

### 2. Remove debug print statements from linker

The linker was printing 200+ `DEBUG:` lines during normal compilation. Removed all unconditional debug output from:
- `crates/rue-linker/src/linker.rs`
- `crates/rue-linker/src/macho.rs`

Also converted a sanity check from `eprintln!` to `debug_assert_eq!` for the text_file_offset verification.

## Testing

- `./quick-test.sh` passes
- `./buck2 run //crates/rue-spec:rue-spec -- "arithmetic"` passes (70 tests)
- Verified `hello.rue` compiles and runs correctly
- Verified no more DEBUG output during compilation

### DIFF
--- a/crates/rue-linker/src/linker.rs
+++ b/crates/rue-linker/src/linker.rs
@@ -417,16 +417,9 @@ impl Linker {
         // This keeps rodata addresses close to code for PC-relative addressing
         for (obj_idx, obj) in self.objects.iter().enumerate() {
             for (sec_idx, section) in obj.sections.iter().enumerate() {
-                eprintln!("DEBUG: Checking section '{}' for rodata", section.name);
                 if !is_rodata_section(&section.name) {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


steveklabnik:
Thanks for this, and apologies it sat so long. Closing as stale: the linker debug-output removal (the substantive part) has since been done on trunk independently — there are no `DEBUG`/`eprintln!` lines left in linker.rs or macho.rs — so this part is superseded and now conflicts. The only remaining bit is the nice `hello.rue` 'print Hello, World!' tweak, which we'll re-land as a one-liner. Closing in favor of that.

PR #881, opened 2026-01-17
Add search to specification

seth-schultz:
## Summary

Implements client-side search for the language specification (fixes #774).

## Changes

- **config.toml**: Enable Zola's built-in `build_search_index` and configure search settings
- **search.js**: Add elasticlunr-based client-side search with debounced input handling
- **spec/base.html**: Add search input to the specification sidebar with keyboard shortcut hint
- **base.html**: Load elasticlunr CDN and search.js scripts
- **input.css**: Add search UI styles with light/dark mode support

## Features

- Full-text search across all spec pages
- Search results show page titles and body snippets
- Keyboard shortcut: **⌘K** (Mac) / **Ctrl+K** (Windows/Linux) to focus search
- Shortcut hint displayed in the search input
- Dark mode support

## How It Works

1. Zola generates a `search_index.en.json` at build time containing all page content
2. elasticlunr loads this index and provides fast client-side full-text search
3. Results show in a dropdown with page titles and body snippets
4. Clicking a result navigates to that spec page

## Testing

Built locally and verified:
- Search works for queries like "integer", "overflow", "struct", etc.
- Keyboard shortcut ⌘K focuses the search input
- Shortcut hint displays correctly and hides when focused

PR #884, opened 2026-01-17
Implement lazy module loading for nested module access

seth-schultz:
## Summary

Implements lazy module loading to fix nested module member access (e.g., `std.math.abs()`).

## Problem

ADR-0026 claimed nested module access worked, but it didn't. Accessing `std.math.abs()` would fail because module exports weren't being loaded on demand.

## Solution

- Add `ModuleDef` export tracking (functions, structs, enums, constants with visibility)
- Create `module_loader.rs` for lazy scanning of module exports when first accessed
- Extract `check_visibility()` helper to reduce code duplication in `analyze_ops.rs`
- Clean up `types.rs` docstrings

## Testing

- All 1341 spec tests pass
- All 445 unit tests pass
- Nested module access now works correctly


steveklabnik:
Thanks for tackling this — the underlying bug is real (I reproduced `std.math.abs()` → E0707 on current trunk; ADR-0026 does overstate nested-module support), and it's now tracked as RUE-122. Closing this PR, though, because it isn't mergeable as-is: (1) it bundles all six `website/` files from PR #881 (merging this would silently land #881's spec-search and conflict with it); (2) it adds unrelated anon-struct preview-gating; and (3) the core `module_loader.rs` resolves module exports by line-by-line text-scraping of source (`strip_prefix("pub ")`, `find("//")`) rather than using the existing lexer/parser/AST, which will drift from the grammar (block comments, strings containing `fn `, attributes, generics). The visibility-tracking `ModuleExport`/`check_visibility` design and the tests here are worth salvaging into an AST-sourced reimplementation — captured in RUE-122. Closing in favor of that.

PR #994, opened 2026-07-01
Fix benchmarks: valid YAML, per-commit runs, actionlint gate

### LINKED ISSUE #991: Benchmarks are busted
image

Separately: benchmarking "every 15 minutes" seems unnecessary. We should probably trigger these per-commit, rather than on a schedule.

steveklabnik:
Fixes #991

## What was broken

`benchmarks.yml` has been **invalid YAML since 2026-01-12** — commit `9db23d8e` ("Fix second shell syntax error in benchmarks workflow") de-indented a Python heredoc body to column 0. That's fine in bash, but it terminates a YAML `run: |` block scalar, so GitHub couldn't parse the file at all. Every run since — including all ~100 during the June sprint — failed in 0 seconds at trigger time ("workflow file issue"), and the performance dashboard has been serving January data. Nothing in `./test.sh` can see workflow-file validity, which is why it stayed broken for five months.

## The fix (and the simplification requested in #991)

- **Per-commit benchmarks.** The workflow now runs once per commit pushed to trunk (plus `workflow_dispatch`), with a concurrency group that coalesces rapid merges to the newest commit. The 15-minute cron, the queue-depth checker with adaptive sampling, and the time-based batching / commit-range machinery are all deleted (~330 lines → ~200, most of it now plain steps). A data point now means exactly one commit.
- **No more inline Python in YAML** — the bug class that caused this. Result validation moved to `scripts/validate-benchmark.py`; the v2-schema enhancement moved into `scripts/append-benchmark.py --reason`.
- **Latent collector bug fixed:** platform extraction used a greedy `sed` that mapped every platform to its final hyphen-segment (`x86-64-linux` → `linux`), which would have collided all three platforms into one history file the first time the collector ever ran.
- **CI now runs actionlint**, so an unparseable workflow file can never merge again. Verified it flags the exact line-334 breakage on current trunk.
- **bench.sh** invokes the compiler with `-o` instead of the legacy two-positional output form (RUE-130 territory).
- **Performance page:** the Benchmark Coverage section and batching methodology (which only existed to explain the 15-minute cron) are removed; the benchmark-suite list now matches the real 7-program manifest.

## Verification

All by execution, locally: `bench.sh` end-to-end on a release build (7/7 benchmarks, both before and after the `-o` change), `append-benchmark.py --reason push` + `generate-charts.py` on the fresh results (all SVGs + metadata generate), full `website/build.sh`, and actionlint clean on all four workflows.

After this merges, the first trunk push will repopulate the perf branch and the dashboard redeploys automatically via the existing `workflow_run` trigger.


steveklabnik:
Added a second commit: the first CI round exposed that `facebook/install-dotslash@latest` broke macOS runners on 2026-06-18 — the `latest` branch gained release-digest verification whose `sha256sum` invocation doesn't work there, so every macOS job in this repo has failed at setup since (this would have blocked any PR, not just this one). All four workflows now pin `@v2`, which is functionally identical to what was green here in June.

Also noting #993, which correctly removed the 15-minute schedule but didn't catch that the workflow file itself was unparseable — that's why the failures persisted after it was closed.

PR #1060, opened 2026-07-03
Release-mode CI + promote correctness debug_assert! to assert! (RUE-45)

steveklabnik:
Per Steve's ruling: **test both modes**, and **`debug_assert!` is not for correctness** (RUE-24 precedent — an aggregate miscompile masked by a `debug_assert!` that vanishes in release turned an ICE into silent wrong codegen).

**Part A — test both modes:** new CI job `test (linux-x64, release)` runs the full suite against a `-Copt-level=3` compiler, via a `RUE_BUCK_MODIFIERS` env var threaded through `test.sh` (debug path unchanged). actionlint clean.

**Part B — debug_assert audit:** promoted every correctness-guarding `debug_assert!`/`debug_assert_eq!` in the codegen path to always-on `assert!` across **both** backends — cfg_lower slot-count checks (incl. the RUE-24-class one), aarch64 emit encoding guards, CFG `set_terminator`, AIR u32-overflow; `verify_stack_alignment` now runs in every build; the `specialize` `debug_assert!(false)` is now an unconditional `panic!`. Only genuine non-correctness dev checks remain `debug_assert!` (documented).

**Part C** (opt-level `-O0/-O2/-O3` cross-check) deferred → **RUE-236**: the `opt_level` Case field is gone and `-O2/-O3` alias `-O1` today.

Verified: `./test.sh` green in **both** debug and release (1677 unit + 21 suite each); release build compiles + runs the RUE-24-class nested-aggregate program correctly (exit 42, no spurious assert); actionlint + fmt clean.

Part of RUE-45

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,3 +109,42 @@ jobs:
 
       - name: Run tests
         run: ./test.sh
+
+  # Release-mode build + test (RUE-45). CI otherwise builds/tests the compiler
+  # in debug ONLY, so release-only miscompiles never run — a correctness check
+  # written as `debug_assert!` (or `#[cfg(debug_assertions)]`) is compiled out
+  # in release and silently turns an ICE into wrong codegen (RUE-24 is the
+  # precedent). This job exercises the whole suite against a `-Copt-level=3`
+  # build of the compiler. linux-x64 only: the aim is to catch opt-level and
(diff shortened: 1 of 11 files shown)


steveklabnik:
Superseded: stuck on clippy (+ macOS) failure, 38 PRs behind trunk. RUE-45 (release-mode CI + debug_assert->assert correctness audit) is being re-implemented fresh on current trunk, clippy-clean, in a new PR.

PR #1068, opened 2026-07-03
Generic types usable in function signatures via a type constant — real error handling (RUE-241, part of RUE-6)

steveklabnik:
**Named form:** a const bound to a type value (const R: type = Result(i32,i32);, a primitive alias const Int = i32;, etc.) is now a TYPE CONSTANT usable in function parameter and return positions. Was E0434/E0204/E0421. Sema-only: a type_constants table fed by the existing type-alias-init path; resolve_type consults it; each function's comptime_type_vars is seeded with file-level type constants so R.Ok(..)/patterns resolve (locals shadow).

**This unblocks real error handling end-to-end** — with RUE-237's aggregate-return fix now merged, a function returning a generic Result by value works:

    fn divide(a, b) -> R { if b == 0 { R::Err(1) } else { R::Ok(a / b) } }
    match divide(84, 2) { R::Ok(v) => v, R::Err(e) => 0 - e }   // 42

(Removed the known_bug=RUE-237 marker on the return case — RUE-237 landed, so it now passes.)

Verified by execution: generic Option/Result named via a type const, constructed, passed as a PARAMETER, RETURNED by value, and matched give 42; primitive alias param+return give 42; shadowing; non-type annotation is a type error. Full test.sh green; aarch64 asm clean.

Remaining (follow-ups): the DIRECT call form fn f() -> Result(i32,i32) is still parser E0100 (type-expr grammar); type consts in signatures must be declared before use.

Part of RUE-241

Generated with Claude Code

### DIFF
--- a/crates/rue-air/src/sema/analysis.rs
+++ b/crates/rue-air/src/sema/analysis.rs
@@ -1720,10 +1720,21 @@ impl<'a> Sema<'a> {
             value_subst,
         )?;
 
-        // Create analysis context with resolved types
-        // If type_subst is provided, initialize comptime_type_vars with the substitutions
-        // so that type parameters can be resolved during struct initialization.
-        let comptime_type_vars = type_subst.map(|s| s.clone()).unwrap_or_else(HashMap::new);
+        // Create analysis context with resolved types.
+        // Base the comptime type environment on file-level type constants
+        // (`const R: type = Result(i32, i32);`, RUE-241) so that enum/struct
+        // paths and match patterns that name `R` resolve to its concrete type,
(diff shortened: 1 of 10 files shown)


steveklabnik:
Closing: stuck on clippy failure. RUE-241 is already Done via the DIRECT form (#1083, fn f() -> Result(i32,i32)). The NAMED form here (const R: type = Result(...); fn f() -> R, currently E0434) is a minor follow-up gap; the diff (cycle59) is preserved and can be re-landed clippy-clean if the named binding is wanted separately.

PR #1070, opened 2026-07-03
CFG verifier pass + opt-level differential harness (RUE-227, RUE-236)

steveklabnik:
Two robustness/CI wins.

**RUE-227** — a CFG verifier (crates/rue-cfg/src/verify.rs): Cfg::verify() checks every reachable block for a real terminator, all referenced CfgValues defined, and matching block-param arities at each edge; violations panic!/assert! (NOT debug_assert, per RUE-45 — runs in release) with function name + block id. Wired at the end of opt::optimize (the choke point all build sites pass through before codegen). Turns the recurring block-has-no-terminator SIGABRT class (RUE-213/217/224) into a loud, localized compiler-bug report. Part B: lower_value now distinguishes true divergence (None) from continues-but-no-value (materializes a real Unit), generalizing the RUE-224 fix to the whole no-value class.

**RUE-236** — opt-level differential harness: a differential_opt marker runs a CLI-test case across -O0/-O1/-O2/-O3 and asserts identical exit + stdout, naming the diverging level on mismatch. Marks a bounded 4-case subset (arithmetic, control-flow, by-value aggregate, call/recursion). -O2/-O3 alias -O1 today so results match; the net is set so a future opt divergence is caught. Test-infra only.

Verified: verifier fires in debug AND release (59 rue-cfg tests pass in both; deliberately-broken-CFG tests panic as expected); opt divergence detection confirmed (mangled -O3 gives a clear failure, reverted). Full test.sh green.

Fixes RUE-227
Fixes RUE-236

Generated with Claude Code

### DIFF
--- a/crates/rue-cfg/src/build.rs
+++ b/crates/rue-cfg/src/build.rs
@@ -2387,15 +2387,35 @@ impl<'a> CfgBuilder<'a> {
         self.value_cache[air_ref.as_u32() as usize] = Some(value);
     }
 
-    /// Lower an instruction and return its value, or None if it diverged.
-    /// This is a helper for use with the `?` operator when processing operands.
-    /// If the operand diverged, the caller should propagate the divergence.
+    /// Lower an instruction used in **value position** and return its value, or
+    /// `None` **only** if it diverged (return/break/continue).
+    ///
+    /// Callers use the `let Some(v) = self.lower_value(x) else { return
+    /// Self::diverged(); }` idiom, so `None` here means "control did not reach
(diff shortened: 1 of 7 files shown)


steveklabnik:
Superseded: this stuck on a clippy failure (clippy.sh wasn't run before push) and is 38 PRs behind trunk. RUE-227 (CFG verifier) + RUE-236 (opt-level differential harness) are being re-implemented fresh on current trunk, clippy-clean + test.sh green, in a new PR.

PR #1089, opened 2026-07-03
CFG verifier pass + opt-level differential harness (RUE-227, RUE-236)

steveklabnik:
Re-landed clippy-clean (the original PR #1070 stuck on an unrun clippy gate). Two robustness/CI wins.

**RUE-227:** crates/rue-cfg/src/verify.rs adds Cfg::verify() — for every reachable block it asserts (1) a real terminator, (2) all referenced CfgValues defined, (3) block-param arities match per edge; release-safe panic!/assert! (NOT debug_assert) with fn name + block id; unreachable blocks skipped. Wired via mod verify + a cfg.verify() call at the end of opt::optimize (the choke point all build sites funnel through before codegen). Turns the recurring block-has-no-terminator SIGABRT class (RUE-213/217/224) into a loud, localized report. Part B: build.rs lower_value materializes a real Unit Const when an expr continues but yields no value.

**RUE-236:** a differential_opt marker on a CLI-test Case runs it across -O0/-O1/-O2/-O3 and asserts identical exit + stdout, naming the diverging level. 4 marked cases (arithmetic, control-flow, by-value aggregate, recursion) + 2 RUE-227 regression cases. -O2/-O3 alias -O1 today so results match; the net catches future opt divergence.

Verified: ./clippy.sh clippy-gate-passed; ./test.sh green; 59 rue-cfg tests incl. two should_panic proving verify() fires on a missing terminator + an arity mismatch; the 6 new CLI cases run across -O levels.

Fixes RUE-227
Fixes RUE-236

Generated with Claude Code

### DIFF
--- a/crates/rue-cfg/src/build.rs
+++ b/crates/rue-cfg/src/build.rs
@@ -2387,15 +2387,35 @@ impl<'a> CfgBuilder<'a> {
         self.value_cache[air_ref.as_u32() as usize] = Some(value);
     }
 
-    /// Lower an instruction and return its value, or None if it diverged.
-    /// This is a helper for use with the `?` operator when processing operands.
-    /// If the operand diverged, the caller should propagate the divergence.
+    /// Lower an instruction used in **value position** and return its value, or
+    /// `None` **only** if it diverged (return/break/continue).
+    ///
+    /// Callers use the `let Some(v) = self.lower_value(x) else { return
+    /// Self::diverged(); }` idiom, so `None` here means "control did not reach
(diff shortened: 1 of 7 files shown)


steveklabnik:
Superseded by a re-land that also fixes a macOS-only oracle-diff-test stack overflow: the oracle interpreter recurses per-expression, and the depth-60 deep_nesting corpus case (re-enabled in #1086) overflowed macOS's smaller default stack. The new PR runs the oracle-diff harness on a 256MB-stack thread.

PR #1110, opened 2026-07-03
First real Rue program: stdin integer stats (RUE-226 milestone)

steveklabnik:
examples/first/stats.rue — reads a count N then N integers from stdin and prints count/sum/max. The first program written in Rue as real code, flag-free, now that the four dogfooding features are stabilized (#1108).

Exercises the set end to end: @read_line, @parse_i64, String concat + @to_string, println, a payload enum (Maybe) modelling 'max of possibly-empty input', and a while loop with a match expression. Locked in as a CLI regression (cli.first_program: three-numbers + empty-set).

**Dogfooding surfaced real friction (the point of RUE-226):**
- @to_string only accepts i64 (i32 needs a cast) -> filed RUE-314.
- @parse bare-int failure + @read_line EOF both trap -> already tracked under RUE-6 error handling (stdin.toml); this is why the program uses a count-prefixed format instead of read-until-EOF.

Verified: '3/7/5/1' -> count 3/sum 13/max 7; '0' -> max '(no input)'; negatives -> max -5/sum -14. clippy-gate-passed; test.sh Pass 24.

Part of RUE-226
Part of RUE-1

Generated with Claude Code

### DIFF
--- a/.buckconfig.local
+++ b/.buckconfig.local
@@ -1,0 +1,2 @@
+[buck2]
+file_watcher = fs_hash_crawler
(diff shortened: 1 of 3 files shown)


steveklabnik:
Superseding with a version that uses the real Option (via @import) instead of a local Maybe enum — more honest about how optionals work in Rue today (no prelude → you import Option).

PR #1134, opened 2026-07-03
Buck2: link linux via cc + lld — full remote execution now default-capable (RUE-320)

steveklabnik:
Finishes RUE-316's remote execution. The remote worker needs the linker driver to be `cc` (clang++ isn't in the container), but `cc` alone falls back to `collect2 -> ld`, which the hermetic CI sandbox lacks. Fix: `linker = "cc"` **plus** `_LINUX_LINKER_FLAGS = ["-Clink-arg=-fuse-ld=lld"]` on the linux rust toolchains, so `cc` uses the rust-bundled lld and never touches system `ld`. **Global + CI-safe — no platform-scoping needed.**

Verified: native `./test.sh` Pass 24 with the wrapper emitting `cc -fuse-ld=lld`; and `--prefer-remote` links beef 5/5 remote (failed on 'clang++ not found' before). Both native-green and full-RE simultaneously — the acceptance criteria.

Fixes RUE-320

Generated with Claude Code

### DIFF
--- a/docs/process/build-cache.md
+++ b/docs/process/build-cache.md
@@ -7,23 +7,22 @@ tier) for a shared remote action cache — and full remote **execution** was pro
 to work locally: with the config below, the entire compiler built on BuildBuddy's
 workers (171/175 actions remote, 0 local, ~103s cold).
 
-> **Status (RUE-316 landed the groundwork; RUE-320 finishes it).** The remote
-> platform + the `$ORIGIN` toolchain-tree fix are in place, but full RE is **not
-> enabled by default and not yet landable as-is**. It needs the linker driver to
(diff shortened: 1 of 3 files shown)


steveklabnik:
Closing: this approach (unconditional -fuse-ld=lld, global cxx linker=cc) breaks the 'Build all targets' CI step on all 3 platforms — it only passed locally because test.sh builds tested targets (not //...) and dev cc has a working ld. Flagged on RUE-320 (moved to Backlog, High) for an infra decision: platform-scoped linker select() vs a fully-hermetic clang/lld toolchain. The remote action *cache* is unaffected (still opt-in). Will reopen a fresh PR once the approach is chosen.

PR #1279, opened 2026-07-06
[codex] Add module-qualified function identity

steveklabnik:
## Summary

- key free-function signatures by an internal module-qualified identity when duplicate source names exist across files
- resolve unqualified and module-qualified function references through the defining file before analysis/type-function reduction
- preserve existing type/function collision checks while allowing same-named functions in distinct modules
- add compiler and sema regressions for same-named module functions and function/type collisions

Fixes RUE-441.

## Validation

- ./fmt.sh
- ./buck2 test //crates/rue-air:rue-air-test //crates/rue-compiler:rue-compiler-test
- ./buck2 test //:cli-tests
- ./clippy.sh

### DIFF
--- a/crates/rue-air/src/sema/analysis.rs
+++ b/crates/rue-air/src/sema/analysis.rs
@@ -170,13 +170,17 @@ fn analyze_all_function_bodies_sequential(sema: &mut Sema<'_>) -> MultiErrorResu
                 continue;
             }
 
+            let function_key = sema
+                .resolve_function_name_in_file(*name, inst.span.file_id)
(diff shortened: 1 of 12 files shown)


steveklabnik:
Codex CI follow-up:

- Removed the stale `known_bug = "RUE-440"` marker because this PR fixes that regression.
- Updated the spec composition case that expected duplicate same-named imported functions to fail; module-qualified access is now expected to succeed.

Local validation:

- `./fmt.sh`
- `./buck2 test //:cli-tests //:spec-tests`

PR #1280, opened 2026-07-06
[codex] Allow module-local value constants

steveklabnik:
## Summary

- store value constants by defining file and source name while preserving bare-name lookup for globally unique constants
- resolve unqualified constants with current-file precedence and module-member constants through the module file
- keep same-file duplicate constants illegal while allowing same-named constants in distinct imported modules
- add a CLI regression for sibling modules exporting VALUE

Fixes RUE-456.

## Validation

- ./fmt.sh
- ./buck2 test //crates/rue-air:rue-air-test //:cli-tests
- ./clippy.sh


steveklabnik:
Codex CI follow-up:

- Updated the stale spec composition case for same-named value constants across imported modules. RUE-456 intentionally makes module-qualified access succeed.

Local validation:

- `./fmt.sh`
- `./buck2 test //:spec-tests //:cli-tests`

PR #1350, opened 2026-07-10
RUE-572: declaration-table checks resolve file-aware

steveklabnik:
Three sema checks still consulted the global by-name tables where the `(FileId, name)` maps are the truth (found by the RUE-431 flat-world audit, Tier 2):

- **Cross-kind collisions**: `check_const_cross_kind_collisions` compared every file-scoped constant against the **global** function/struct/enum tables, so a `const SHARED` in one module falsely collided (E0436) with a `fn SHARED` in an unrelated one. Items are module-local (RUE-490/491 removed unqualified cross-file resolution), so the check is now per-file — the same-file collision still errors (control case included).
- **Module-const widths**: `infer_const_init_types`' `FieldGet` arm resolved `m.CONST`'s declared integer width through the global constants table (its comment openly leaned on flat-namespace uniqueness). It now resolves the member in the receiver module's own file, so a same-named, different-width constant elsewhere can't hijack the width check; the global table remains a fallback for legacy flat-mode inputs only.
- **Enum visibility**: module-qualified enum lookup fell back to the global enums table after the file-aware miss, letting a same-named enum from an unrelated file satisfy a lookup (and its visibility attribution) that should fail. When the module's file is known, the member must be defined there.

**Verified by execution:** cross-module `const SHARED` + `fn SHARED` coexist (42); same-file collision still E0436; `b.WIDTH` (i64, 5000000001 arithmetic) unaffected by a u8 `WIDTH` in a sibling module (42); wrong-module enum access rejected. 4 new CLI cases. Full `./test.sh` exit 0.

Fixes RUE-572

### DIFF
--- a/crates/rue-air/src/sema/declarations.rs
+++ b/crates/rue-air/src/sema/declarations.rs
@@ -365,11 +365,10 @@ impl<'a> Sema<'a> {
 
         // Free functions duplicate-conflict only within their defining file
         // (RUE-441). Types duplicate-conflict only within their defining file
-        // (RUE-454). Function/type cross-kind collisions remain global until
-        // the remaining flat-namespace compatibility rules are removed.
-        let mut seen_function_names: HashMap<Spur, Span> = HashMap::new();
+        // (RUE-454). Function/type cross-kind collisions are per-file too
+        // (RUE-572, spec 10.5:1/10.5:2): top-level names are module-scoped,
+        // so a fn in one file and a struct in another may share a name.
         let mut seen_functions_by_file: HashMap<(FileId, Spur), Span> = HashMap::new();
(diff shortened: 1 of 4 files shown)


steveklabnik:
Queue bounce resolved: the failure was a semantic collision with the RUE-239 cases that landed on trunk after this branched — they asserted the cross-file cross-kind collisions this PR (and the now-merged #1351 spec) removes. Amended: both cross-kind duplicate detectors (sema's `check_top_level_name_collisions` AND the driver-level `detect_duplicate_symbols`) are per-file now, the three cross-file RUE-239 CLI cases and the `merge_symbols` unit test assert module-scoped coexistence, same-file collisions unchanged. `buck2 test //...` Pass 30 / Fail 0 on current trunk (tally checked directly). Re-armed.

PR #1370, opened 2026-07-10
Use split_once in diagnostic rendering test

steveklabnik:
Current trunk fails the clippy gate on `clippy::manual_split_once` in a diagnostic-rendering test helper.

Replace the manual `splitn(...).nth(1)` with `split_once`. This is a test-only mechanical cleanup; diagnostic behavior is unchanged.

Validation:

- `./buck2 test root//crates/rue-compiler:rue-compiler-test` — 218 passed
- `./clippy.sh` — all 41 targets clean
- `./fmt.sh check`


steveklabnik:
Superseded by trunk commit 47fef623, which applies the same split_once cleanup and restores the clippy gate.

PR #1445, opened 2026-07-11
RUE-619: stabilize ad-hoc Mach-O signing identity

steveklabnik:
## Summary

- pass an explicit `org.rue-lang.program` identifier to ad-hoc codesign and explicitly disable timestamp services
- strengthen the end-to-end reproducibility gate so its two native artifacts use different output basenames as well as different parent paths
- on macOS, strictly verify both code signatures and assert the exact stable identifier before accepting the artifacts

## Root cause

Standalone Rue Mach-O executables have no Info.plist. When no identifier is supplied, codesign derives one independently from each pathname; command-line tools therefore inherit their output filename. That identifier is sealed into the CodeDirectory, so identical source and options written under different names produce different final bytes.

The codesign contract also leaves timestamp behavior system-specific when no policy is supplied. Ad-hoc signatures are not timestamped today, but `--timestamp=none` makes that reproducibility boundary explicit.

Reference: [Xcode codesign(1)](https://keith.github.io/xcode-man-pages/codesign.1.html)

## Impact

The output path remains a destination choice, not a semantic compiler input. Rue still produces a valid ad-hoc signature required for ARM64 execution; the test runs the artifact and uses strict codesign verification on both copies.

## Validation

- `./fmt.sh check`
- `./buck2 test //crates/rue:rue-test` — 132 passed
- `./buck2 test //:reproducible-programs`
- `bash -n scripts/test-reproducible-output.sh`
- independent audit against Apple Security signing-source behavior and codesign flag semantics

The required macOS CI lane provides the platform-native byte and strict-signature validation.

### DIFF
--- a/crates/rue/src/main.rs
+++ b/crates/rue/src/main.rs
@@ -1744,8 +1744,21 @@ fn main() {
             {
                 // Only codesign if target is macOS (cross-compilation check)
                 if compile_options.target.is_macho() {
+                    // A command-line Mach-O has no Info.plist, so codesign's
+                    // default identifier comes from the output filename. Pin
+                    // both identity and timestamp policy: choosing a different
+                    // destination path must not change the program bytes
+                    // (RUE-619).
                     let result = Command::new("codesign")
-                        .args(["-f", "-s", "-", &options.output_path])
+                        .args([
+                            "-f",
+                            "-s",
+                            "-",
+                            "--identifier",
+                            "org.rue-lang.program",
+                            "--timestamp=none",
+                            &options.output_path,
+                        ])
                         .output();
                     match result {
                         Ok(output) => {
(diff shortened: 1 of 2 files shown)


crates/rue/src/main.rs
```diff
@@ -1744,8 +1744,21 @@ fn main() {
...
+                    // destination path must not change the program bytes
+                    // (RUE-619).
                     let result = Command::new("codesign")
-                        .args(["-f", "-s", "-", &options.output_path])
+                        .args([
+                            "-f",
+                            "-s",
+                            "-",
+                            "--identifier",
+                            "org.rue-lang.program",
```
steveklabnik, line 1758: this should be dev.rue-lang.program, no? our url is rue-lang.dev not rue-lang.org (which is a different rue)

PR #1512, opened 2026-07-12
RUE-707: implement macOS aarch64 stack-overflow clean abort

steveklabnik:
Replaces the aarch64-macOS no-op stub (from RUE-645) with a real handler: maps a 64 KiB alternate signal stack (`sigaltstack`, SYS 53) and installs a SIGSEGV handler via the raw Darwin `sigaction` syscall (SYS 46) with a hand-written `sa_tramp` trampoline (Darwin, unlike Linux, requires a caller-supplied signal trampoline). Since our handler exits, the trampoline only needs to invoke it with the signal number — no `sigreturn` epilogue. A macOS stack overflow now aborts cleanly ("stack overflow", exit 101) instead of a raw SIGSEGV (139), matching the Linux targets.

**Best-effort throughout**: any syscall failure leaves the default disposition, and a wrong trampoline only affects the rare stack-overflow path (worst case = the prior exit-139), never a normal program.

**Verification**: type-checks for aarch64-apple-darwin (`rustc --target aarch64-apple-darwin -C panic=abort`, clean). Behavioural correctness is CI's `test (macos)` job — the `stack_overflow.toml` case is now scoped `only_on` all three targets including aarch64-macos, so a regression there fails loudly (not silently masked). Full `./test.sh` green on x86-64.

Fixes RUE-707

### DIFF
--- a/crates/rue-cli-tests/cases/stack_overflow.toml
+++ b/crates/rue-cli-tests/cases/stack_overflow.toml
@@ -15,15 +15,15 @@
 # stack overflow: safe Rue has no other way to segfault (bounds-checked, no null
 # derefs, arithmetic traps). See crates/rue-runtime/src/entry.rs.
 #
-# Platform scope (`only_on` the two Linux targets): the handler is implemented
-# for x86-64 Linux and aarch64 Linux. macOS is a documented TODO — Darwin's
-# `sigaction(2)` requires a caller-supplied signal trampoline that cannot be
-# verified on a Linux host, so its `install_stack_overflow_handler` is a no-op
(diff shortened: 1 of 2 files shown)


steveklabnik:
Closing: CI's test (macos) job proved this blind Darwin sigaction implementation is wrong — it faults at install in _start and breaks 1509/2049 macOS tests (not just the stack-overflow case). The syscall struct layout / trampoline convention is off in a way I cannot debug without Apple Silicon. Reverting to the honest no-op stub; RUE-707 needs on-hardware work. See RUE-707.

PR #1692, opened 2026-07-16
Draft ADRs for review: function inlining (0049) and loop optimizations (0054)

steveklabnik:
## NOT for auto-merge — design drafts for maintainer review

These are the design notes RUE-915 and RUE-916 call for, status `proposed`. They analyze and recommend; they decide nothing. Please read at leisure — no urgency.

**ADR-0049 (function inlining, RUE-915):** recommends a dedicated inter-procedural stage in `rue-compiler` between CFG construction and per-function optimization, consuming ADR-0050's `body_dependencies` edges for incremental invalidation rather than inventing a parallel call graph. Covers the mechanics (slot/value/block rebasing, by-ref parameter redirection to caller places, return wiring via block params, drop handling — drops copied verbatim, never re-elaborated), conservative thresholds (leaf + small + single-call-site at -O2), recursion refusal via SCC, and the honest cost: the current fused build+optimize parallel map must split. One open question flagged: drop-ownership of a moved-and-dropped by-value argument needs a concrete elaboration example before implementation.

**ADR-0054 (LICM + unrolling, RUE-916):** builds natural-loop detection and preheaders on the dominator tree that landed with RUE-914. The central design ruling requested: LICM's trap-safety rule — hoisting a possibly-trapping op onto a zero-iteration path manufactures a trap the source never executes (the inverse of RUE-57), so the initial version never hoists trapping ops; the relaxation path (guarded hoisting after loop rotation) is named but deferred. Unrolling: constant-trip-count full unroll only, under a code-growth budget shared with future inlining. (Renumbered from 0051, which canonical-import-resolution claimed.)

Both drafts' code citations were verified against current trunk after tonight's optimizer work landed.

Related: RUE-915, RUE-916 (both Backlog pending these decisions), RUE-909.

### DIFF
--- a/docs/designs/0049-function-inlining.md
+++ b/docs/designs/0049-function-inlining.md
@@ -1,0 +1,683 @@
+---
+id: 0049
+title: Function Inlining
+status: accepted
+tags: [compiler, codegen, optimization]
+feature-flag: none
+created: 2026-07-15
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+---
+
+<!-- Note: Inlining is a compiler-internal optimization, not a language feature.
+     It never changes what a program means, only how the emitted code is shaped
+     (see ADR-0044's observable-behavior invariant). No preview gate applies. -->
+
+# ADR-0049: Function Inlining
+
+## Status
+
+Proposal
+
+This is a **draft design note for maintainer review**. It analyzes the
+architecture of function inlining and recommends a direction; it makes no
+commitment and nothing here is accepted. ADR-0044 already fixed the *level*
+placement — conservative small/leaf-function inlining at `-O2`, aggressive
+thresholds at `-O3` — so this note answers the *architecture* questions that
+sit under that placement: where inlining runs, what it does mechanically to the
+CFG IR, how it integrates with the **already-landed durable CFG cache**, how it
+handles **parameter ownership and drop**, and how it stays inside the
+observable-behavior invariant. Since the durable CFG cache (`DurableCfgArtifact`)
+and the parameter-mode physical ABI both landed since the first draft, **durable
+caching and parameter ownership are core design constraints here, not deferred
+implementation details**. Tracks RUE-915.
+
+## Summary
+
+Inlining replaces a `Call` to a suitable callee with a copy of the callee's body
+spliced into the caller, renumbered onto the caller's frame and value space, so
+that intra-procedural passes (constant folding/propagation, peephole, CFG
+simplification, value forwarding, CSE, DCE, and the future loop passes) can then
+see across the former call boundary. The central tension
+is that Rue's optimizer is today strictly **per-function**: `build_functions_and_cfgs`
+builds and optimizes each function's CFG independently and in parallel
+(`crates/rue-compiler/src/queries.rs:198`, `into_par_iter().map(...)`), whereas
+inlining is inherently **cross-function** — it needs the callee's body while
+optimizing the caller. This note recommends running inlining as a **dedicated
+inter-procedural stage orchestrated by `rue-compiler`, operating on already-built
+callee CFGs**, sequenced between CFG construction and the existing per-function
+`opt::optimize` pass. That choice keeps `rue-cfg`'s passes purely local and
+reuses the CFG the builder already produces.
+
+Two consequences are load-bearing and are given their own decision sections:
+
+- **Durable cache integration is not optional and cannot be deferred.** Rue
+  already retains, imports, and exports optimized `DurableCfgArtifact`s keyed by
+  optimization level and target (`crates/rue-compiler/src/queries.rs:137`,
+  `217-232`, `330-343`). Inlining changes *what* a cached caller artifact
+  contains (foreign callee domains) and *what invalidates it* (callee body
+  edits). The ADR decides now: cached CFGs are **final optimized, post-inlining**
+  CFGs; caller cache keys must incorporate **callee body fingerprints**; and the
+  caller-only domain projection (`CfgDomainProjection::from_body`) must be
+  extended to carry **foreign callee domains** (§4).
+- **The inliner's call-site graph is a separate structure from the semantic
+  dependency manifest** (§5). Call-site multiplicity, leafness, candidate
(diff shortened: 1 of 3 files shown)


steveklabnik:
I think the overall direction here is strong, but after checking the proposals against the PR's base commit, my current take is:

- ADR-0054 is close to acceptance with targeted corrections.
- ADR-0049 has the right architectural direction, but needs a substantial revision before acceptance.

The strongest parts are the layering decision (interprocedural orchestration in `rue-compiler`, CFG surgery in `rue-cfg`), LICM's conservative initial rule that trapping operations never move, recomputing loop analysis after mutations, differential optimization coverage, SCC refusal, and bounded code growth.

For ADR-0049, the main issue is that its caching premise is already outdated on this PR's base. Rue currently retains, imports, and exports optimized `DurableCfgArtifact`s in `queries.rs`, keyed by optimization level and target. That means cache integration cannot be deferred to Phase 5. The ADR needs to decide now:

- whether cached CFGs represent pre-inlining or final optimized CFGs;
- how caller cache inputs incorporate relevant callee body fingerprints and inlining policy;
- how durable domain projection handles types, strings, symbols, and spans copied from the callee.

The last point is particularly important: `CfgDomainProjection::from_body` currently derives its mapping solely from the caller's AIR body and rejects spans outside that body. An inlined callee introduces foreign domains by definition, so the existing export/import boundary cannot simply treat the resulting CFG as an ordinary caller artifact.

I would also separate the semantic dependency graph from the inliner's call-site graph:

- scan actual CFG `Call` instructions for call-site multiplicity, leafness, candidate lookup, and recursion SCCs;
- use the semantic dependency manifest for invalidation and durable reuse.

`body_dependencies` are deduplicated definition dependencies containing calls, types, destructors, and other inputs. They are not a call-site multigraph, so they cannot distinguish one call from ten, and using the whole dependency graph for recursion analysis can produce false cycles. Scanning the CFG is not a competing invalidation path; it answers a different question.

The parameter/drop mechanics also need tightening:

- A materialized by-value parameter needs explicit callee-lifetime `StorageLive`/`StorageDead` and the copied parameter-drop behavior, not ordinary caller-scope drop treatment.
- Aggregate arguments need their full ABI slot width.
- Logical `Borrow`/`Inout` mode does not always imply physical by-reference transport: slice borrows can be passed by value. The splice must consult the callee's physical `is_param_by_ref` information.
- Projected by-ref arguments need a worked proof that direct place substitution preserves address-formation timing and trap behavior. I would initially exclude them, or support only simple local/parameter roots.
- Removing a single-call-site callee should be a separate whole-program dead-function-elimination decision.

For ADR-0054, I would make four corrections:

1. Define a preheader as a dedicated block. Even with one outside predecessor, split the edge when that predecessor has other successors, and specify how header block arguments are forwarded.
2. Include `IntCast` explicitly in the trapping set; DCE already treats it as trapping. Ideally factor a shared `may_trap` / `is_speculatable` classification instead of making DCE's private predicate the authority.
3. Specify post-unroll cleanup. `constopt` currently runs near the beginning of `optimize`; if unrolling happens after recognition, another const-fold/simplify/DCE sequence is needed to collapse induction variables and cloned control flow.
4. Restrict the initial unroller to an explicit canonical shape—probably single header, single latch, recognized IV, and no unsupported exits—or describe full CFG-subgraph cloning rather than “straight-line copies.”

So my recommendation is: accept the core LICM decision, revise the unrolling mechanics, and rewrite ADR-0049 around the already-landed durable CFG cache. The inlining architecture still looks promising, but caching and parameter ownership are central design constraints now rather than later implementation details.

steveklabnik:
Revision pushed incorporating the full review; every mechanism cited was re-verified against current trunk before the text references it.

**ADR-0049** (substantial rewrite around the landed durable CFG cache):
- §4a: cached CFGs are **final post-inlining** artifacts — export already sits after `opt::optimize` (`queries.rs:282, 330`); pre-inlining caching rejected with rationale.
- §4b: caller keys gain `{callee → StableDefinitionInputFingerprint}` for every callee actually inlined, plus an `inlining_policy_version` (initially folded into the schema version). Callers that inline nothing keep today's key.
- §4c: domain projection becomes a caller-projection **unioned with each inlined callee's projection** (callee spans anchored to their own `body_span` with a callee-identity tag); the `durable_cfg.rs:237-242` span check and the `import_cfg` remap table change accordingly. Until that lands, inlined callers fail the export gate and simply aren't cached — correct, just slower.
- §5: call-site graph scanned from CFG `Call` instructions (multiplicity, leafness, SCC recursion) vs the dependency manifest (invalidation/durable reuse only), with the one-call-vs-ten and false-cycle rationale.
- §3: materialized by-value parameters get callee-lifetime `StorageLive`/`StorageDead` and discharge the callee's copied `param_drops` obligation — no caller-scope drop, no double drop (`functions.rs:387`, `build.rs:329` cited). Aggregates use full ABI slot widths; the splice consults physical `is_param_by_ref`; projected by-ref arguments are **excluded initially** (simple local/parameter roots listed as the possible relaxation with the address-formation proof obligation); single-call-site callee removal deferred to a separate whole-program DFE decision.

**ADR-0054** (four corrections, LICM core intact):
1. Preheader is a dedicated block: the edge is split even with one outside predecessor when that predecessor has other successors, and header block-argument forwarding is specified.
2. `IntCast` is in the trapping set (verified: DCE treats it as trapping at `dce.rs:157` via `__rue_intcast_overflow`), and a shared `may_trap`/`is_speculatable` classifier is proposed so DCE's private `has_side_effects` stops being the cross-pass authority — it conflates the side-effect and trap axes.
3. Post-unroll cleanup is explicit: `constopt` runs first in `optimize()` (`opt/mod.rs:160`), so unrolling is followed by another const-fold/simplify/DCE sequence to collapse IVs and cloned control flow.
4. Initial unroller restricted to a canonical shape (single header, single latch, recognized IV, no unsupported exits) with full CFG-subgraph cloning language replacing "straight-line copies".

Housekeeping: both drafts shipped an invalid registry status (`proposed`); corrected to `proposal` and the README index regenerated via `scripts/validate-adrs.py --write` (55 records valid).

PR #1697, opened 2026-07-16
RUE-920: spec carve-out for program-wide main uniqueness

steveklabnik:
The compiler deliberately rejects a second `main` in any loaded module (E0436 at declaration-gathering time, program-wide — landed via RUE-582 per ADR-0047), but the normative spec said the opposite: 10.5:1 scoped the duplicate check strictly per-file and 10.5:2 allowed any top-level name to repeat across loaded files.

This aligns the prose with the decided behavior:

- New legality rule **10.5:5** and example **10.5:6**: `main` is the one top-level name whose uniqueness is program-wide, checked eagerly at declaration gathering, with ADR-0047 cited for rationale.
- 10.5:1 and 10.5:2 now point at the exception; every other name stays per-file/module-scoped.
- Cross-reference added from the 6.1 entry-point rules.
- Two new spec cases citing 10.5:5: an imported module defining an unused `main` is rejected with E0436, and duplicate non-`main` names across files still compile.
- Diagnostic corpus snapshot re-blessed on this tree (additive only: accepted 3763 → 3768, rejected rows unchanged).

Verification: E0436 rejection reproduced by hand on this checkout; composition (11/11) and functions (129/129) spec suites; traceability gate green at 100% normative coverage; quick suite 27/27.

Note: this will need a trivial snapshot re-bless if it lands after #1696 (which also grows the corpus); whichever bounces gets rebased and re-blessed.

Fixes RUE-920

### DIFF
--- a/.agents/skills/integrate-worker/SKILL.md
+++ b/.agents/skills/integrate-worker/SKILL.md
@@ -1,0 +1,78 @@
+---
+name: integrate-worker
+description: Integrate a worker agent's diff onto fresh trunk — apply, verify repros by hand, cross-check interacting mechanisms, full suite, ship + queue. Use after any worktree worker (fix cycle, hunt follow-up) returns a diff file.
+---
+
+# Integrating a worker diff
+
+Battle-tested protocol from the 2026-06 autonomous runs (~30 PRs integrated).
+Each step exists because skipping it once caused a real failure.
+
+## Steps
+
+1. **Fresh trunk, always.** `jj git fetch && jj new 'trunk()'` then
+   `git reset --quiet` — the colocated git index goes stale after jj
+   working-copy switches and makes `git apply --3way` fail with
+   "does not match index".
+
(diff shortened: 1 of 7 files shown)


steveklabnik:
I don't think that this is intended. The top-level `main` should be special and required, but a `main` defined in another module shouldn't cause a compilation failure. There's no chance of conflict because there is namespacing now.

The spec language implying it was whole-program unique was because we didn't have said namespacing at the time, not because that was the intended forever semantics.

steveklabnik:
Closing without merge per the review comment: the program-wide `main` uniqueness check is not the intended semantics — it predates module namespacing. The top-level `main` remains special and required, but a `main` defined in another module should be an ordinary namespaced function, not a compile error.

That inverts this PR: instead of documenting the program-wide check in the spec, the compiler check in `canonical_merge.rs` gets narrowed to the root module (per-file duplicates still E0436), and the existing per-file spec text stands. Rework tracked on RUE-920/RUE-921.

PR #1705, opened 2026-07-16
RUE-920: scope main to the root module

steveklabnik:
The compiler rejected a second top-level `main` in **any** loaded module (program-wide E0436 at declaration-gathering time — the RUE-582/ADR-0047 transitional guard). Per the maintainer decision on #1697, that predates module namespacing: only the root module has an entry point, and a `main` defined elsewhere should be an ordinary namespaced function.

Two mechanisms enforced the old rule, both fixed:

- `canonical_merge` drops the program-wide check. Same-file duplicates — including two `main`s in one file — remain per-file E0436 under the ordinary spec 10.5:1 rule.
- Symbol naming was the deep enforcement: `internal_function_name` gave **every** `main` the bare entry symbol. Now only the root modules `main` keeps it; a non-root `main` always mangles (`__rue_fn_<module>__main`), so it cannot collide with the entry point, and a root without `main` still fails with E0200 rather than silently adopting an imported entry.

Spec: new rule 6.1:38 (root-module entry-point selection), cited by new spec cases; 10.5 prose unchanged. ADR-0047 gains an amendment note recording that RUE-920/RUE-921 retired the transitional check.

Verification, all by execution on the integrated tree: root + imported `main` compiles and runs (exit 42); explicit `helper.main()` call returns the imported value (99); same-file duplicate still E0436; root-missing-`main` with an import that has one still E0200; mangled non-root `main` confirmed in both x86-64 and AArch64 `--emit asm`; composition (11), functions (130), traceability (100% normative), quick, and corpus suites green; full local suite 36/37 with only the pre-existing `benchmark-tool-tests` Python-env failure (host Python 3.9.6 lacks `tomllib`; identical on clean trunk).

Fixes RUE-920


steveklabnik:
CI failure root-caused and fixed: five legacy positional flat-mode CLI cases (`args = ["lib.rue", "main.rue", "-o", "prog"]` with `main` in a non-first file) contradicted root-scoped `main`. Per maintainer decision, the cases are modernized to `@import` form — each regression surface (RUE-175 span attribution to a non-root FileId, RUE-130 qualified enum patterns in a non-root instruction space) is preserved, with the error under test now living in the imported file. RUE-923 (High) tracks finishing the flat-mode removal; RUE-924 tracks why the local full suite never ran `//:cli-tests` and reported false-green.

Re-verified: `multifile_errors` 12/12, `modules` 123/123, `duplicate_main` 3/3, corpus snapshot re-blessed and green.

steveklabnik:
Second CI failure fixed: the reproducibility gate had two more flat-mode-era shapes — the semantic-order fixture deliberately kept `main` outside the designated root ("the root need not have the lowest numeric FileId"), and the source-order IR check permuted `main.rue` across noise siblings. Both redesigned for root-owned entry (spec 6.1:38): the semantic-order root now imports both siblings and delegates to `left.run()` (runtime behavior, colliding-glue symbols, ADR-0051 absent-arm probing, and manifest/env/parallelism perturbation all preserved); the source-order check pins `main.rue` first and permutes the noise files.

Verified locally: `//:reproducible-programs` fully green (all targets × opt levels byte-identical, fixture-runs assertions pass) and the full CLI corpus 1468/1468. These fixtures are more evidence for RUE-923 (finish the flat-mode removal) — they were load-bearing consumers of the vestige.

PR #1717, opened 2026-07-17
RUE-838: remove canonical type migration scaffolding

steveklabnik:
## Summary

- remove the retired `InternedType` universe, compatibility encoding, and whole-definition overwrite APIs
- make composite category IDs opaque and keep ordering, naming, and raw storage operations pool-owned
- add a glob-backed architecture inventory covering every live-type consumer and mark ADR-0024 implemented

## Validation

- `scripts/rue test`
- `scripts/rue fmt`
- `scripts/validate-adrs.py`
- adversarial architectural review: no findings

Fixes RUE-838

### DIFF
--- a/crates/rue-air/src/api_inventory.rs
+++ b/crates/rue-air/src/api_inventory.rs
@@ -48,3 +48,92 @@ fn canonical_import_consumers_do_not_grow_resolution_policy() {
         );
     }
 }
+
+#[test]
+fn canonical_type_surface_has_one_checked_handle_and_private_storage_ids() {
+    let types = include_str!("types.rs");
+    let pool = include_str!("intern_pool.rs");
+    let encoding = include_str!("type_encoding.rs");
+    let exports = include_str!("lib.rs");
+    let public_surface = [types, pool, encoding, exports].concat();
+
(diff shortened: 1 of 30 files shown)


crates/rue-air/src/api_inventory.rs
```diff
@@ -48,3 +48,92 @@ fn canonical_import_consumers_do_not_grow_resolution_policy() {
         );
     }
 }
+
+#[test]
+fn canonical_type_surface_has_one_checked_handle_and_private_storage_ids() {
```
steveklabnik, line 53: @DorianListens these sorts of tests are things codex has been writing and it is *so interesting*! I think that they make sense for allowing it to validate its own work, but they aren't as useful in the future, so we might want to do a cleanup pass of "which tests are pulling their weight and which ones can be removed" at some point.

DorianListens: Yeah this is almost the test equivalent of the "architecture evolution" comments we've been removing, I guess. Reasonable as a proof of a specific kind of change, but not useful later on. That said: it should run extremely fast, so fairly low cost to keep around.

PR #1732, opened 2026-07-17
RUE-927: LICM hoists trap-free invariant ops to the preheader at -O3

steveklabnik:
ADR-0054 Phase 2 — the first pass that makes `-O3` distinct from `-O2`. The new `opt/licm.rs` pass recomputes the dominator tree and natural-loop forest per sweep, processes loops innermost-first, and marks a body instruction invariant when every operand is defined outside the body or by an already-hoisted invariant. Hoisting relocates the instruction into the RUE-926 preheader; value ids are unchanged so uses stay valid.

**The governing rule is absolute:** only `classify::is_speculatable` ops move. Overflow-checked arithmetic, `IntCast`, and indexed reads never hoist — a trapping op moved into the preheader of a zero-trip loop would manufacture a trap the source never executes (the inverse of RUE-57). Block-param values are never invariant, and memory reads are non-invariant whenever the loop body contains any observable side effect (phase-2 conservatism, relaxable with memory versioning). `dce::visit_instruction_uses` becomes `pub(super)` so LICM reuses the single exhaustive operand enumeration rather than duplicating a match that would rot.

Integrated across RUE-840 (typed CFG payload side tables): `licm::run` threads the frozen type pool into preheader materialization and returns `Result<Stats, CfgOptimizationError>`; the `optimize()` pass closure widens to `CfgOptimizationError` (the existing `From<CfgEditError>` absorbs the other passes unchanged); test scaffolding migrated to typed payloads.

Work counters per RUE-794 with a bound test; nine unit tests (hoisting, trapping refusal, block-param refusal, nested placement, irreducible no-op, verifier cleanliness); three differential CLI cases per RUE-236 (hoistable bitwise computation, the zero-trip trap-safety shape, effect ordering).

Verified by execution on the integrated tree: a speculatable bitwise computation is hoisted into the preheader at `-O3` (CFG dump shows the `shl` before the loop) with output identical to `-O0`; a zero-trip loop containing `100/d` with `d=0` exits cleanly at both levels — the trap is **not** manufactured; `rue-cfg` suite and re-blessed corpus green.

Fixes RUE-927

### DIFF
--- a/crates/rue-cfg/src/opt/dce.rs
+++ b/crates/rue-cfg/src/opt/dce.rs
@@ -195,9 +195,12 @@ fn visit_terminator_uses(cfg: &Cfg, term: &Terminator, mut f: impl FnMut(CfgValu
 /// Visit values used by an instruction.
 ///
 /// Calls the provided function for each value used by the instruction.
-/// This avoids allocating a Vec for each call.
+/// This avoids allocating a Vec for each call. Exposed to sibling passes
+/// (LICM, RUE-927) so operand enumeration lives in exactly one exhaustive match
+/// over `CfgInstData` here — a new instruction variant cannot silently gain an
+/// operand that LICM's invariance check fails to see.
 #[inline]
-fn visit_instruction_uses(cfg: &Cfg, value: CfgValue, mut f: impl FnMut(CfgValue)) {
+pub(super) fn visit_instruction_uses(cfg: &Cfg, value: CfgValue, mut f: impl FnMut(CfgValue)) {
     match &cfg.get_inst(value).data {
         // Constants and parameters have no uses
         CfgInstData::Const(_)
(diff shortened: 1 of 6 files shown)


steveklabnik:
Converting to draft: CI caught a real defect — at `-O3` the merge tree ICEs with "value vN is unattached" post-optimization on `cli.differential_opt::merged_blocks_across_opt_levels` and `cli.second_program::calculator_left_associative_with_ascii_whitespace`. This did not reproduce on the pre-bounce integration tree, so it is an interaction between LICM's hoist (instruction relocation between block inst lists) and something that landed on trunk after — suspicion: the block-merging/preheader-reuse seam under the newest rue-cfg changes. Needs a real fix, not a re-bless; do not merge.

PR #1824, opened 2026-07-18
Accept parallel demand-driven incremental compilation

steveklabnik:
## Summary

- accept ADR-0063 as Rue's parallel demand-driven incremental compilation architecture from source inputs through per-function code generation
- define immutable revision-pinned attempts, exact per-key claim-or-join, red/green validation, deterministic diagnostics, bounded parallel execution, and position-stable semantic identity
- amend ADR-0051's whole-graph import staging into batched rooted demand fulfillment; supersede ADR-0045 and ADR-0053 with explicit survivorship rules
- establish CodegenUnit and ProgramImagePlan as the stable seam while keeping the first implementation on the existing fresh linker
- link the High-priority Linear project, RUE-648 epic, and RUE-1021 through RUE-1033 phase issues

## Adversarial review

All eleven review threads are addressed in the current revision. In particular, the ADR now specifies request-versus-attempt revision pinning, batched demand rounds, speculation without host I/O, database-owned reachability with correct edge deletion, scheduler progress under permit exhaustion, shared Rayon/query budgeting, position-independent identities and red/green equality, key-versus-fingerprint semantics, warning-free image plans, and a family-by-family compatibility-shim migration.

The review threads remain open for reviewer verification.

## Tracking

- Project: https://linear.app/steve-klabnik/project/parallel-demand-driven-incremental-compilation-b720ac2f2c9f
- Epic: RUE-648
- Phases: RUE-1021 through RUE-1033

## Validation

- python3 scripts/validate-adrs.py
- scripts/rue quick (33 passed)
- git diff --check

### DIFF
--- a/docs/designs/0063-parallel-demand-driven-incremental-compilation.md
+++ b/docs/designs/0063-parallel-demand-driven-incremental-compilation.md
@@ -1,0 +1,712 @@
+---
+id: 0063
+title: "Parallel demand-driven incremental compilation"
+status: proposal
+tags: [architecture, compiler, incremental, parallelism, codegen, linker, performance]
+feature-flag: null
+created: 2026-07-18
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["ADR-0045", "ADR-0050", "ADR-0051", "ADR-0052", "ADR-0053", "ADR-0055", "ADR-0058", "ADR-0061", "RUE-328", "RUE-812"]
+---
+
+# ADR-0063: Parallel demand-driven incremental compilation
+
+## Status
+
+Proposal. This ADR does not authorize implementation yet. It is intended to
+receive architectural review before acceptance and before an implementation
+project is created.
+
+If accepted, this ADR supersedes ADR-0045's compiler-architecture rollout and
+its exclusion of cross-request and cross-invocation incremental state. It
+retains and restates ADR-0045's language-semantic rule that observable semantic
+checking and emitted code are selected by explicit roots. It also supersedes
+ADR-0053's single-threaded, one-selected-key execution boundary and its decision
+to end the query database before code generation. The attempt, diagnostic,
+dependency, current-versus-last-good, and fail-closed publication invariants
+from ADR-0053 remain requirements.
+
+Until this proposal is accepted, ADR-0045 and ADR-0053 remain authoritative and
+their metadata is unchanged.
+
+## Summary
+
+Rue will organize compilation as a revisioned graph of typed, memoized queries
+from source input through per-function code generation. Observable compilation
+is demand-driven from an explicit root set, normally the executable entry
+point. Query results use stable semantic identities and request-independent
+artifacts, so unchanged results can be retained across edits, evaluated in
+parallel, and eventually persisted across processes.
+
+The execution substrate will support immutable pinned input revisions,
+per-key claim-or-join computation, exact dependency tracking, red/green change
+propagation, cancellation, deterministic diagnostics, and bounded retention.
+The first implementation may schedule queries serially, but it may not encode a
+global mutable computation stack, a single selected key per family, or another
+constraint that would require replacing the state model to run independent
+queries in parallel.
+
+The terminal compiler query artifact is a stable per-function `CodegenUnit`,
+not serialized object-file bytes. A deterministic `ProgramImagePlan` collects
+the reached units and runtime requirements. The initial implementation may
+perform a fresh full internal link from that plan. A later incremental linker
+can consume additions, removals, and changed unit fingerprints directly,
+without changing frontend, semantic, CFG, or codegen query boundaries.
+
+## Motivation and target
+
+Rue already has useful pieces of incremental compilation:
+
+- immutable parsed-module artifacts and per-module parse reuse;
+- stable module and definition identities;
+- a typed query graph with dependency stamps and retained attempts;
+- demand-driven ordinary body analysis from `main`;
+- canonical semantic body/type/value artifacts and fail-closed re-import;
+- per-function durable CFG projections; and
+- per-function native code generation on both supported architectures.
+
+The pieces do not yet form a demand-driven pipeline. Import discovery reads the
+complete transitive import graph. Parse, merge, RIR lowering, and declaration
+resolution process every loaded module. Named destructors are implicit roots,
+drop glue scans the complete type pool, semantic and optimized CFG artifacts
+share one query terminal, and machine code/object/link work is performed afresh
+for every executable request.
+
+The current query engine is also deliberately serial. A family owns one selected
+state, a second in-flight key is rejected, and active computation/cycle state is
+tracked by query-family type rather than exact keyed node. `CompilerSession`
+orchestration requires exclusive mutable access. Making those fields individually
+thread-safe would preserve the serial ownership model while adding contention;
+it would not produce a parallel query engine.
+
+The long-term performance objective is an interactive warm rebuild: an edit to
+one reachable function should normally re-run only the queries whose observable
+inputs changed, generate a replacement code unit, and update a runnable internal
+link in tens of milliseconds on a representative project. This ADR does not set
+a hardware-independent latency promise. It requires the architecture,
+structural work metrics, and benchmark scenarios needed to pursue and measure
+that class of result.
+
+## Decision
+
+### 1. Observable roots are separate from speculative computation
+
+Every compilation request owns an explicit, ordered root set. An executable
+request normally roots the target-specific `main` definition. A future
+`check-all` request may add public or all declarations as semantic roots. Other
+tools may request syntax, definition, semantic, CFG, or codegen artifacts
+directly through supported session operations.
+
+The transitive dependency closure of those roots determines:
+
+- which semantic failures and warnings are observable for that request;
+- which runtime functions, user functions, specializations, destructors, and
+  drop-glue units enter the program image; and
+- which codegen units are required before linking.
+
+The scheduler may speculatively evaluate a pure query which is not yet in the
+root closure. A speculative result may populate the memo database, but its
+diagnostics, warnings, or emitted unit do not become observable until a rooted
+query reads it. This separates language semantics from scheduling policy and
+allows safe background work without turning `check-all` into the default build
+mode.
+
+Lexing and parsing remain whole-file operations after a module is demanded.
+Therefore a demanded module must be syntactically valid as a whole. An
+undemanded module is not read or parsed and contributes no syntax diagnostics.
+Semantic checking, CFG construction, and code generation are per declaration or
+per concrete function instance. Item-level error-tolerant parsing is separate
+future work and is not required by this ADR.
+
+### 2. Inputs are immutable revisions
+
+The query database retains immutable input revisions. A request pins exactly one
+revision for its lifetime. Publishing a source edit, target/configuration edit,
+or host import observation creates a successor revision; it does not mutate the
+inputs observed by already-running work.
+
+Input leaves are granular:
+
+- one root/request configuration;
+- one module source and its accepted-read provenance;
+- one import observation;
+- target, preview, optimization, ABI, and layout configuration; and
+- compiler/schema implementation epochs where they affect retained artifacts.
+
+A module query depends on its exact module source leaf, not on a whole-program
+snapshot containing every source currently known to the host. Adding an
+unrelated source therefore does not invalidate an already-parsed module.
+
+Current-versus-last-good selection belongs to a request/session publication
+layer over immutable revisions. It is not mutable state inside every query
+algorithm.
+
+### 3. Query identity, computation, and publication
+
+A logical query key identifies a semantic entity plus explicit configuration.
+Examples include a module identity, stable definition key, concrete
+specialization key, type-instance key, target, or optimization level. A logical
+key never embeds terminal stamps from its dependencies. Dependency stamps are
+recorded on observed graph edges.
+
+The database allocates one logical memo node per `(query family, logical key)`
+and may retain versioned attempts/results for that node. Requesting a node from
+a pinned revision validates a compatible retained result or performs an atomic
+claim-or-join operation:
+
+- a valid terminal is reused;
+- one task claims a dirty or unpublished key and computes it;
+- another request for that exact in-flight key and compatible pinned inputs
+  joins the computation;
+- a different ready key may compute concurrently; and
+- a request which would wait through a true dependency cycle receives the
+  family-defined cycle result or a query-control abort.
+
+Requests for the same logical key under incompatible revisions do not join
+blindly. They may compute concurrently or reuse only after dependency validation
+proves that the observed inputs are equivalent.
+
+No database shard or query-state lock may be held while executing a query body
+or requesting a dependency. Computation builds its value, diagnostics, work,
+and observed dependencies privately and publishes them atomically after
+checking that its pinned revision is still eligible for the requesting
+publication.
+
+Publication is red/green. If a recomputed query has an outcome and diagnostic
+batch equal under its family-owned comparison, its observable terminal stamp is
+preserved. Dependents are not invalidated merely because an implementation ran
+again or an input revision number advanced.
+
+Attempt history, current/last-good behavior, deterministic failures,
+cancellation, and fail-closed retained-artifact validation continue to follow
+ADR-0053. Duplicate in-flight work changes from a rejected attempt to a joined
+request. Cancellation of one waiter does not cancel shared work while another
+live waiter or retained root still requires it.
+
+### 4. Parallel execution is schedule-independent
+
+The database and all retained query keys/records are `Send + Sync`. Query
+algorithms consume immutable dependencies and return immutable `Arc` artifacts.
+Shared mutable interners, globally appended type pools, discovery-order counters,
+and request-local indices do not cross query boundaries.
+
+The evaluator maintains task-local dependency stacks and a cross-task wait graph
+for cycle detection. Stable output does not depend on worker order:
+
+- module, definition, specialization, symbol, anonymous nominal, and string/data
+  identities are derived from stable semantic inputs rather than allocation
+  order;
+- diagnostic presentation is sorted by stable module/source identity, source
+  anchor, and producer-defined order after collection;
+- reached codegen units and image atoms are sorted by stable identity before
+  image planning; and
+- reductions of work metrics are deterministic.
+
+The runtime owns one concurrency budget. Query-level parallelism and parallel
+algorithms inside a query must use structured permits from that budget rather
+than nesting uncoordinated global thread pools. The first implementation may run
+the same interfaces with one worker. Correctness and artifact identity must be
+identical for one worker and many workers.
+
+### 5. Stable identity and canonical artifacts are the interchange format
+
+Query results which outlive one local computation contain no request-local
+`FileId`, raw `Span`, `Spur`, `InstRef`, live `Type`, nominal pool ID, AIR/CFG
+offset owned by another artifact, or pointer identity.
+
+The stable identity domain covers:
+
+- named definitions and named nominal types;
+- free functions, methods, associated functions, and destructors;
+- concrete generic/comptime specializations with canonical type/value
+  arguments;
+- anonymous structs/enums and their methods/destructors, keyed by their stable
+  producing definition or specialization, source anchor, and canonical
+  arguments;
+- synthesized drop glue keyed by canonical type instance; and
+- runtime/compiler-provided symbols from the typed ABI manifest.
+
+ADR-0058's canonical semantic body/type/value algebra is the semantic query
+interchange format. Compact live AIR, live type pools, parser interners, CFG
+storage, and backend MIR remain optimized local representations. A query may
+materialize a fresh local epoch from canonical dependencies, compute one
+artifact, export its canonical result, and discard the local epoch. There is no
+whole-reachable-program semantic epoch whose reconstruction gates downstream
+parallel work.
+
+### 6. Query and projection boundaries
+
+The intended graph is conceptual; exact Rust names may differ:
+
+```text
+module source
+  -> ParseModule
+  -> ModuleIndex / ModuleRir
+  -> LookupName / ResolveImport
+
+stable definition or instance
+  -> Declaration / Signature / ConstValue
+  -> Body
+       -> CanonicalBody
+       -> BodyReferences
+
+root set + BodyReferences
+  -> Reachability
+  -> demanded TypeFacts / Layout / CallAbi / DropGlue
+  -> Cfg
+  -> OptimizedCfg
+  -> optional LoweredMir
+  -> CodegenUnit
+
+sorted reached CodegenUnits + runtime requirements
+  -> ProgramImagePlan
+  -> fresh internal link initially
+  -> incremental internal link later
+```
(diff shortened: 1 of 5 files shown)


docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+title: "Parallel demand-driven incremental compilation"
+status: proposal
+tags: [architecture, compiler, incremental, parallelism, codegen, linker, performance]
+feature-flag: null
+created: 2026-07-18
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["ADR-0045", "ADR-0050", "ADR-0051", "ADR-0052", "ADR-0053", "ADR-0055", "ADR-0058", "ADR-0061", "RUE-328", "RUE-812"]
```
steveklabnik: The Status section declares supersession of ADR-0045 and ADR-0053 in prose, but the frontmatter never says so — there is no `supersedes:` key, only `relates`. `scripts/validate-adrs.py` already understands and resolve-checks the relation keys (`supersedes`, `superseded-by`, `amends`, `amended-by`), so making the relationship machine-readable is free: `supersedes: ["ADR-0045", "ADR-0053"]`.

Related bookkeeping: nothing in the validator enforces reciprocity, so it's worth stating in the Status section that *acceptance* of this ADR carries the concrete obligation to flip 0045 and 0053 to `status: superseded` with `superseded-by:` back-pointers. Otherwise that step is easy to lose, and the index would keep presenting 0045/0053 as fully authoritative after this ADR replaces parts of them.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+its exclusion of cross-request and cross-invocation incremental state. It
+retains and restates ADR-0045's language-semantic rule that observable semantic
+checking and emitted code are selected by explicit roots. It also supersedes
+ADR-0053's single-threaded, one-selected-key execution boundary and its decision
+to end the query database before code generation. The attempt, diagnostic,
+dependency, current-versus-last-good, and fail-closed publication invariants
+from ADR-0053 remain requirements.
+
+Until this proposal is accepted, ADR-0045 and ADR-0053 remain authoritative and
+their metadata is unchanged.
```
steveklabnik: This retained-invariant list is materially narrower than ADR-0053's actual contract. 0053 enumerates **19** invariants (0053:668–711); this sentence names five groups. Not named anywhere in the Status section: #4 publication identity, #16 bounded ownership, #17 single computation path, and — most important for this proposal — **#18 observational equivalence** (cold-vs-reused determinism), which is exactly the property parallel execution puts under the most pressure. 0053's cold-vs-reused equivalence argument (0053:588–621) is built on single-threaded validation; a parallel engine has to re-establish it from scratch, and the differential oracle (`crates/rue-compiler/tests/differential_oracle.rs`, which compares down to per-function assembly and executable bytes) is the artifact that will hold you honest.

Also, "duplicate in-flight work changes from a rejected attempt to a joined request" is a *modification* of invariant #9 (abort fidelity: DuplicateInFlight publishes no terminal, never advances last-good), not a retention of it — and 0053 explicitly deferred join semantics to future work (0053:803), so this ADR is the first document to define them.

Since the Status supersession statement is what future readers will hold other ADRs against, I'd enumerate explicitly which of the 19 remain binding verbatim, which are superseded, and which are restated in modified form.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+Semantic checking, CFG construction, and code generation are per declaration or
+per concrete function instance. Item-level error-tolerant parsing is separate
+future work and is not required by this ADR.
+
+### 2. Inputs are immutable revisions
+
+The query database retains immutable input revisions. A request pins exactly one
+revision for its lifetime. Publishing a source edit, target/configuration edit,
+or host import observation creates a successor revision; it does not mutate the
+inputs observed by already-running work.
```
steveklabnik: "A request pins exactly one revision for its lifetime" collides with §7's "controlled retry across the successor revision." Trace a cold build: the root request pins R0, which contains no import observations. Every newly demanded import publishes observations into a successor revision R1..Rn. Either (a) the request's lifetime spans revisions — contradicting this sentence — or (b) the request restarts once per successor revision, i.e. O(transitive imports) restarts, each re-validating the already-computed prefix. That is quadratic validation work on exactly the build you most want to be cheap, and red/green validation is not free at that multiplicity.

The current protocol sidesteps this by driving a whole frontier per round — ADR-0051's fixed point publishes a *plan's worth* of observations per staging snapshot, not one revision per file read. Three things I think this section needs to pin down:

1. What a "request" is across missing-input fulfillment — e.g. "a request is a sequence of pinned attempts; the pin invariant is per-attempt," with join eligibility defined across that boundary.
2. That demand fulfillment is batched: the coordinator accumulates a round of deduplicated demands and the host publishes one successor revision per round (0051's fixed-point shape), so cold-build revision count scales with import *depth*, not import *count*.
3. That 0051's observation-reuse epoch rule (observations reusable only within one immutable discovery epoch, revalidated absent a watch/read-policy token — 0051:394–401) carries over to successor-revision validation, so republishing an unchanged observation into Rn+1 does not spuriously invalidate work pinned at Rn.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+The request coordinator deduplicates demands from concurrent queries. The host
+performs stable reads and policy checks, then publishes accepted, absent, denied,
+unreadable, ambiguous, or canceled observations into a successor immutable
+revision. Work pinned to the previous revision does not observe an in-place
+mutation. Candidate precedence and provenance remain identical under serial and
+parallel fulfillment.
+
+The implementation may use suspension/joining or controlled retry across the
+successor revision. It may not let semantic queries access the filesystem or
+maintain a second import graph outside the canonical query database.
```
steveklabnik: §7 quietly rewrites more of ADR-0051 than the Status section admits — the Status only declares supersession of 0045 and 0053, and 0051 sits in `relates` as if untouched. But 0051's normative protocol is: the compiler emits an ordered discovery *plan*; the host executes only that plan; parse is all-or-nothing per attempt ("if any module fails lexing or parsing, abort the attempt… There is no partial import-site artifact and no discovery plan from a failed parse," 0051:156–159); and semantic analysis is gated on a *closed, valid* `CanonicalImportGraph` (fail-closed at the graph, 0051:319–376). ADR-0053 reinforces that discovery is an external-input protocol, "not another query engine" (0053:473–492).

This section replaces the fixed-point-closure model with per-query lazy demands, and §1 replaces whole-graph parse validity with per-demanded-module validity ("a demanded module must be syntactically valid as a whole" — per module, not per graph). I think the demand-driven direction is right, but a future reader currently cannot tell which document wins on the points where they differ.

Suggest either adding 0051 to the supersedes/amends set with an explicit survivorship list — candidate precedence, provenance, the read-policy/observation-ledger/accepted-read-manifest separation, and the typed observation outcomes survive; the fixed-point closure, whole-graph parse abort, and graph-gated sema are replaced — or reworking §7 to stay inside 0051's plan-owned model. The last paragraph's "may not maintain a second import graph outside the canonical query database" is a good constraint, but it's not a substitute for saying which 0051 provisions still bind.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+
+The body result includes a canonical body plus a `BodyReferences` projection
+containing the stable callable/specialization/type/module/glue dependencies
+discovered after current-target comptime evaluation. Ordinary call recursion is
+therefore a legal cycle in the reachability graph, not a query dependency cycle.
+
+The reachability coordinator expands reference projections from the root set,
+deduplicates stable identities, and schedules newly ready work. It can discover
+and analyze independent callees in parallel. Its reached-set stamp depends on
+reference projections, not on unrelated instruction content inside bodies.
```
steveklabnik: The reachability coordinator is the least-specified load-bearing piece of the design, and it's the one with a known-hard incremental algorithm underneath. Three concerns:

**1. Is it a database citizen or not?** §6's conceptual graph lists `Reachability` as a query node, but this section describes a "coordinator" that expands frontiers and schedules work — which reads like a mechanism *outside* the memo database. §7 forbids "a second import graph outside the canonical query database," and 0053's invariant #17 (single computation path) points the same way; a call-graph coordinator with its own state is exactly that shape unless the ADR says otherwise. A fixpoint whose dependency set is discovered by its own evaluation is also the hardest kind of node to make red/green — the ADR should state how the coordinator's observed edges are recorded and validated, or explicitly carve it out as a non-query with its own invariants.

**2. Incremental maintenance under deletion.** "Its reached-set stamp depends on reference projections" means any edit that changes any reached body's `BodyReferences` reds the single reached-set stamp. Additions are cheap monotone delta-expansion, but *removals* are the classic incremental transitive-closure problem: when an edge disappears you cannot locally decide whether a formerly-reached identity is still reached — you re-derive from roots, or maintain counting/tracing structures. With a tens-of-milliseconds warm-edit target, O(reached) re-derivation on every reference-changing edit (a very common edit class) may dominate the budget. Phase 7 should name the maintenance strategy or carry an explicit measurement gate for it, and it's worth considering per-identity membership projections so `ProgramImagePlan` and `DropGlue` demand don't all hang off one global stamp.

**3. What does a failed body publish?** If `Body(f)` has semantic errors, does it still publish a (partial) `BodyReferences` projection so errors in its callees are also observable, or does expansion stop at the failure? Either choice is defensible, but the observable-diagnostic closure, the "unreachable declaration contributes no diagnostic" acceptance test, and error-cascade UX all depend on which one you pick — it should be pinned down here, not discovered during Phase 7.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+
+The stable identity domain covers:
+
+- named definitions and named nominal types;
+- free functions, methods, associated functions, and destructors;
+- concrete generic/comptime specializations with canonical type/value
+  arguments;
+- anonymous structs/enums and their methods/destructors, keyed by their stable
+  producing definition or specialization, source anchor, and canonical
+  arguments;
```
steveklabnik: "Source anchor" appearing inside stable identity keys is a classic incremental-compilation trap, and as written the ADR doesn't defend against it. If an anchor is any absolute source coordinate (byte offset, line), then inserting a blank line or comment at the top of a module shifts the anchors of everything below it, and two bad things follow:

1. **Identity churn:** every anonymous nominal below the edit gets a new identity → new symbols, new `CodegenUnit` fingerprints → a spurious `ProgramImagePlan` delta for a whitespace-only edit. That's precisely the edit class the warm-rebuild target cares about.
2. **Red/green churn:** §3 says the red/green comparison covers "outcome *and diagnostic batch*" (lines 178–181), and §4 sorts diagnostics by source anchor (202–203). A retained warning whose anchor shifted makes the batch unequal → red → downstream invalidation, again for a semantically empty edit.

Two requirements would close this: (a) anchors used in identity must be stable under unrelated edits — e.g. a structural ordinal within the producing definition, or a definition-relative rather than module-absolute coordinate; (b) either diagnostic-batch equality is computed modulo position, or positions live in a side channel excluded from the red/green comparison (the rustc span-untracking approach).

Related gap in the acceptance list (lines 553–578): it tests unreachable-source edits and callee-implementation edits, but not the *position-shift* edit. I'd add: "inserting a line above N definitions changes no stable identity, content fingerprint, codegen unit, or plan entry; only position metadata." That's arguably the most common warm edit a user will ever make, and nothing currently forces the implementation to get it right.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+form directly in the eventual incremental path.
+
+Object-file encoding is a projection from `CodegenUnit` for the system linker,
+object presentation, and compatibility testing. The internal path does not need
+to serialize each function to ELF/Mach-O and immediately parse it back.
+
+Machine-code keys include the optimized CFG, target architecture/OS and code
+model, relevant ABI/layout facts, runtime ABI manifest version, backend/schema
+epoch, and only the strings/data actually referenced by that unit. Linker mode
+is not a codegen input.
```
steveklabnik: "Machine-code **keys** include the optimized CFG …" contradicts §3's definition of a logical key ("a semantic entity plus explicit configuration. … A logical key never embeds terminal stamps from its dependencies," lines 149–153). If the key literally contains CFG content, then memo lookup requires computing the CFG first, and the model silently becomes a content-addressed cache — a different beast (no red/green needed, different eviction, no stable node to hang attempt history on).

I assume the intent is: logical key = `(FunctionInstanceKey, target/OS/code-model, opt level, ABI/layout/backend epochs)`, with the optimized CFG and the referenced strings/data as *recorded dependencies* whose content feeds the unit's **fingerprint**. If so, this paragraph should say "inputs"/"dependencies" rather than "keys," and the ADR should define the key-vs-fingerprint distinction once: fingerprint is derived metadata consumed by the linker delta, never part of memo identity. Right now §3, §11, and the `CodegenUnit` sketch ("identity and content fingerprint") each gesture at this without one authoritative statement, and Phase 0/1 implementers will need it unambiguous.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+
+### 12. `ProgramImagePlan` preserves the incremental-linking seam
+
+A deterministic `ProgramImagePlan` contains the sorted reached `CodegenUnit`
+identities/fingerprints, entry point, target/object format, runtime ABI/archive
+identity, required runtime symbols, and final user-visible warnings. It contains
+no external system-toolchain state.
```
steveklabnik: "…and final user-visible warnings" doesn't belong in `ProgramImagePlan`. §13 establishes that diagnostics belong to *requests*, and the plan is defined as the artifact a future incremental linker diffs. Embedding warnings means a warning-text change — or, per my comment on §5, a mere position shift of a retained warning — produces a plan delta with zero link-relevant content, and every future plan-comparison consumer must know to ignore the warning component. That's fingerprint pollution on the one seam this ADR works hardest to keep stable.

Suggest: the plan carries strictly link-relevant facts (sorted unit identities/fingerprints, entry point, target/object format, runtime ABI/archive identity, required runtime symbols), and warnings flow to the request through the §13 diagnostic channel like every other diagnostic. If the intent was "the executable request's observable output includes warnings," that's a property of the *request adapter*, not of the plan artifact.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+  anchor, and producer-defined order after collection;
+- reached codegen units and image atoms are sorted by stable identity before
+  image planning; and
+- reductions of work metrics are deterministic.
+
+The runtime owns one concurrency budget. Query-level parallelism and parallel
+algorithms inside a query must use structured permits from that budget rather
+than nesting uncoordinated global thread pools. The first implementation may run
+the same interfaces with one worker. Correctness and artifact identity must be
+identical for one worker and many workers.
```
steveklabnik, line 271: Two gaps in the concurrency-budget story:

**1. Claim-or-join + one bounded budget can deadlock without any dependency cycle.** If every permit-holding worker is blocked *joining* an in-flight key, and the claimed computations need permits for their own dependencies, the pool wedges whenever the budget is smaller than the depth of the active dependency chains. The §4 wait graph detects cycles; it does not detect permit starvation — the blocked waiters aren't waiting on each other, they're waiting on tasks that can't get a thread. The standard escapes are: a joining waiter releases its permit while parked, or donates its thread to the claimed computation (the salsa/rayon approach). One of these should be a stated requirement, and Phase 0's gate should include "no deadlock under budget exhaustion and adversarial join schedules" — right now Phase 0 proves claim-or-join works, not that the scheduler guarantees progress.

**2. The existing global Rayon pool is unaccounted for.** The compiler already runs process-global Rayon parallelism today: `configure_thread_pool` builds the global pool from the driver (`rue-compiler/src/lib.rs:236-243`, `rue/src/main.rs:976`), and per-function CFG build/optimize plus both backends are `par_iter` inside the (serial) semantic/backend tail (`queries.rs:203-204`, `backend.rs:57,103`). Those are exactly the "parallel algorithms inside a query" this paragraph says must use structured permits — but the ADR never names them or the migration. Until they move onto the budget, query-level parallelism stacked on the global Rayon pool *is* the uncoordinated nesting this paragraph forbids. Worth an explicit sentence (and probably a phase assignment) so this doesn't survive as an accidental violation. Note also that `configure_thread_pool` is a Keep-class stable operation in ADR-0061's inventory, so changing its meaning touches the facade contract.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+- which runtime functions, user functions, specializations, destructors, and
+  drop-glue units enter the program image; and
+- which codegen units are required before linking.
+
+The scheduler may speculatively evaluate a pure query which is not yet in the
+root closure. A speculative result may populate the memo database, but its
+diagnostics, warnings, or emitted unit do not become observable until a rooted
+query reads it. This separates language semantics from scheduling policy and
+allows safe background work without turning `check-all` into the default build
+mode.
```
steveklabnik, line 143: Speculation and §7's host-demand protocol interact in a way this section doesn't address: may a *speculative* query emit a missing-input demand? "Pure query" suggests no side effects, but under §7 an import lookup on a module whose observation isn't in the pinned revision generates a typed demand that the host fulfills with real filesystem reads under real read policy. If speculation can do that, then "safe background work" causes host I/O, read-policy checks, and observation-ledger growth for modules the rooted program never imports — observable side effects, a bigger accepted-read manifest, and a policy surface (should a *denied* read triggered by speculation surface anywhere?).

Suggest an explicit rule: speculative execution never emits host demands — a speculative query that hits a missing input parks or abandons, and only rooted (or explicitly policy-authorized, e.g. idle-time cache warming per Future Work) work may drive the import protocol. Wherever the rule lands, the acceptance bullet "importing one std submodule does not read unrelated std modules" should be strengthened to hold *including under speculative execution*, the way the unreachable-declaration bullet already is.

---

docs/designs/0063-parallel-demand-driven-incremental-compilation.md
```diff
@@ -0,0 +1,712 @@
...
+- [ ] **Phase 0: Runtime prototype and benchmark gate.** Prove exact-key
+  claim-or-join, different-key parallel execution, red/green propagation,
+  deterministic diagnostics, cancellation/revision isolation, exact cycle
+  handling, and bounded retention on representative query shapes. Compare an
+  in-house evolution with a query-library prototype if substrate choice remains
+  open.
+- [ ] **Phase 1: Revisioned keyed database.** Replace family-selected mutable
+  state with immutable revisions, per-key nodes, joined waiters, task-scoped
+  dependency recording, and single-worker scheduling through the parallel-ready
+  interface.
```
steveklabnik: Phase 1's blast radius deserves an honest sentence here. The state being replaced is concentrated (`SelectedQueryState`/`selected`/`last_good` in `typed_query_store.rs`, the `TypeId`-keyed `computing`/`aborted_computations` in `query_graph.rs` — ~2,200 engine LOC), but all ten query families and essentially all of `session.rs`'s ~8,300 non-test orchestration lines call through `request_selected`/`begin_selected`/`publish_selected*`. As written, Phase 1 is a big-bang cutover of every family's call discipline at once — the largest, least-reviewable diff in the whole plan, and it lands *before* the per-family migrations that would exercise it.

Consider requiring the new per-key engine to expose a compatibility shim over the selected-state API so families migrate one at a time, each behind the cold-vs-reused differential oracle (`tests/differential_oracle.rs` already compares down to per-function assembly and executable bytes — it's the right harness to hold constant through the swap). That also gives Phase 1 a natural completion proof: the shim's deletion.

---

steveklabnik (commented):
Adversarial review as requested, with a focus on the parallel/incremental architecture and the incremental-linking seam. I verified the ADR's claims about the current implementation against trunk source and the prior ADRs before commenting.

**What checks out.** The Motivation's characterization of today's compiler is accurate on every point I verified: the serial selected-state engine and `DuplicateInFlight` rejection (`typed_query_store.rs:472-490,661-686`), family-`TypeId` cycle/active tracking (`query_graph.rs:62-128`), eager parse/merge/RIR/declaration processing, named destructors as implicit roots with whole-type-pool drop-glue synthesis (stated verbatim at `rue-air/src/sema/analysis.rs:1018-1021`, `drop_glue.rs:44-79`), the fused semantic+CFG terminal (`CanonicalSemanticOutput` owning `FunctionWithCfg` + durable CFGs), and the uncached backend tail. The big design calls also look right to me: roots-vs-speculation separation, `BodyReferences`/reachability decoupling so callee edits don't invalidate callers, `CodegenUnit`/`ProgramImagePlan` as the linking seam with object encoding demoted to a projection, and red/green publication. This is a strong foundation document.

**Where I pushed.** The inline comments carry the substance; the themes:

- **Spec-consistency:** the §2 "pins exactly one revision for its lifetime" invariant contradicts §7's retry-across-successor-revision protocol, with a quadratic cold-build failure mode if fulfillment isn't batched; §11's "machine-code keys include the optimized CFG" contradicts §3's logical-key definition (key vs. fingerprint needs one authoritative statement); warnings inside `ProgramImagePlan` contradict §13's "diagnostics belong to requests" and pollute the linker-delta fingerprint.
- **Underspecified load-bearing pieces:** the reachability coordinator (database citizen or not; incremental maintenance under edge *deletion* — the classic incremental-transitive-closure problem — versus the tens-of-ms target; what a failed body publishes), and scheduler progress guarantees (claim-or-join + one bounded permit budget can deadlock with no dependency cycle; the existing global Rayon pool in CFG/codegen `par_iter`s is unnamed and currently *is* the nesting §4 forbids).
- **Edit-stability:** source anchors inside stable identity keys plus diagnostic batches inside the red/green comparison make a whitespace-only edit at the top of a module churn identities, fingerprints, and the plan — the acceptance list has no position-shift test, and that's the most common warm edit there is.
- **ADR bookkeeping:** §7 materially rewrites ADR-0051's fixed-point, plan-owned, fail-closed-at-the-graph discovery protocol but the Status section only supersedes 0045/0053; the frontmatter has no `supersedes:` key though the validator resolve-checks one; the retained-invariant list names 5 groups of ADR-0053's 19 invariants and is silent on the two that parallelism stresses most (single computation path, cold-vs-reused observational equivalence — and "join" is a modification of invariant #9, not a retention).
- **Migration risk:** Phase 1 is a big-bang cutover of all ten families' call discipline at once; a per-family shim held against the existing cold-vs-reused differential oracle would make it reviewable.

Nothing here argues against the direction — the issues are about making the contract precise enough that Phase 0/1 implementers and the eventual incremental-linker ADR can't diverge on interpretation. (Also noting the ADR-0062→0063 renumber after the trunk collision is already handled in the current push.)

---

PR #1830, opened 2026-07-18
RUE-320: scope linker to remote execution

steveklabnik:
## Summary
- add a remote-execution constraint to the BuildBuddy execution platforms and preserve the base exec marker
- make the C++ tools provider an exec-dep so the real configured graph selects `cc` remotely and retains `clang++` natively
- add a merge-group-only, cache-disabled remote execution canary with hybrid fallback disabled
- document supported opt-in `--prefer-remote` behavior

## Why
A direct `select()` on `system_cxx_toolchain` does not see execution-platform constraints; the toolchain itself remains output-target configured. Moving the selectable linker into `CxxToolsInfo`, which `cxx_tools_info_toolchain` consumes as an exec-dep, puts the selection in the resolved execution configuration.

The pinned Buck2 prelude now adds `-fuse-ld=lld` on Linux even with `cc`, so the worker only needs the ubiquitous driver and Rust's bundled lld does the actual link.

## Validation
- configured-provider audit: native execution resolves `clang++`; fake Linux BuildBuddy execution resolves `cc` with `-fuse-ld=lld`
- cache-disabled native `//crates/rue:rue` build: 231 local actions, green
- all crate unit targets: 34 passed, 152 local actions
- wrapper-script tests: 69/69 checks
- build-sharing tests: 29/29 checks
- pinned actionlint + ShellCheck: green
- merge-group canary will run a real cache-disabled remote build and reject remote fallback/local-only false greens

Fixes RUE-320

### DIFF
--- a/.buckconfig.local.example
+++ b/.buckconfig.local.example
@@ -14,10 +14,8 @@
 # With this config, the repository's `./buck2` wrapper adds `--prefer-local` to
 # normal execution commands: the remote action cache is queried first, then
 # misses execute locally. Adding an explicit execution-mode flag overrides the
-# wrapper, but full remote execution is not supported until RUE-320. It was
-# previously demonstrated for compilation, but complete builds still need the
-# platform-scoped linker fix. Remote execution is
-# possible because the rust toolchain ships its own tree (rustc finds its libs via
+# wrapper. Remote execution is possible because the Rust toolchain ships its own
+# tree (rustc finds its libs via
 # $ORIGIN under RE) and links via `cc`+lld; the remote_cache platform pins the
 # rbe-ubuntu22-04 container image (Python 3.10 for the prelude's rustc wrapper).
 #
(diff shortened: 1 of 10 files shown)


steveklabnik:
The first merge-group canary executed 344/344 actions remotely with no local fallback, then BuildBuddy OOM-killed a rustc action. The amended commit requests 4 GB per remote action via the documented EstimatedMemory platform property; the next merge-group run will validate that sizing.

steveklabnik:
The second merge-group run proved full RE (364/364 remote, 0 local) but exposed a cache-only/native interaction on Linux ARM64: the cache-only platform carried the remote-tools constraint even though --prefer-local executed misses natively, selecting cc and reproducing collect2: cannot find ld. The amended branch marks only the explicit full-remote platform. Configured-graph audits now resolve cache-only ARM64 to the native linker and full remote to cc.

PR #1891, opened 2026-07-21
RUE-1085: hash family memo lookup and defer display identity

steveklabnik:
First step of the replanned RUE-1083/ADR-0063 sequence: make family memo lookup amortized O(1) and remove eager display-identity allocation from the query hot path, ahead of per-lookup query granularity.

## What changed

- **Hashed typed-key node index.** `QueryFamily`'s node lookup was a deliberate Phase-1 O(n) `Vec` scan; with the 65,536-entry body/declaration retention it became an O(n²)-shaped term at large-program scale. The index is now `HashMap<K, Arc<Node<K, V>>>` under the std keyed (SipHash) hasher — typed equality stays authoritative, hash collisions resolve through `Eq` (covered by a forced-collision test with a constant-`Hash` key). Nodes carry their key so eviction reverse-locates them in O(1) behind an `Arc::ptr_eq` guard, preserving incarnation semantics; eviction *order* remains owned by the retention queue, which the `Vec` never encoded.
- **`QueryKey: Hash` with a full implementor audit** — most keys derive it; `CompatibilityKey`, `InvalidationPlanQueryKey`, and `SourceMetadata` get manual impls consistent with their structural `Eq` (documented inline).
- **Lazy display identity.** Nested query requests formatted a `Debug` display string on every call, including memo hits. Dependency/red-green validation keys on node incarnation, not the string, so display identity is now materialized lazily: the terminal path reuses the identity built once at node creation; only cold abort paths format keys.

## Measurements (isolated `rue-query-bench --memo-hit` mode)

| retained nodes | before (linear) | after (hashed) |
|---|---|---|
| 1,000 | 4,404 ns/hit | 2,382 ns/hit |
| 10,000 | 22,587 ns/hit | 2,768 ns/hit |
| 100,000 | — (O(n²) populate) | 3,348 ns/hit |

~8× at 10k; flat through 100k, confirming O(n) → amortized O(1).

## Validation

- rue-query suite 32/32 (three new focused tests: forced-hash-collision resolution via `Eq`; evict-then-recreate mints a new incarnation; six concurrent same-key requests join one computation with `claims == 1`).
- rue-compiler 665/665, rue 172/172, quick suite 35/35, fmt clean, rill end-to-end compile sanity.

Known follow-up (out of scope, noted on the Linear issue): node *creation* still pays an O(n²) incarnation-registry `retain` sweep; irrelevant to memo hits but a candidate before per-lookup granularity work.

Fixes RUE-1085

---
_Generated by [Claude Code](https://claude.ai/code/session_019h99YFzR3fuDy2phd2vU9r)_

### DIFF
--- a/crates/rue-compiler/src/diagnostic_attempt_store.rs
+++ b/crates/rue-compiler/src/diagnostic_attempt_store.rs
@@ -99,7 +99,7 @@ pub struct FrontendDiagnosticSnapshot {
 
 /// Producer inputs that can change presentation without changing the stable
 /// source revision or public query stage.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) enum DiagnosticAttemptProvenance {
     Canonical,
     Presentation(Arc<[ModuleId]>),
--- a/crates/rue-compiler/src/revisioned_query_database.rs
+++ b/crates/rue-compiler/src/revisioned_query_database.rs
@@ -77,9 +77,16 @@ impl<K: PartialEq> PartialEq for CompatibilityKey<K> {
 
 impl<K: Eq> Eq for CompatibilityKey<K> {}
 
+impl<K: std::hash::Hash> std::hash::Hash for CompatibilityKey<K> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // Agrees with the `PartialEq` above, which compares only `key`.
+        self.key.hash(state);
+    }
+}
+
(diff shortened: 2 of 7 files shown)


steveklabnik (commented):
codex: Reviewer sign-off on ae67e174. I found no actionable correctness or design issues after checking the typed-key Hash/Eq contract, memo-lock ordering and eviction, lazy display identity, and the benchmark shape; an independent adversarial pass also found no issues. Local `scripts/rue quick` passed 35/35. Because author and reviewer share this GitHub account, GitHub cannot record a formal self-approval; treat this comment review as my approval. The only remaining CI item is the large-examples compile guard.

PR #1895, opened 2026-07-22
RUE-1087: lease observed terminals for the lifetime of their request

steveklabnik:
Third step of the replanned RUE-1083/ADR-0063 sequence: make bounded retention correctness-safe — a terminal the current computation still needs can never be evicted. This revision incorporates both blockers from the pre-review of the initial diff, and both blocking findings from the on-PR review of 67ad09b2.

## Design

The lease extends the runtime's existing pin ownership model (no parallel cache): `TerminalPin` is the lease object; the type-erased `Task` carries the lease set (dedup by node incarnation + stamp, pins held across every family the task touches), and since nested `context.query` calls share the root request's `Task`, nested observations inherit the root lease automatically — the observed Caldera failure was precisely a nested projection observing on behalf of the root.

**Atomic protection handoff (pre-review finding 1).** Lease acquisition is atomic with retention at all three sites: publication pins under the node-state lock, after the terminal state is set but before it is exposed or enqueued to any enforcer; a joining waiter's protection transfers into a pin *before* the waiter count decrements — including the last waiter, so no unprotected instant exists; reuse candidates are pinned while still retained under the node lock and hold that pin across recursive validation, transferring it into the request lease on success. Speculative validation-driven publications (`observe_result == false`) take no pin and stay evictable.

**Attempt-carried result lease, released at selection (pre-review finding 2 + review finding).** `QueryRequestAttempt` carries a private type-erased result lease acquired while the producing task is still alive, bridging the gap between request completion and `QuerySelection::publish` registering successor protection. The bridge ends the instant selection succeeds: `RevisionedFamily::select` calls the attempt's explicit `release_result_lease` immediately after `publish` establishes the successor root — a pure narrowing with no unprotected instant — so the compiler's bounded ledger of completed attempts (up to 256) no longer pins selected terminals for the record's whole life. Aborted/setup-failed attempts and never-selected callers release at teardown as before.

**Linear task-lease teardown (review finding).** A rooted request may hold thousands of observation pins in one family; releasing them one `enforce_retention` scan at a time was quadratic exactly at request completion. `TaskLeases` teardown now decrements every held pin first (decrement-only release; no still-leased terminal is ever unprotected), then runs exactly one enforcement pass per distinct family, deduplicated by the stable `FamilyInner` identity. A `deferred` flag suppresses the released pin's own `Drop`; ordinary per-pin users keep unchanged semantics, and cancellation takes the same path. A new `retention_enforcements` metric makes the linearity observable.

Pressure policy: a fully protected family grows past its soft bound and records a `retention_growth` metric — never a silent eviction of a needed terminal. The interim 65,536-entry retention constants remain until database-owned rooted membership (RUE-1028 resumption) replaces fixed caps; leases remove the hazard, not the caps.

## Tests

Six adversarial acceptance tests with tiny retention caps (long in-progress rooted traversal; the Caldera early-producer projection shape; cancellation releasing request-scoped leases; successful-revision promotion; successor-revision reclamation; speculative terminals evictable — publication alone does not pin), plus four deterministic barrier-based race tests using a test-only interposition hook at the exact handoff windows: `publish_window_pin_precedes_evictability`, `join_window_last_waiter_handoff_leaves_no_unprotected_instant`, `reuse_window_candidate_survives_between_discovery_and_lease`, `promotion_gap_result_lease_bridges_until_selection`. Three more from the review round: `batched_task_lease_release_enforces_once_per_family_and_converges` (64 one-family pins → exactly 1 enforcement pass, convergence to cap, pinned sentinel survives), `selected_result_releases_bridge_lease_before_attempt_drop`, and the compiler-level `selecting_a_result_releases_the_attempt_bridge_lease_before_ledgering`. Every race/regression test was verified to fail with its corresponding fix reverted.

## Validation

rue-query suite 44/44; `revisioned` compiler suite 60/60; scaling-harness smoke 8/8; fmt clean; warning-clean builds.

Fixes RUE-1087

### DIFF
--- a/crates/rue-query/src/lib.rs
+++ b/crates/rue-query/src/lib.rs
@@ -467,13 +467,22 @@ pub struct QueryRequestAttempt<V> {
     /// its request-scoped leases) were still alive, so the published result stays
     /// retained across the gap between this request completing and the caller
     /// registering a successor protection (a session/revision selection root via
-    /// [`QuerySelection::publish`]). The lease releases when the attempt drops —
-    /// which callers keep alive until after selection — giving continuous
-    /// protection with no window in which the result can be evicted and a later
-    /// selection pin a detached recompute. `None` for aborted or setup-failed
-    /// requests, which carry no terminal. Type-erased over the family key so the
-    /// attempt need not name `K`.
-    result_lease: Option<Box<dyn ObservedLease>>,
+    /// [`QuerySelection::publish`]). Its sole job is to bridge that gap.
+    ///
+    /// Once a successor protection exists, the bridge is redundant and must end
+    /// promptly: an attempt record may be held for a long time (the compiler
+    /// ledgers up to 256 completed attempts), and a lingering bridge pin would
+    /// keep an otherwise-evictable terminal retained for the record's whole life.
+    /// [`release_result_lease`](QueryRequestAttempt::release_result_lease) ends it
+    /// explicitly the instant selection has pinned the same terminal — a pure
+    /// narrowing of protection with no window in which the result is unprotected.
+    /// Callers that never register a successor simply drop the attempt, releasing
+    /// the bridge at teardown. Held behind a `Mutex` so the release is a `&self`
+    /// operation on the possibly-shared attempt record.
+    ///
+    /// `None` for aborted or setup-failed requests, which carry no terminal.
+    /// Type-erased over the family key so the attempt need not name `K`.
+    result_lease: Mutex<Option<Box<dyn ObservedLease>>>,
 }
 
 impl<V: fmt::Debug> fmt::Debug for QueryRequestAttempt<V> {
@@ -489,7 +498,7 @@ impl<V: fmt::Debug> fmt::Debug for QueryRequestAttempt<V> {
             .field("inputs", &self.inputs)
             .field("work", &self.work)
             .field("nested_attempts", &self.nested_attempts)
-            .field("result_lease", &self.result_lease.is_some())
+            .field("result_lease", &lock(&self.result_lease).is_some())
             .finish()
     }
 }
@@ -548,6 +557,25 @@ impl<V> QueryRequestAttempt<V> {
     /// Origin terminal revision for reuse/join provenance.
     pub fn origin_revision(&self) -> Option<Revision> {
         self.terminal.as_ref().map(|terminal| terminal.revision())
+    }
+
+    /// Ends the attempt-carried bridge lease now that a successor protection
+    /// holds the result.
+    ///
+    /// Call this immediately after registering a successor protection for the
+    /// terminal — a [`QuerySelection::publish`] root, typically — and never
+    /// before. The bridge lease exists only to keep the result retained between
+    /// request completion and that registration; once the successor pins the same
+    /// terminal, continuing to hold the bridge just pins an otherwise-evictable
+    /// terminal for as long as the (possibly long-lived, ledgered) attempt record
+    /// survives. Releasing here is a pure narrowing of protection: the successor
+    /// still holds the terminal, so there is no instant in which it is
+    /// unprotected. Idempotent, and a no-op for aborted or setup-failed attempts
+    /// that never held a bridge lease. The released pin's own decrement and single
+    /// retention pass run here, outside the attempt lock.
+    pub fn release_result_lease(&self) {
+        let released = lock(&self.result_lease).take();
+        drop(released);
     }
 
     /// Converts this attempt to the legacy terminal-or-abort call shape.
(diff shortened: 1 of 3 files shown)


crates/rue-query/src/lib.rs
```diff
@@ -2383,6 +2607,49 @@ struct Task {
...
+/// Terminals leased for the lifetime of one rooted request (its task).
+#[derive(Default)]
+struct TaskLeases {
+    /// `(node incarnation, red/green stamp)` of every terminal this task has
+    /// already leased, so re-observing a terminal never double-pins it.
+    observed: BTreeSet<(u64, u64)>,
+    /// Live pins, type-erased across families. Dropping the task drops these,
+    /// each of which decrements its terminal's pin count and re-enforces the
+    /// owning family's retention bound.
+    held: Vec<Box<dyn ObservedLease>>,
```
steveklabnik, line 2684: codex: Releasing a large task lease set is quadratic once the live closure grew past the family cap. Dropping this vector drops each `TerminalPin` separately; every `TerminalPin::drop` calls `enforce_retention()`. While the family remains over limit, that pass evicts the one newly unpinned terminal and scans or rotates all still-pinned entries, then the next pin repeats the scan. For N observed terminals in one family this is quadratic work precisely at rooted-request completion—the Caldera-shaped path this change is intended to make practical. Please batch task-lease release by decrementing all relevant pins first and enforcing once per family, or use an equivalent linear design.

crates/rue-query/src/lib.rs
```diff
@@ -1094,6 +1175,18 @@ impl QueryRuntime {
...
                 let origin_request = terminal.origin_request_id();
+                // Carry a live result lease out of the request. The producing task
+                // is still alive here (it drops at the end of this function), and
+                // it still holds its own request-scoped lease on `terminal`, so
+                // `terminal` is currently retained. Pinning it now hands protection
+                // from the task's about-to-drop lease to the returned attempt with
+                // no gap. The caller keeps the attempt alive until after it
+                // registers a successor protection (selection root), closing the
+                // promotion window entirely.
+                let result_lease = family
```
steveklabnik, line 1223: codex: This bridge pin is never transferred or released when selection succeeds; it lives until the entire `QueryRequestAttempt` drops. That is substantially longer in the real compiler: `RevisionedFamily::select` only borrows the attempt, then `attempt_view` embeds its `Arc<QueryRequestAttempt>` in `RuntimeAttemptView`, and `CompilerSessionMetrics` retains those views in the 256-entry attempt ledger. Consequently completed diagnostic/history attempts become independent terminal roots even after `QuerySelection::publish` has installed the successor pin. This contradicts the ADR-0066 completion/release invariant and can keep old terminals pinned until unrelated history eviction. The handoff needs an explicit transfer/release operation, likely requiring a small compiler-side call-site change, so the attempt pin ends immediately after successor protection is established.

steveklabnik:
codex: Reviewed head `67ad09b2`. The atomic publication, join, reuse, and promotion-window handoffs themselves now look sound. I left two blocking inline findings: the attempt-carried bridge pin outlives selection in real compiler attempt-history callers, and task lease teardown is quadratic after growth past a family cap. I did not apply or build the patch, per request; CI remains the build authority.

steveklabnik:
Fable: Both blocking findings addressed in 46959404.

**Quadratic teardown** — `TaskLeases` teardown now releases in two phases: every held pin is decrement-released first (pure narrowing; no still-leased terminal is ever unprotected), then exactly one `enforce_retention` pass runs per distinct family, deduplicated by the stable `FamilyInner` address (kept alive by the enforcer's captured family clone, so no ABA). A `deferred` flag suppresses the released pin's own `Drop`; all ordinary per-pin users keep unchanged semantics, and cancellation takes the same path since it drops the same `Task`. New `retention_enforcements` metric makes linearity observable: `batched_task_lease_release_enforces_once_per_family_and_converges` leases 64 terminals in one family from one rooted request and asserts a delta of exactly 1 (per-pin path measured 64), plus convergence to the cap and survival of an externally pinned sentinel.

**Bridge pin outliving selection** — `QueryRequestAttempt` now exposes `release_result_lease(&self)` (lease moved behind a `Mutex` so release works on the shared/ledgered record; the drop runs outside the lock). `RevisionedFamily::select` calls it immediately after `selection.publish` establishes the successor root, so the handoff is a pure narrowing with no unprotected instant; aborted/setup-failed attempts and never-selected callers keep current teardown behavior. Tests: `selected_result_releases_bridge_lease_before_attempt_drop` (rue-query; also shows the result evicts once the selection drops while the attempt still lives) and `selecting_a_result_releases_the_attempt_bridge_lease_before_ledgering` (compiler; a ledgered completed attempt no longer pins its terminal — pre-fix this reuses, post-fix it recomputes).

All three new tests were verified to fail with their fix reverted. Full `rue-query` suite (44) and compiler `revisioned` suite (60) green, fmt clean, warning-clean builds. Ready for re-review.

---

crates/rue-query/src/lib.rs
steveklabnik: codex: Addressed in `46959404`. I verified that `RevisionedFamily::select` establishes the selection pin before releasing the attempt bridge, and the compiler regression test keeps the completed attempt alive while proving it no longer retains the terminal. The handoff remains continuous and the history-root leak is gone.

crates/rue-query/src/lib.rs
steveklabnik: codex: Addressed in `46959404`. I verified the two-phase teardown decrements every task lease before enforcing once per distinct family. The first retained enforcer keeps the `FamilyInner` allocation alive, so pointer-key deduplication has no ABA window, and concurrent enforcement during narrowing remains safe. The quadratic completion path is removed.

steveklabnik:
codex: Re-reviewed head `46959404`, limiting this pass to correctness and material retention-contract failures. Both prior blockers are resolved: selection now releases the bridge only after successor protection exists, and task teardown batches decrements before one enforcement per family. I found no new correctness blocker in the updated ownership, race, cancellation, or reclamation paths. The two review threads are resolved; this is reviewer sign-off for this head, with CI remaining the build authority.

PR #1898, opened 2026-07-22
RUE-1089: producer-nominal anonymous type identity

steveklabnik:
D1 of the ADR-0066 sequence, landed as a reviewed **partial** RUE-1089 cut: anonymous types are producer-nominal. Two anonymous type expressions denote the same type iff they come from the same producer site under the same static comptime specialization — structural equivalence across producers is gone, and with it the representative/alias reconciliation machinery and its body-restart scheduling. RUE-1089 stays open: one structural remnant remains (below) and its removal is a **required follow-up tracked as RUE-1112**, which now blocks RUE-1089's completion and gates the RUE-1090 measurement.

## Language cut

- `types_equivalent` collapses anonymous arms to exact `AnonymousNominalKey` identity; the cross-producer structural search in `find_or_create_anon_struct`/`find_or_create_anon_enum`, the representative swap, and the recursive structural relation are deleted, along with the whole representative abstraction (`AnonymousNominalIdentitySet`, `canonical_anonymous_aliases`, stable-min selection; `SemaOutput` maps `Type -> AnonymousNominalKey` directly).
- Forwarding (`fn Id(comptime T: type) -> type { T }`) preserves identity; aliases preserve identity; distinct producers with identical shape are distinct types (E0206 on mixing).
- Annotation-position anonymous types (`let x: struct { … }`, returns, intrinsic type arguments) are rejected with a targeted parse error at every entry to the shared type grammar.
- Recursive instantiation that selects a new specialization each step no longer converges by shape: it hits the deterministic specialization-depth limit (E1200).
- Spec: rules 4.14:8/:15/:21/:22/:25 replaced and 4.14:23a added per the staged amendment in ADR-0066; the spec corpus is migrated.

## Single anchor authority + transport (the correctness core)

Anonymous-type identity anchors have exactly one minting algorithm — the shared `rue_rir::anonymous_type_sites` walk. AstGen consumes its output by exact source-span lookup (its own `anonymous_type_anchor`/`structural_path` mint is deleted); the durable fragment evaluator consumes the same sites transported into fragment-local coordinates at construction time (`RawAnonymousSite`), validated ONCE into a keyed index (`TransportedAnonymousSites`, O(log S) lookup). The table is atomic: missing, duplicate, kind-mismatched, or divergent entries are a typed E9000-class invariant failure — validated against the producer's complete minted-nominal set BEFORE the producer's `ComptimeCall` terminal publishes, so no nominal/member/alias/cache entry ever exists under a corrupt anchor. No structural recomputation fallback anywhere; genuinely unavailable input aborts/retries at the query boundary.

## Symbol stability

Anonymous symbol names derive from a stable digest of the producer-nominal key over the canonical **logical module path** (never request-local FileIds — a permuted-FileId test pins this), so independent cold compiles, warm/fresh pairs, and schedule permutations emit identical symbols. A shared exact-key collision registry fronts both anonymous struct and enum registration: same digest with a distinct key is a typed fail-closed error with zero publication (never a panic, never silent EnumId reuse); forced-collision tests cover both insertion orders and same-key reuse. Named-type symbols are always qualified (`{name}${file}`); the collide-then-qualify gate is deleted; lang-item builtins stay unqualified for durable import.

## Remaining structural remnant — required follow-up (RUE-1112, blocker)

The bare-fallible-intrinsic `?` path still selects its `Option` via `find_compatible_anon_enum` — a structural shape-scan with stable-min selection. This is NOT an accepted endpoint: it preserves exactly the locality dependence this cut removes, and an unrelated compatible producer can change the selected nominal. Review established the correct design (recorded on RUE-1112, now High/blocking RUE-1089): rule 4.15:9's intrinsic owns an exact compiler-known `Option(payload)` operand, rooted pre-AIR through the semantic nucleus into a narrow well-known-type registry AIR only consumes — no lexical import, no pool scan, no structural fallback, and user-defined return `Option`s stay independent (the freestanding case is answered, not sidestepped). RUE-1090's measurement runs only after that deletion lands. `anonymous_key_cmp` retains one min-selection consumer (this scan) until then; its remaining role after RUE-1112 is deterministic export ordering only.

## Tests

Acceptance corpus: `producer_nominal_acceptance.toml` (14 spec cases) + programmatic tests (criterion→test map in `docs/notes/rue-1089-acceptance-ledger.md`): same-producer distinctness, nesting positions, sibling-reorder/unrelated-edit identity stability, cold/warm/fresh symbol determinism incl. permuted FileIds, the Wrap single-nominal-identity payload regression, host-gated execution on both backend targets, frontend bijection, evaluator correspondence under reversed evaluation order, selected-branch subset, trivia-shift stability, fault injection for all four transport-corruption modes asserting the typed failure as the direct producer-terminal value with zero publication, and forced digest-collision coverage.

## Validation

air 568 · compiler 687 · rir 59 · cfg 209 · codegen 606 · parser 43 · query 44 · full spec 2173 · ui 204 · CLI (incl. try) green · clippy/fmt clean. Direct executions re-verified on the final head: Wrap→42, Box destructor→100,7, distinct producers→E0206, recursion→E1200. Release-build examples: meridian 416s — identical to trunk's 416s (serial A/B), byte-identical program output; caldera exceeds a 900s cap under both trunk and this branch on the reference container (pre-existing host-scale cost; CI's guard is the authority).

Part of RUE-1089 (kept open — completion blocked by RUE-1112).


steveklabnik (commented):
codex: Reviewer sign-off for exact head 9c2ef683 as the partial RUE-1089 cut described in the updated PR body. Themes 1, 2, 4, 5, and 6 were adversarially re-reviewed; focused producer-nominal, RIR anchor, collision, macOS, and direct-terminal checks are green, and required CI is fully green. RUE-1112 remains a real blocker for completing RUE-1089 and running RUE-1090; this PR correctly no longer closes RUE-1089. GitHub cannot record a formal approval because author and reviewer share this account. No further changes requested on this head.

PR #1900, opened 2026-07-22
RUE-1093: draft incremental acceptance contract

steveklabnik:
## Summary

- record the post-D1 artifact, diagnostic, reachability, retention, type/drop, codegen, and image-plan contracts needed before ADR-0063 implementation resumes
- freeze the exact trusted `Option`/`Result` and rooted toolchain-module-demand boundary from RUE-1112
- replan the remaining compiler work into independently keyed phases with explicit deletion guards

## Why

The producer-nominal cut removes the old structural identity machinery, but it does not by itself prove that the remaining compiler pipeline has the right incremental ownership boundaries. This draft makes those boundaries falsifiable before implementation resumes so a new query cannot quietly coexist with an aggregate computation path.

The draft deliberately keeps ADR-0066 in proposal status. It must not be accepted or merged until RUE-1090 and RUE-1092 complete, and RUE-1091 also completes if the mechanical scaling gate activates it. Gate evidence and any resulting amendments will be added to this PR before it leaves draft.

## Review history

An adversarial architecture pass found and corrected three serious issues in the first version:

- the exact `?` and fallible-intrinsic semantics were under-specified
- the canonical `ProgramImagePlan` incorrectly contained predecessor-dependent deltas
- membership removal could be read as releasing retention owned by another live root

## Validation

- `git diff --check`
- source-grounded audit against ADR-0063 and the current compiler ownership boundaries
- adversarial contract review after the fixes above

Fixes RUE-1093

### DIFF
--- a/docs/designs/0066-producer-nominal-anonymous-types-and-incremental-locality.md
+++ b/docs/designs/0066-producer-nominal-anonymous-types-and-incremental-locality.md
@@ -421,10 +421,13 @@ retains the independently useful body references needed for reachability;
 cancellation or incomplete work publishes neither a body terminal nor a
 reference terminal.
 
-Language-owned optionality and fallibility use the exact trusted producers
-`\0rue-std/option.rue::Option` and `\0rue-std/result.rue::Result`. Every
-fallible intrinsic returns an exact specialization of the trusted `Option` in
-every context. An expected type only checks compatibility; it never selects the
+RUE-1112's independent amendment to ADR-0038 establishes that language-owned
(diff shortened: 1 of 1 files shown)


DorianListens:
@steveklabnik youll need to rebase this one to get it to pass ci.

steveklabnik:
Gross, it did a merge instead of a rebase. before this lands we should rebase it to get that merge commit out of here

DorianListens:
Ewwwwww

PR #1903, opened 2026-07-22
RUE-1119: BTD affected-corpus test selection on PR runs

DorianListens:
## Summary

On `pull_request`, select only the heavy corpus suites a change actually affects instead of running the full per-platform matrix. `merge_group` and `workflow_dispatch` keep running the authoritative full corpus.

Per the RUE-1119 design, this uses Meta's Buck Target Determinator (BTD) rather than a hand-rolled `owner()`/`rdeps()` query. The `affected-targets` job dumps the Buck graph at the merge-base and head, feeds both dumps plus Git `--name-status` changes to BTD, and intersects the impacted-target closure with the selectable corpus set.

## Safety model

Under-selection silently drops coverage, so every uncertain path fails open toward running the corpus:

- Out-of-graph or graph-global changes force a full run.
- VCS, DotSlash, Buck, BTD, and BTD-output failures produce a full decision.
- BTD is pinned by a checked-in DotSlash manifest with reviewed archive sizes and BLAKE3 digests.
- BTD receives Rue's explicit Buck wrapper rather than relying on a `buck2` executable in `PATH`.
- A failed or skipped determinator job still starts the corpus matrix; missing outputs run every corpus.
- Only the gate's explicit exit status 1 records an intentional deselection. Unknown targets and malformed selection state run or fail open to running.
- Selection occurs within each `platform-corpus` job so every required check still reports a result.

The selection manifest accounts for every selectable corpus as RUN or DESELECTED, and each deselected job records its intentional skip.

## Changes

- Add the BTD DotSlash manifest and remove the bespoke release installer.
- Generate BTD's required `git diff --name-status --no-renames` input while projecting paths separately for force-full policy.
- Pass BTD the Rue Buck wrapper explicitly and strictly parse its JSON-lines output.
- Make dependency failure and gate errors fail open without silently dropping corpus jobs.
- Validate matrix and selected targets against the canonical selectable set.
- Add local Git/Buck/BTD stubs that exercise a real selective decision without network or a real Buck graph.
- Document the operating and fail-open model.

## Testing

- `scripts/test-affected-targets.sh` — 53 checks, including gate crash/missing cases and the selective end-to-end stub path
- `./buck2 test //:affected-targets-tool-tests`
- `scripts/validate-cli-shard-coverage.py`
- `scripts/test-required-ci-container-pins.py`
- `./btd --help` through DotSlash, verifying and executing the pinned macOS ARM artifact
- shell syntax, Python compilation, and `git diff --check`

Fixes RUE-1119


steveklabnik (changes requested):
codex: Reviewed head `b7b2e28f`. The overall BTD direction is sensible, but this is not safe to merge yet. The first CI run only exercised the force-full path because the PR changes the workflow itself. On the unexercised selective path, the change list has the wrong Git format and BTD cannot find Rue Buck2, so the feature always falls back to full. Separately, failed dependencies, gate crashes, malformed BTD output, and selectable-set drift can all skip corpus jobs despite the stated fail-open contract. I left concrete blockers inline, plus the request to use the repository-standard DotSlash pin for the downloaded executable. Please add an end-to-end selective-path test or CI exercise after fixing these; the current 39 checks do not execute BTD or the job-level failure paths.

scripts/affected-targets
```diff
@@ -0,0 +1,274 @@
...
+    if [[ -z "$base_ref" ]]; then
+        decide_full "could not resolve a base ref for the diff"
+    fi
+
+    # Merge-base diff (three-dot) so unrelated changes on the base branch do not
+    # inflate the changed-file set.
+    local changes_file
+    changes_file="$(mktemp)"
+    _af_cleanup+=("$changes_file")
+    if ! git diff --name-only "${base_ref}...${head_ref}" >"$changes_file" 2>/dev/null; then
```
steveklabnik: codex: **[P1] Feed BTD Git status records, not path-only records.** The pinned 2026-07-20 BTD contract parses `git diff --name-status` tab-separated lines when `--vcs git` is set; its own Git helper uses `--name-status --no-renames`. A line such as `crates/rue-air/src/lib.rs` has no status field, so BTD returns nonzero and every otherwise-selectable PR falls back to full. Keep a status-bearing changes file for BTD and a separate path-only projection for `force_full_match`, including conservative add/delete handling for renames, and cover the real parser contract in a test.

scripts/affected-targets
```diff
@@ -0,0 +1,274 @@
...
+    if ! ( cd "$base_tree" && buck_targets_dump "$base_tree/buck2" "$work/base.jsonl" ) >&2; then
+        base_ok=0
+    fi
+    if [[ "$base_ok" -ne 1 ]]; then
+        decide_full "buck2 targets dump at base failed"
+    fi
+
+    # Run BTD over the two dumps and the changed-file list.
+    local btd_out="$work/impacted.jsonl"
+    if ! "$btd" \
```
steveklabnik, line 260: codex: **[P1] Give BTD the pinned Rue Buck2 command.** Before it reads `--changes`, BTD obtains cells and config by invoking its `--buck` command unless explicit `--cells`/`--config` files are supplied. The default is bare `buck2`; `facebook/install-dotslash` only adds `dotslash` to PATH, while Rue exposes Buck2 through `./buck2`. On a normal runner this therefore fails before analysis and again forces every PR to full. Pass `--buck "$repo_root/buck2"` or supply captured cell/config inputs, and exercise this path rather than relying on the fallback.

.github/workflows/ci.yml
```diff
@@ -291,7 +353,13 @@ jobs:
   # (BUCK) hash-partitioned shards per platform so its ~12 min collapses to
   # ~N-way parallel slices; //:cli-shard-coverage-validation keeps this matrix
   # and the BUCK shard set in lockstep.
+  #
+  # RUE-1119: each job first consults the affected-targets determinator and,
+  # on a pull_request where its corpus is provably unaffected, skips the heavy
+  # steps (recording an intentional deselection) instead of running the suite.
+  # merge_group always reports full=true, so every corpus runs there.
   platform-corpus:
+    needs: affected-targets
```
steveklabnik, line 363: codex: **[P1] A failed determinator currently skips the corpus as success.** With a plain `needs`, any checkout/action/runner failure in `affected-targets` skips every `platform-corpus` matrix job. GitHub documents that skipped jobs report success and may not block merging even when required; the new determinator check is not made required here. Add a job-level `if: ${{ always() }}` (then treat a missing/failed dependency result as full) so an upstream crash runs the corpus instead of silently satisfying its checks.

.github/workflows/ci.yml
```diff
@@ -390,14 +458,33 @@ jobs:
...
+      # deselection and skips every step below, so a deselected corpus costs
+      # only the runner spin-up. Fail-open: unless the determinator explicitly
+      # says full=false and omits this target, the corpus runs.
+      - name: Corpus selection
+        id: sel
+        env:
+          RUE_AFFECTED_FULL: ${{ needs.affected-targets.outputs.full }}
+          RUE_AFFECTED_TARGETS: ${{ needs.affected-targets.outputs.selected }}
+        run: |
+          if scripts/ci-corpus-selected "${{ matrix.target }}"; then
```
steveklabnik: codex: **[P1] Distinguish intentional deselection from gate failure.** This `if/else` maps every nonzero status to `run=false`, including exit 2, command-not-found, permission loss, or a shell/parser error. That is fail-closed: a broken selection script can skip all heavy suites even when the determinator emitted `full=true`. Only the one documented deselection status may set `run=false`; any other status must log the failure and set `run=true`. Add a regression that injects a crashing/missing gate.

scripts/parse-btd-impacted.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def main() -> int:
+    seen = []
+    seen_set = set()
+    for line in sys.stdin:
+        line = line.strip()
+        if not line:
+            continue
+        try:
```
steveklabnik, line 34: codex: **[P1] Malformed BTD output must force full, not erase targets.** Every malformed JSON line and every object without a string `target` is skipped, and `main` always returns 0. I verified that `not-json`, `{}`, and `{"target":42}` together produce empty output with exit 0; the caller then publishes `full=false, selected=''` and deselects every corpus. Since the safety contract says any parse failure runs full, reject any nonempty malformed/unexpected line with a nonzero exit and test both partial and wholly malformed streams. A truly empty valid BTD stream can remain distinguishable from parse failure.

scripts/ci-corpus-selected
```diff
@@ -0,0 +1,45 @@
...
+
+full="${RUE_AFFECTED_FULL:-}"
+selected="${RUE_AFFECTED_TARGETS:-}"
+
+# Anything other than an explicit selective decision runs the suite.
+if [[ "$full" != "false" ]]; then
+    exit 0
+fi
+
+# Selective run: run only if this exact target is in the selected set.
```
steveklabnik, line 56: codex: **[P1] Unknown matrix targets are currently deselected, contrary to the fail-open claim.** The gate has no knowledge of `SELECTABLE_CORPUS`; on a selective run any target absent from `selected` returns 1. If a future required corpus is added to the workflow but not the script list, BTD can never select it and every PR silently skips it. I verified `//:future-required-corpus` exits 1. Pass the complete selectable set as an output or add a single-source validation, and make a target outside that set return 0.

scripts/install-btd
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+# Provision a pinned `btd` (Buck Target Determinator, from Meta's
+# facebookincubator/buck2-change-detector) into a destination directory.
+#
+# BTD is the off-the-shelf Buck2 target determinator used by scripts/affected-
+# targets (RUE-1119) to select which corpus suites a pull request actually
+# affects. It is deliberately NOT pinned via DotSlash like ./buck2 or
```
steveklabnik: codex: **[P1] Pin this executable with DotSlash instead of a same-channel checksum installer.** Downloading the archive and its checksum from the same replaceable GitHub release does not give Rue an immutable trust anchor: an upstream asset replacement or compromise can replace both, and this binary runs in CI after the private cache is provisioned. Rue already checks content digests and sizes into `buck2-bin`/`reindeer`; a `btd` DotSlash manifest gives the same content-addressed verification, platform dispatch, and cache behavior while deleting this bespoke installer. CI only requires linux-x86_64 initially, so capturing every optional local platform is not a prerequisite.

scripts/affected-targets
steveklabnik: codex: Fixed in c261e352. The selector now gives BTD  records, projects A/M/D paths separately for force-full policy, fails open on unfamiliar records, and tests both the projection and a selective Git/BTD integration.

scripts/ci-corpus-selected
steveklabnik: codex: Fixed in c261e352.  is the shared validation source. Unknown matrix targets run, and unknown selected-output targets produce an error that the workflow adapter converts to running. Both paths are covered.

.github/workflows/ci.yml
steveklabnik: codex: Fixed in c261e352.  now uses job-level ; failed/skipped dependency outputs are empty, which the gate treats as full, so the matrix runs instead of silently disappearing.

scripts/install-btd
steveklabnik: codex: Fixed in c261e352. The installer is deleted and replaced by an executable  DotSlash manifest with reviewed per-platform sizes and BLAKE3 digests. I independently downloaded all three pinned archives, corrected their metadata, and executed the macOS ARM artifact through DotSlash.

scripts/parse-btd-impacted.py
steveklabnik: codex: Fixed in c261e352. Nonempty malformed JSON, non-object records, and missing/non-string targets now return nonzero and force full; an empty stream remains a valid empty impact set. Partial and wholly malformed cases are covered.

scripts/affected-targets
steveklabnik: codex: Fixed in c261e352. BTD now receives  with Rue's explicit wrapper. The stubbed selective integration asserts that exact flag and value, and the pinned BTD binary itself was executed through DotSlash to confirm the option contract.

.github/workflows/ci.yml
steveklabnik: codex: Fixed in c261e352. The workflow-facing  adapter maps only exit 1 to ; every other status runs. Tests inject exit 2 and a missing gate executable and require .

steveklabnik (approved):
codex: Re-reviewed head c261e352 after applying the fixes directly. All seven correctness blockers are addressed, the selective path is exercised end to end with local stubs, gate crash/missing cases fail open, the focused Buck target passes, and the content-pinned BTD artifact executes through DotSlash. Approved.

PR #1905, opened 2026-07-23
RUE-1120: key successor parse by exact segment

steveklabnik:
codex:

Successor parse keys now carry only the exact source segment appended by that parse stage, rather than the cumulative trusted-toolchain lineage delta. Preparation derives each segment from the appended module/file pairs, verifies its exact source identity against the capability-authorized cumulative delta, and records key-comparison work for only that segment.

The acquisition-profile regression pins the two-stage fixture at four key entries for four dispatched modules. The old cumulative shape would charge six entries for the same work.

Validation:

- `scripts/rue unit compiler successor_close_reuses_the_staged_terminal_with_a_predecessor_edge`
- exact `source_loader::tests::read_line_reclose_roots_only_new_leaves_independent_of_project_topology`
- `scripts/rue quick` (one parallel-load timeout in the frontend differential)
- isolated `//crates/rue-frontend-diff:frontend-diff-test` rerun: 1443/1443 corpus files and both syntax probes pass
- `git diff --check`

Fixes RUE-1120

### DIFF
--- a/crates/rue-compiler/src/session.rs
+++ b/crates/rue-compiler/src/session.rs
@@ -1937,7 +1937,7 @@ pub(crate) enum ParseQueryKey {
     /// equality never touch a predecessor entry.
     Successor {
         revision: crate::ImportInputRevision,
-        delta: Arc<[crate::ModuleRevision]>,
+        segment: Arc<[crate::ModuleRevision]>,
         /// The exact retained predecessor parse terminal this successor
         /// extends, verified by structural source ancestry at preparation.
         predecessor: rue_query::Revision,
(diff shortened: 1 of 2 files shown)


steveklabnik:
fable: Reviewed `e59bad07` — approved; leaving it in the queue. The segment derivation is compiler-side (structural-ancestry appended set), cross-verified against the capability-authorized cumulative delta by module identity **and** byte-exact source identity — the source-equality rejection is a strengthening the cumulative shape never had. Per-stage keys stay distinct across chained stages (revision + predecessor + segment), the staged-terminal reuse path is untouched, an observation-only stage degenerates to an empty segment under a fresh revision, and the profile pins are now exact (4 entries / 4 dispatches, cumulative-6 ruled out) rather than non-vacuous-only. The frontend-diff timeout in `scripts/rue quick` matches the documented parallel-load flake; the isolated rerun plus CI shards cover it.

---

PR #1906, opened 2026-07-23
RUE-1039: audit frontend differential corpus

steveklabnik:
## Summary

- replace the frontend differential corpus count with an explicit root/path identity manifest
- fail with exact added, removed, and unexpected paths, including owning roots
- add a deterministic scripts/rue frontend-manifest refresh command backed by the Buck corpus target

## Validation

- scripts/rue unit frontend-diff (9 passed)
- frontend-diff-test (1443/1443 files and 2 syntax probes agree)

Fixes RUE-1039


steveklabnik:
fable: Reviewed `c0c9721c` — approved; queueing it. The audit is exact in both directions with owning roots named, path handling is hardened (`..`, absolute, duplicate, non-UTF-8 all rejected), determinism is proven by the render-stability test, the refresh command consumes the same Buck corpus target the test does (no shadow filesystem walk to disagree with), and the identity audit runs before the compiler is even built so drift is the first and loudest failure. The in-PR mutation tests cover added/removed/unexpected/missing-root exactly. One noted (and I think intentional) property: `--refresh-manifest` only rewrites declared roots, so introducing a new corpus root requires a reviewed manifest edit — while the audit still catches an undeclared root as `unexpected`, so nothing can drift silently either way.

---

PR #1907, opened 2026-07-23
RUE-1121: pin body-context repair acceptance

steveklabnik:
codex:

This lands the mechanism-independent RUE-1091 acceptance contract before the repair:

- all four observable per-body context counters must remain exactly flat as unrelated declarations grow, while preserving RUE-1090's original exact-ratio decision;
- edit rows require the exact recomputed body identities as well as the exact count, measured from retained body-terminal provenance so recomputing the right number of wrong bodies cannot pass;
- unrelated addition targets `{}`, a body-text edit targets `{b0}`, declaration-value consumers target `{left, right}`, and negative-to-positive lookup targets `{extra, main}`;
- every warm revision is checked against an independently compiled fresh revision, and repaired work must be strictly cheaper than fresh;
- the existing two-envelope discipline remains intact, so current whole-universe behavior is tracked but any partial shape between witness and target fails loudly.

The retained-origin inspection is test-only and reads canonical body-query terminal provenance; it does not introduce production instrumentation or a parallel semantic authority.

Validation:

- `scripts/rue fmt`
- `scripts/rue unit compiler invalidation_`
- `scripts/rue unit compiler scaling_smoke_fixed_bodies_growing_declarations`
- `scripts/rue unit compiler:scaling-matrix-test scaling_matrix_`
- `git diff --check`

Fixes RUE-1121

### DIFF
--- a/crates/rue-compiler/src/revisioned_query_database.rs
+++ b/crates/rue-compiler/src/revisioned_query_database.rs
@@ -9191,6 +9191,59 @@ impl RevisionedQueryDatabase {
         self.body_transactions.any_retained_key(|_| true)
     }
 
+    /// Test-only exact provenance for retained ordinary body terminals.
+    ///
+    /// Looking up retained keys directly keeps failed-revision acceptance tests
+    /// independent of the aggregate stable-definition request, which may itself
+    /// fail after a negative name lookup even though unrelated body terminals
+    /// remain reusable.
+    #[cfg(test)]
+    pub(crate) fn retained_body_transaction_origins_for_test(
+        &self,
+        revision: Revision,
+        names: &[String],
+    ) -> BTreeMap<String, u64> {
+        let mut origins = BTreeMap::new();
+        for name in names {
+            let mut retained_key = None;
+            self.body_transactions.any_retained_key(|candidate| {
+                let crate::FunctionInstanceKey::Definition(definition) = &candidate.instance else {
+                    return false;
+                };
+                if definition.name() != name.as_str() {
+                    return false;
+                }
+                retained_key = Some(candidate.clone());
+                true
+            });
+            let Some(key) = retained_key else {
+                continue;
+            };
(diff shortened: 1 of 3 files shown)


steveklabnik:
fable: Pulled this from the queue temporarily — my verification pass on the acceptance rows is still in flight (mechanism-independence, the exact edit-row identity sets, pre-repair envelope state, provenance-read observer effects, and warm/fresh isolation). Since these rows are the contract the production repair will be built and judged against, they get the same review-before-merge bar as everything else in this arc. I'll re-queue immediately on a clean result or post findings if not — expected within the hour.

---

steveklabnik:
fable: Verification complete — approved, re-queueing now. Highlights of the pass: the mechanism-independence claim was verified in the strongest sense (the PR *deletes* the recipe/durable-base-coupled assertions its predecessor rows had, replacing them with artifact-level observables — origin-identity sets and cold-body counts — with stage counters asserted only for same-implementation relative flatness); the between-witness-and-target hard-fail was empirically confirmed by injecting a wrong identity set and observing the panic; every new row on this head resolves PASS or tracked-XFAIL exactly as described ({left,right} and {b0} already meet target; unrelated-add and negative-to-positive correctly track the whole-universe witness); the {extra,main} set was validated against the corpus (main is the genuine consumer of the flipping lookup, with the occurrence-count guard proving uniqueness); the provenance read is provably non-perturbing (cancel-compute idiom, no retained lease, measured-before-read ordering); and the fresh-session isolation is mechanical (fresh cold_bodies == full count asserted, strict inequality against warm). RUE-1090's ACTIVATE verdict prints unchanged. Full suites green (compiler 734, scaling matrix 3, invalidation rows 6, clippy/fmt clean).

One recorded non-blocking note for the future: origin selection via first-retained-key-per-name is deterministic only while the corpus stays single-configuration per name; a multi-configuration corpus would need a tiebreak.

---

PR #1908, opened 2026-07-23
RUE-1122: reject out-of-range relocation offsets

steveklabnik:
## Summary

- validate relocation patch widths against their source sections before merging
- use checked arithmetic for merged offsets and patch virtual addresses
- centralize overflow-safe patch-range construction for ELF and Mach-O
- cover maximum, near-boundary, and real parsed-ELF relocation offsets

## Root cause

ELF `r_offset` values were retained as arbitrary `u64` values, then combined with merged section offsets using unchecked addition. Patch bounds checks also added the relocation width before checking the buffer length, so malformed offsets could overflow, panic, or wrap in optimized builds.

## Validation

- `scripts/rue unit linker` — 145 passed
- `scripts/rue quick` — 36 targets passed
- `bash ./clippy.sh` — 66 targets passed
- `scripts/rue fmt`

Fixes RUE-1122

### DIFF
--- a/crates/rue-linker/src/linker.rs
+++ b/crates/rue-linker/src/linker.rs
@@ -94,6 +94,96 @@ fn is_bss_section(name: &str) -> bool {
     classify_section(name) == SectionKind::Bss
 }
 
+/// Number of bytes written by a supported relocation.
+///
+/// Keeping this next to the shared patcher gives collection and application
+/// one width authority. That lets the collector reject a malformed patch site
+/// while it still knows the source section's exact bounds.
+fn relocation_patch_size(rel_type: RelocationType) -> Result<usize, LinkError> {
+    match rel_type {
+        RelocationType::Abs64 | RelocationType::Aarch64Abs64 => Ok(8),
+        RelocationType::Pc32
+        | RelocationType::Plt32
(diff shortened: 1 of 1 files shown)


steveklabnik:
fable: Reviewed `21c1f6ae` — approved. The substantive fix is better than the summary suggests: the old per-arm checks bounded the patch against the *merged* buffer, so a malformed section-relative offset could escape its own section yet land inside another section's bytes and silently corrupt it — the new collection-time `checked_patch_range` against the source section's own length closes that class, with the apply-time check retained as defense in depth. Width authority is centralized next to the patcher (collection and application can't disagree), all offset composition is checked arithmetic with typed errors, and the parsed-ELF test (byte-surgering the RELA entry to `u64::MAX` and asserting the typed rejection with exact width/section context) proves the real pipeline path, not just the helpers. One noted behavioral nuance, fine as-is: bounds are now validated before per-arm value-overflow checks, so doubly-malformed input reports `RelocationPatchOutOfBounds` where it previously reported `RelocationOverflow` — the new precedence is the sensible one and the tests pin it. Updating the branch onto current trunk and queueing.

---

PR #1923, opened 2026-07-23
RUE-1044: prefer Option for lookup absence

steveklabnik:
## Summary

- add a narrow warning for `get_or(_, sentinel)` comparisons that use the same integer sentinel to test absence
- document `Option` as the canonical no-index/no-value representation and reserve `get_or` for genuine fallback values
- give the hashmap example an `Option`-returning `get` API and use it for membership testing

## Why

Rue can already represent optional copy values, and `ArrayBuf.index_of` already returns `Option(u64)`. This closes the remaining guidance and adoption gap without touching the incremental compiler work or converting legitimate fallback uses.

Fixes RUE-1044.

## Validation

- `scripts/rue fmt`
- `scripts/rue unit parser`
- `scripts/rue ui sentinel_lookup`
- `scripts/rue cli hashmap`
- `scripts/rue quick`
- `scripts/rue test`

### DIFF
--- a/crates/rue-air/src/sema/analysis/builtin_ops.rs
+++ b/crates/rue-air/src/sema/analysis/builtin_ops.rs
@@ -5,6 +5,12 @@
 
 use super::*;
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum IntegerSentinel {
+    UnsignedMax,
+    NegativeOne,
+}
+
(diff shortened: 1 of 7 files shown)


steveklabnik:
This is an interesting idea for a warning, let's see what it shakes out like and if it sucks we can remove it.

PR #1926, opened 2026-07-24
RUE-1091 r1c: hoist aggregate resolution behind epoch facts

steveklabnik:
Hoist family 1D aggregate, field, variant, pattern, and visibility resolution behind an internal `AggregateFacts` seam.

## What changed

- Add provider-generic selection logic for unqualified aggregate heads, module-member precedence, qualified enum/struct dispatch, module spines, and visibility.
- Add a short-lived concrete `EpochFacts` adapter that reproduces the existing epoch reads and lookup order.
- Repoint only `aggregates.rs` and `visibility.rs`; the family-1C equality-method lookup remains untouched.
- Preserve the original path borrowing and clone points so the refactor adds no hidden allocation work.

This is a refactor-only r1c slice. It does not add `ProviderFacts`, change diagnostics, or alter the production path.

## Validation

- rue-air: 574/574
- rue-compiler: 790/790
- full spec corpus: 2,193
- full CLI corpus: 1,735
- differential oracle: 4/4
- scaling filter: 13/13; `RUE-1090 VERDICT: ACTIVATE RUE-1091` unchanged
- clippy: 66 targets
- fmt
- byte-equality spot proof on `dijkstra`, `jsonfmt`, and `bignum`: stderr and linked artifacts identical before/after

Part of RUE-1091

### DIFF
--- a/crates/rue-air/src/sema/aggregate_resolution.rs
+++ b/crates/rue-air/src/sema/aggregate_resolution.rs
@@ -1,0 +1,280 @@
+//! Provider-generic aggregate, field, and variant resolution.
+//!
+//! Selection order lives here. [`EpochFacts`] is the production adapter that
+//! reproduces the declaration-epoch reads used by body analysis.
+
+use lasso::Spur;
+use rue_rir::{InstData, InstRef, Rir};
+use rue_span::FileId;
+
+use super::context::{ConstValue, LocalVar};
+use super::{ConstInfo, DeclarationPhase, Sema};
+use crate::types::{EnumId, ModuleId, StructId, Type};
+
+pub(crate) trait AggregateFacts {
+    fn value_const(&self, file: FileId, name: Spur) -> Option<ConstInfo>;
+    fn module_binding(&self, file: FileId, name: Spur) -> Option<ConstInfo>;
+    fn struct_in_file(&self, file: FileId, name: Spur) -> Option<StructId>;
+    fn enum_in_file(&self, file: FileId, name: Spur) -> Option<EnumId>;
+    fn builtin_struct(&self, name: Spur) -> Option<StructId>;
+    fn builtin_enum(&self, name: Spur) -> Option<EnumId>;
+    fn module(&self, module: ModuleId) -> AggregateModuleFact;
+    fn file_path(&self, file: FileId) -> Option<&str>;
+    fn source_path(&self, span: rue_span::Span) -> Option<&str>;
+}
+
+pub(crate) struct AggregateModuleFact {
+    pub(crate) file: FileId,
+    file_path: String,
+    import_path: String,
+}
+
+impl AggregateModuleFact {
+    pub(crate) fn file_path(&self) -> &str {
+        &self.file_path
+    }
+
+    pub(crate) fn import_path(&self) -> &str {
+        &self.import_path
+    }
+}
+
+pub(crate) struct EpochFacts<'s, 'a, D: DeclarationPhase> {
+    sema: &'s Sema<'a, D>,
+}
+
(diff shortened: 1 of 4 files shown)


steveklabnik (commented):
Adversarial review against base `3de83771`. **One blocker**, one note; everything else reconciles cleanly. Please hold this out of the queue until the blocker is fixed.

**Blocker — `resolve_module_ref` changes the aggregates-path VarRef fall-through.** The two base helpers being unified had different semantics: `aggregates.rs::try_module_id_of` lets a local that exists but is **not a module** fall through to `module_binding`, while `visibility.rs::module_file_for_ref` short-circuits on any existing local (`.map().or_else()`). The new generic (`aggregate_resolution.rs:224`) adopts the visibility flavor for both callers, so on the aggregates path a non-module local shadowing a same-named module binding (e.g. `let helper = 5;` alongside a module-typed `const helper = @import(...)`, used as `helper.Foo(...)`) now returns `None` where the base resolved the module — routing the expression to a different resolution path with different AIR/diagnostics, and conditionally dropping a `module_binding` read the base issues. The `root_file` parameter correctly reproduces the file-key difference between the two callers, but the fall-through must be reproduced too — e.g. a flag or a variant that preserves "non-module local falls through" for the aggregates caller. Narrow reachability, but the slice's proof obligation is byte-identity of EpochFacts selection with the old inline selection, and this is provably not identical. A regression test pinning the shadow case would be welcome in the fix.

**Note (resolved by fiat, no change needed):** the plan named one `SemanticCallResolutionProvider` in `sema/call_resolution.rs`; this PR introduces a per-family `AggregateFacts` trait in `aggregate_resolution.rs`. Given the r1a/r1b/r1c split, per-family seams are the right shape — keep it as is; the plan doc will be aligned to per-family traits in the next coordination amendment.

Everything else verified against the base: read-for-read reconciliation (keys, maps, short-circuit order, `get_def` call counts including the recursive module spine), struct-literal head selection order, module-member struct→enum→const precedence with identical short-circuiting, diagnostic identity across all paths, observation-recording identity (`record_body_named_dependency` unchanged), scope discipline (family-1C equality lookup untouched), and template conformance (monomorphized generic, short-lived borrows, no `&dyn`). After the fall-through fix, I'll rerun the byte-equality spot-proof on the updated head before this queues.

---

steveklabnik:
Addressed the blocker in the updated head. The generic module-reference logic is now split into exact aggregate and visibility ports: aggregate lookup preserves non-module-local fallthrough, while visibility preserves any-local shadowing. The provider seam remains shared only at the facts layer. Fresh post-fix evidence: rue-air 574/574, rue-compiler 790/790, differential oracle 4/4, spec 2175 passed with 18 expected ignores, scaling 13/13 with `RUE-1090 VERDICT: ACTIVATE RUE-1091`, clippy 66 targets, and fmt. A controlled isolated r0/current proof is byte-identical for artifacts and stderr on dijkstra, jsonfmt, and bignum.

steveklabnik (commented):
Fix verified against both base helpers: the split into `resolve_aggregate_module_ref` (fall-through preserved, `span.file_id` root) and `resolve_visibility_module_ref` (short-circuit preserved, `inst.span.file_id`) makes each flavor verbatim-traceable to its base, with flavor-consistent FieldGet recursion — a stronger identity argument for this corner than the 3-program spot-proof, whose programs don't exercise the shadow path at all. Blocker resolved; enabling auto-merge.

One follow-up I'm taking rather than asking for another round here: a CLI regression case pinning the shadow scenario (non-module local shadowing a same-named module binding in a module-qualified position) so the fall-through can't be silently re-unified later — it'll ride with the r1b slice.

---

PR #1931, opened 2026-07-24
RUE-1091: add inert flip-slice capability guard

steveklabnik:
## Summary

- add a static source-inventory classifier for exact-body epoch-table capabilities
- attribute every observed read to an epoch-backed facts implementation, the plan's explicit whitelist, or an unaccounted category
- keep the guard inert until the flip via a single `PROVIDER_ONLY_MODE` constant
- document each whitelist carry-forward and its expected removal slice

## Why

Building the capability inventory before the analyzer flip keeps the flip itself focused on deleting the epoch path. A static inventory fits the existing Buck guard pattern and avoids adding a production observation hook.

The current tree reports 86 epoch-authority reads, 30 whitelisted reads, and 0 unaccounted reads. This includes bare `self.<field>` reads in legacy `impl Sema` spans. The test does not require either allowed category to shrink, and the field-classification coverage fails closed when `Sema` or `DeclarationNamespace` gains an unclassified field.

## Validation

- `./buck2 test //:body-analysis-capability-inventory-validation //:body-analysis-capability-inventory-tool-tests`
- `scripts/rue fmt`

Refs RUE-1091.

### DIFF
--- a/BUCK
+++ b/BUCK
@@ -477,6 +477,24 @@ sh_test(
     },
 )
 
+sh_test(
+    name = "body-analysis-capability-inventory-validation",
+    test = "scripts/validate-body-analysis-capabilities.py",
+    args = [
+        "--source", "rue-air=$(location //crates/rue-air:body-analysis-capability-inventory-sources)",
+        "--source", "rue-compiler=$(location //crates/rue-compiler:body-analysis-capability-inventory-sources)",
+    ],
+)
+
+sh_test(
+    name = "body-analysis-capability-inventory-tool-tests",
+    test = "scripts/test-body-analysis-capabilities.py",
+    resources = ["scripts/validate-body-analysis-capabilities.py"],
+    env = {
+        "PYTHONDONTWRITEBYTECODE": "1",
+    },
+)
+
 test_suite(
     name = "payload-ownership-compile-fail-tests",
     tests = ["//crates/rue-rir:rue-rir[doc]"],
--- a/crates/rue-air/BUCK
+++ b/crates/rue-air/BUCK
@@ -18,6 +18,12 @@ filegroup(
     visibility = ["PUBLIC"],
 )
 
+filegroup(
+    name = "body-analysis-capability-inventory-sources",
+    srcs = glob(["src/**/*.rs"]),
+    visibility = ["PUBLIC"],
+)
+
 rue_crate(
     name = "rue-air",
     deps = [
--- a/crates/rue-compiler/BUCK
+++ b/crates/rue-compiler/BUCK
@@ -12,6 +12,12 @@ filegroup(
     visibility = ["PUBLIC"],
 )
 
+filegroup(
+    name = "body-analysis-capability-inventory-sources",
+    srcs = glob(["src/**/*.rs"]),
+    visibility = ["PUBLIC"],
+)
+
 # Per-target runtime staticlibs are mapped into src/ so the compiler can embed
 # all supported runtimes and select the matching archive at link time.
 _RUNTIME_MAPPED_SRCS = {
--- a/scripts/test-body-analysis-capabilities.py
+++ b/scripts/test-body-analysis-capabilities.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+"""Focused tests for the RUE-1091 body-analysis capability inventory."""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+SCRIPT = Path(__file__).with_name("validate-body-analysis-capabilities.py")
+SPEC = importlib.util.spec_from_file_location("body_analysis_capabilities", SCRIPT)
+assert SPEC is not None and SPEC.loader is not None
+inventory = importlib.util.module_from_spec(SPEC)
+sys.modules[SPEC.name] = inventory
+SPEC.loader.exec_module(inventory)
+
+
+class CapabilityInventoryTests(unittest.TestCase):
+    def classify(self, files: dict[str, str]) -> list[object]:
+        with tempfile.TemporaryDirectory() as directory:
+            source_root = Path(directory) / "materialized"
+            for relative, source in files.items():
+                path = source_root / relative
+                path.parent.mkdir(parents=True, exist_ok=True)
+                path.write_text(source)
+            return inventory.classify([("rue-air", source_root)])
+
+    def test_epoch_impl_whitelist_and_unaccounted_are_distinct(self) -> None:
+        hits = self.classify(
+            {
+                "sema/body_endpoint.rs": (
+                    "struct EpochFacts<'a> { sema: &'a Sema }\n"
+                    "impl Facts for EpochFacts<'_> {\n"
+                    "  fn read(&self) { self.sema.functions.len(); }\n"
+                    "}\n"
+                    "fn escaped(sema: &Sema) { sema.methods.len(); }\n"
+                ),
+                "sema/one_body.rs": (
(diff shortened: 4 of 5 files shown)


steveklabnik (commented):
Adversarial review against base `46b93d80`. The inertness/scope/CI framing is honest and clean, but the classifier at the core is **unsound in both directions** — two blockers before this can be trusted to gate the flip.

**Blocker 1 — the receiver-anchored regex is blind to bare `self.<field>` reads inside `impl Sema` blocks.** `EPOCH_ACCESS` matches only `sema.FIELD` / `self.sema.FIELD`. But `typeck.rs` (in `WHOLE_FILE_INVENTORY`) has the sole `impl<'a, D: DeclarationPhase> Sema<'a, D>` block (typeck.rs:1526–2400) where epoch tables are read as bare `self.FIELD` — e.g. `self.interner.resolve(...)` (1717, 1766, 1996, 2024), `self.module_registry.get_def(...)` (2375); 21 such lines, no whitelist entry covers them. If the classifier could see them they'd bucket unaccounted and the guard would fail today — the advertised "0 unaccounted" is a false negative from undercounting, and post-flip a genuine leak written as `self.functions.get()` in any `impl Sema` method sails through. Fix: also scan `self.<field>` within `impl Sema*` spans (the existing lexical-span machinery you built for EpochFacts attribution can carry this), or drop the receiver anchor and attribute by span.

**Blocker 2 — `EPOCH_FIELDS` omits in-scope table families with confirmed live reads.** The 24-entry hand-curated list is missing at least `generated_structs`/`generated_enums` and `anon_struct_identities`/`anon_enum_identities`: `body_endpoint.rs:158,163,180` (`generated_structs`) and `:171,175` (`anon_*_identities`) are live reads inside the `EpochFacts` impl that the classifier cannot see — so "allowed=54" undercounts today AND the flip would fail to flag residual reads of these tables. The Sema struct (sema/mod.rs:272+) has more plausible omissions (`const_candidate_*`, `body_owner_tokens`, `reusable_*_bodies`, `anonymous_*_ids`, `canonical_anonymous_types`). Fix: derive `EPOCH_FIELDS` from the actual `Sema` struct fields, or add a coverage test asserting the list is complete against the struct definition so it can't silently rot as tables are added.

**Minor (non-blocking):** the 5 whitelist entries match the carry-forward intent but the function names appear nowhere in the plan doc — add a comment citing the plan's "Review carry-forwards" section per entry so traceability isn't prose-only; and `universe_cardinality_counters` is narrowed to a single fn+field, worth confirming that's the full set.

What verified clean: `PROVIDER_ONLY_MODE` inertness (single-constant flip, no shrinkage assertions), scope discipline (purely additive, no production code, no r5-live files), comment/string masking with offset preservation, lexical-span (not file-granular) EpochFacts attribution, root-BUCK target placement (matches the existing inventory-guard convention exactly), and CI wiring (`buck2 test //...` auto-discovers the new sh_tests). With the two classifier fixes plus a rerun of the counts, this is the right guard.

---

steveklabnik:
Addressed both blockers in amended commit `cceaf10f`.

- Bare `self.<field>` accesses are now scanned inside lexical `impl Sema` spans. They are reported as legacy epoch-authority accesses in inert mode, and `PROVIDER_ONLY_MODE` rejects them at the flip. A focused regression proves that `self.functions` is visible to strict mode.
- The inventory now includes `generated_structs`, `generated_enums`, `anon_struct_identities`, and `anon_enum_identities`, plus the other epoch tables identified in the review. Every field declared by `Sema` and `DeclarationNamespace` must now be classified as epoch authority or non-epoch/body-local; a new unclassified field fails closed. A regression covers that completeness check.
- Each whitelist entry now cites the plan's “Review carry-forwards” section and its removal slice. The cardinality entry covers the complete current set: both `functions_by_file_name.len()` reads in `analyze_all_function_bodies_mut_for_test`.

Revalidated:

- focused Python classifier tests: 6 passed
- both Buck inventory targets: passed
- current inventory: 86 epoch-authority, 30 whitelisted, 0 unaccounted

steveklabnik (commented):
Re-review of `cceaf10f`: both original blockers are substantively fixed — the bare-`self` handling is span-driven and principled (legacy `impl Sema` spans classified as legacy epoch authority that provider-only mode rejects, with the production `SemaTypeSyntaxProvider` impl separately bucketed; verified all 21 typeck.rs self-reads land correctly and flip to violations under `PROVIDER_ONLY_MODE`), and the completeness gate genuinely parses the struct bodies with all previously-missed fields now classified. The +32 arithmetic composition checks out. Whitelist citations and the `universe_cardinality_counters` scoping are both right now.

**One short round remains — L2:** `declared_sema_fields()` captures `Sema` fields via `^\s*pub\(crate\)\s+(\w+)\s*:` plus hardcoded `"declarations"`, but `Sema` has four private fields today and two of them (`declaration_index`, `bound_const_candidates`) are epoch tables — proving epoch authority can be declared private on this struct. A future private field therefore escapes the fail-closed gate silently: never in `EPOCH_FIELDS`, never flagged unclassified. The PR body's "fails closed when `Sema` gains an unclassified field" holds only for `pub(crate)` (76 of 80 fields). Fix: use the same permissive `^\s*(\w+)\s*:` regex you already use for `DeclarationNamespace` (and extend `test_new_sema_field_requires_capability_classification` to a private field).

**L3:** `canonical_anonymous_types` is classified NON_EPOCH yet has three live reads in `one_body.rs` (361, 841, 1151) — the one genuinely borderline entry. The body-output/canonicalization-accumulator rationale is defensible, but because NON_EPOCH fields are excluded from the scan entirely, a wrong call here is invisible at flip time. Give it a dedicated comment with the plan-grounded justification rather than the generic complement note.

**Latent (comment-grade):** the `self`-receiver skip fires before the `epoch_impl` span check, so a future bare `self.<field>` read inside an EpochFacts impl would be silently skipped rather than counted allowed — today only false positives are dropped, but a reorder (or comment) would keep it honest.

Fully inert, no production code, no correctness risk — this is purely about the guard's own guarantee being as strong as the PR body claims. With the regex fix + citation, ready to go.

---

steveklabnik:
Addressed the remaining L2/L3 hardening in amended commit `0c68e8ab`.

- `declared_sema_fields()` now discovers both private and `pub(crate)` `Sema` fields. The regex rejects `::` path continuations, so multiline `crate::`, `std::`, and `anon_structs::` field types are not mistaken for fields. The completeness regression now uses a private `future_epoch_table`.
- `canonical_anonymous_types` has a dedicated plan-grounded comment: it is the body-local canonicalization accumulator for identities produced/materialized by the current attempt, not a declaration-epoch or whole-program namespace read.
- EpochFacts attribution now runs before the bare-`self` receiver skip. A regression proves a future `self.functions` read inside an EpochFacts impl is counted as allowed rather than silently dropped.

Revalidated:

- focused classifier tests: 7 passed
- both Buck inventory targets: passed
- current inventory remains 86 epoch-authority, 30 whitelisted, 0 unaccounted

steveklabnik (commented):
Round 3 (`0c68e8ab`) verified: the completeness regex now captures private fields (hardcoded `declarations` special-case gone, coverage test flipped to a private field so the fail-closed guarantee is actually exercised for the class that was escaping), `canonical_anonymous_types` carries its dedicated plan-grounded justification, and the latent skip-ordering issue is fixed with a pinning test (`test_bare_self_epoch_read_inside_epoch_impl_is_not_skipped`). All findings across three rounds resolved. Marking ready and enabling auto-merge — the guard is inert by construction, so it can land ahead of the remaining ProviderFacts slices; the flip slice tightens `PROVIDER_ONLY_MODE` when its time comes.

---

PR #1932, opened 2026-07-24
RUE-1091 rFinal: scaffold differential harness

steveklabnik:
## Summary

- adds a test-only whole-body differential interface with the current production analyzer and an explicit future ProviderFacts + overlay slot
- captures canonical artifacts, ordered diagnostic text, canonicalized query dependency edges, and per-body observation counters
- exercises the shape corpus under direct production body-analysis schedule permutations and verifies exact side-A self-equality
- inventories every `TypeInstanceKey` arm and `CanonicalArgumentValue` variant, with named review carry-forward gaps where no body terminal or future provider capability exists yet

## Scope

Production semantic paths are unchanged. The diff adds one test harness file and the minimal test-module registration only.

Refs RUE-1091.

## Validation

- `scripts/rue unit compiler rfinal_harness`
- `scripts/rue quick`
- `scripts/rue fmt`

### DIFF
--- a/crates/rue-compiler/src/session.rs
+++ b/crates/rue-compiler/src/session.rs
@@ -10344,6 +10344,9 @@ impl CompilerSession {
 
 #[cfg(test)]
 mod tests {
+    #[path = "rfinal_harness.rs"]
+    mod rfinal_harness;
+
     use std::{
         collections::{HashMap, HashSet},
         sync::Arc,
--- a/crates/rue-compiler/src/session/tests/rfinal_harness.rs
+++ b/crates/rue-compiler/src/session/tests/rfinal_harness.rs
@@ -1,0 +1,916 @@
+//! RUE-1091 rFinal whole-body differential scaffolding.
+//!
+//! This module is deliberately test-only. It captures the production
+//! `BodyTransaction` path today and leaves exactly one explicit slot for the
+//! future ProviderFacts + overlay implementation. No production path selects
+//! between analyzers.
+
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+    sync::Arc,
+};
+
+use super::*;
+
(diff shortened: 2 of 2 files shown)


steveklabnik (commented):
Adversarial review against parent `65d8075`: **ready to queue, no blockers.** The scaffolding delivers the brief faithfully — durable-body artifact capture (not a lossy summary), diagnostics order-preserved while edges are set-canonicalized on a provably stable sort key (the correct asymmetry), a real first-divergence comparator with body/field/byte-offset, all 22 `TypeInstanceKey` entries + 6 `CanonicalArgumentValue` variants constructed as real values (not vacuous labels), named executable carry-forward gaps with reasons, genuine call-order permutations against fixed inputs, an explicitly inert side B, and clean fencing (two files, zero production changes, CI-visible via the glob'd test target despite the scripts-only local validation). Enabling auto-merge.

Three refinements carried to the slice that fills side B (recorded here so they're not lost; none block this merge):
1. **Compile-time exhaustiveness guard**: the constructed inventory and the `*_COVERAGE` arrays are hand-maintained and compared only to each other — a 21st `TypeInstanceKey` arm would pass silently. A `match` over a sample value with all arms bound would fail-closed at compile time.
2. **Per-body counter honesty**: `provider`/`overlay` metrics are per-case snapshots that read zero until the flip — expected pre-side-B, but the side-B slice must move them to true per-body capture before they carry proof weight.
3. **Permutation depth**: all corpus cases have ≤2 bodies, so rotation degenerates to reverse and single-body cases permute vacuously — add a ≥3-body case when side B lands so the rotation earns its keep.

Also verify at side-B time: the `specialization` case actually mints a `FunctionInstanceKey::Specialization` at runtime (read-only review couldn't execute it; if it doesn't, the case silently falls back to `Definition`).

---

PR #1942, opened 2026-07-24
RUE-1091: refine rFinal harness coverage

steveklabnik:
## Summary

- adds exhaustive no-wildcard matches for the `TypeInstanceKey`, nested `NominalInstanceKey`, and `CanonicalArgumentValue` inventories
- frames provider and overlay metrics as case snapshots because direct production body analysis does not mutate the session-owned counters
- adds a retained three-body corpus case whose three bodies produce distinct values, making forward, reverse, and rotated schedules distinct

## Validation

- `scripts/rue unit compiler rfinal_harness`
- `./buck2 test //crates/rue-compiler:rue-compiler-test`
- `scripts/rue fmt`

Refs RUE-1091.

### DIFF
--- a/crates/rue-compiler/src/session/tests/rfinal_harness.rs
+++ b/crates/rue-compiler/src/session/tests/rfinal_harness.rs
@@ -152,6 +152,13 @@ const SHAPE_CORPUS: &[HarnessCase] = &[
         expected_failure: false,
     },
     HarnessCase {
+        name: "three_body_permutation",
+        source: "fn first() -> i32 { 11 } fn second() -> i32 { 22 } \
+                 fn main() -> i32 { first() + second() }",
+        bodies: &["first", "second", "main"],
+        expected_failure: false,
+    },
+    HarnessCase {
         name: "method",
         source: "struct Counter { value: i32, fn get(self) -> i32 { self.value } } \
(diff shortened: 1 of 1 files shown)


steveklabnik (commented):
Adversarial review against base `1e65db30`: **all three items implemented faithfully, no scope creep, ready to queue.**

Verified: the exhaustiveness guards bind every arm of all three enums including the nested `NominalInstanceKey` subvariants, with full field destructuring (so new *fields* break compilation too, not just new variants) and no catch-alls — and the guards run against live sample values, not dead code. The counter reframing is honest (fields genuinely renamed to `case_*_snapshot`, reasoning verified: the frozen pre-loop snapshot can't be mutated by direct body analysis) and does not foreclose per-body capture — the "until side B lands" note marks the transition. The 3-body case has load-bearing distinct values (a body-swap would fail the keyed comparison), produces three genuinely distinct schedule orderings, and is covered by the self-equality test automatically.

One optional L3 for whenever: the counter doc comment could spell out the mechanical side-B change (move the two snapshot lines inside the body loop / record per-body deltas) so the transition is a recipe, not a hint. Not a gate.

Marking ready and enabling auto-merge.

---

PR #1945, opened 2026-07-24
RUE-1091: share anonymous identity digest

steveklabnik:
## Summary

- move the pinned FNV-1a-128 hasher and stable anonymous-key encoding into a public rue-air module
- route the epoch digest path through the shared canonical function
- pin a representative stable-content key to its 32-hex-digit digest

This is a byte-identical refactor preparing the provider-era identity pool; it does not wire that pool or change identity behavior.

Refs RUE-1091.

## Validation

- `scripts/rue fmt`
- `./buck2 test //:type-architecture-inventory-validation`
- `bash ./clippy.sh` (Homebrew Bash; `/bin/bash` on this macOS host lacks `mapfile`)

### DIFF
--- a/crates/rue-air/src/lib.rs
+++ b/crates/rue-air/src/lib.rs
@@ -32,6 +32,7 @@ mod semantic_identity;
 mod semantic_import;
 mod semantic_type_resolution;
 pub mod specialize;
+pub mod stable_digest;
 mod type_encoding;
 #[cfg(test)]
 mod type_properties;
--- a/crates/rue-air/src/sema/anon_structs.rs
+++ b/crates/rue-air/src/sema/anon_structs.rs
@@ -391,42 +391,6 @@ impl<D: DeclarationPhase> Sema<'_, D> {
     }
 }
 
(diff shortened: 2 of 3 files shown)


steveklabnik (commented):
Adversarial review against base `392583a8`: **ready to queue, no findings.** Move-not-copy verified (the old hasher block is gone; exactly one FNV definition tree-wide); encoding equivalence is structurally guaranteed, not just eyeballed — the encoding is the derived `Hash` of `AnonymousNominalKey` in an unmodified file, and the hasher moved verbatim, so no reorder/separator regression is even reachable; both epoch call sites re-pointed with the `__anon_*_{digest:032x}` construction byte-unchanged and the cfg(test) forced-collision seam correctly retained epoch-side; the stability pin exercises kind/producer/anchor/multi-value arguments (and is fail-loud, not silent-drift, on any arithmetic surprise); pub surface minimal with no "Type"-named items. One optional nit for any later touch: the pin's `arguments.types` slice is empty, so that sub-path rides the values-slice coverage — a second pinned key with a non-empty types slice would close it. Marking ready and enabling auto-merge.

---

PR #1946, opened 2026-07-24
RUE-1091: add step-6 measurement gauntlet runner

steveklabnik:
Refs RUE-1091.

## Summary

- add a nine-stage, resumable post-flip measurement runner with per-stage logs, pass/fail capture, and an explicit pre-flip Caldera allow-fail policy
- keep counting-allocator evidence isolated from ordinary-allocator timing and memory runs
- add a results ledger prefilled with the frozen RUE-1090 ratios and the 41-minute Caldera DNF baseline reference

## Validation

- `scripts/rue-1091-measurement.sh --pre-flip --results-dir /tmp/rue-1091-preflip-validation`
  - stages 1-8 passed, including the 10k declaration row
  - stage 9 enforced the 300-second Caldera timeout and recorded `ALLOW_FAIL` with exit 124
- `scripts/rue fmt`
- `bash -n scripts/rue-1091-measurement.sh`
- schedule-permutation stage rerun after rebasing onto current trunk

### DIFF
--- a/docs/notes/rue-1091-measurement-results.md
+++ b/docs/notes/rue-1091-measurement-results.md
@@ -1,0 +1,65 @@
+# RUE-1091 post-flip measurement results
+
+Use this template for the authoritative run after the analyzer flip. The runner
+retains a `summary.tsv` plus one full log per stage:
+
+```sh
+scripts/rue-1091-measurement.sh --results-dir /absolute/path/to/results
+```
+
+For plumbing validation on current trunk, use `--pre-flip`. That mode changes
+only the Caldera stage to `--allow-fail`; it does not relax any structural or
+correctness stage. Resume an interrupted run with `--stage N` and the same
+results directory.
+
+## Provenance
+
+- Commit:
+- Date/time:
+- Host and OS:
+- CPU / logical processors:
+- Memory:
+- Build configuration: Buck2 release platform for measurement stages
+- Results directory:
+- Mode: post-flip authoritative / pre-flip plumbing
+- Caldera budget: 300 seconds (authoritative runs must leave
+  `RUE_1091_CALDERA_BUDGET_SECONDS` unset)
+
+Timing and memory in stage 8 come from the ordinary allocator binary. Allocation
+counts in stage 7 come from the distinct counting-allocator binary. Do not merge
+or compare their orchestration times from `summary.tsv`; those values include
+build and harness overhead and are operational only. The R5 evidence is the
+compiler timing reported by stage 8 only.
+
+## Stage ledger
+
+| Stage | Metric | Baseline reference | Post-flip value | Verdict |
+| ---: | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


steveklabnik (commented):
Adversarial review against base `392583a8`: strong core (allocator isolation is genuinely solid, allow-fail defaults to the safe direction, `PIPESTATUS` discipline, ratios arithmetically verified against the harness formulas, conventions clean) — but **one more round**: three findings bear on evidence integrity, and this script produces the one-shot post-flip sign-off.

**A (Medium) — the RUE-1090 growing-BODIES axis is never run.** Stages 1–2 invoke only `scaling_matrix_fixed_bodies_growing_declarations`; the harness also defines `scaling_matrix_fixed_declarations_growing_bodies` and `scaling_matrix_identity_invariant` behind the same cfg, and the plan's flip-slice step 5 explicitly requires both axes and the growing-bodies recompute targets. Add stages (or extend 1–2) covering both, and mirror them in the results ledger.

**B (Medium) — stages 1–2 can fail-as-pass on a zero-match filter.** They call the buck target directly, bypassing the `--list` preflight `scripts/rue unit` exists for ("libtest exits 0 when a filter matches nothing"). Today the names are right; a future rename records PASS with zero cases on the two headline stages. Route them through the same preflight or assert a nonzero executed-test count.

**C (Medium) — resume can greenwash.** `--stage N` retains earlier rows without validating they're PASS, and the terminal verdict checks only the in-run failure flag — so resuming past a failed stage prints success while the ledger holds a FAIL row, and nothing enforces the doc's own "must not contain ALLOW_FAIL" rule. Make the final verdict scan the entire `summary.tsv` for any non-PASS row.

**D (Low, with the answer) — the Caldera baseline reference.** `rue-caldera-post1112-baseline/` is not in-repo — it's a session-local measurement artifact from the RUE-1090 activation work (the ~41-minute DNF on the shared-base 100×100 corpus is real, recorded in the RUE-1090 gate evidence). Reword the ledger row to cite the RUE-1090 activation evidence (Linear RUE-1090/RUE-1083) as the baseline's provenance and mark the row "operator confirms baseline at run time" rather than pointing at a directory readers can't find. Same treatment for the `ec99029b` provenance tag (the ratios match the harness formulas regardless — just label them "derived from the harness formulas" or cite the actual emitting run).

**Low/cosmetic:** the 300s Caldera budget could take an env override (post-flip target is seconds, so 300 is fine — but the brief asked for overridability); `stage_9` lacks the local `set -e` fencing its siblings have (fails non-zero either way, just inconsistent).

With A/B/C addressed and D reworded, this is ready — the structure is right and the hard parts (isolation, fail-direction) are already correct.

---

steveklabnik (commented):
Re-review of `9090d3d0`: **all findings fixed, ready to queue.** Verified: stage 1 now runs all three matrices via the `scaling_matrix_` filter (both axes + identity) with the ledger reworded to match; stages 1–2 route through `scripts/rue unit`'s `--list` preflight, and the zero-match path provably exits 1; the terminal verdict scans the full ledger with sequential-stage and exactly-nine-rows checks, ALLOW_FAIL permitted only for pre-flip stage 9, and resume-past-failure is caught (plus resume-without-summary now errors); the baseline row cites the RUE-1090 activation evidence with operator confirmation; `RUE_1091_CALDERA_BUDGET_SECONDS` override validated and documented with authoritative-run guidance; `stage_9` fenced; the unverifiable provenance tag replaced with the harness-formula derivation. Inter-revision diff is confined to the two files with no softened assertions. Marking ready and enabling auto-merge — this is now the sign-off instrument.

---

PR #1954, opened 2026-07-24
RUE-1091: add const identity flip-prep

steveklabnik:
## Summary

- add exact `(declaring_file, source_name)` const-declaration lookup to `BodyRirIndex`
- add durable const assembly with mint-once deduplication and poison-on-partial-failure behavior
- add a LIVE epoch versus pool `ConstInfo` differential using production adapters
- document the module-binding and session-local anonymous-producer refusal boundaries

## Impact

This lands inert pre-flip machinery only. Production analyzer call sites remain unchanged, while the future provider path can assemble value-constant information from durable facts plus the exact current-RIR declaration span.

Refs RUE-1091.

## Validation

- `./buck2 test //crates/rue-air:rue-air-test`
- `./buck2 test //crates/rue-compiler:rue-compiler-test`
- `./buck2 test //:type-architecture-inventory-validation //:body-analysis-capability-inventory-validation`
- `bash ./clippy.sh`
- `scripts/rue fmt`


steveklabnik:
Adversarial review (read-only, contract-level vs base 0dfb994b): **approve-as-is** — no L1/L2 findings. Verified: the `(FileId, Spur)` lookup is the epoch's exact `const_resolutions` storage key with first-wins matching the epoch's source-order pick; mint/dedup/poisoning is a faithful twin of the callable family (`MissingConst` correctly not poisoned, matching the `MissingCallable` precedent; nested type failures poison; zero publication on failure); the differential's epoch side is the live production binding path with span included in the comparison, honoring the span-source rider by assertion; module-binding and session-local refusals fail closed per the r6b rider; inert pre-flip with no inventory-guard impact; commit hygiene clean.

Three L3 doc nits, the first two worth fixing before undrafting:

1. `BodyRirIndex` header doc now claims `one_body.rs` consumes the const handle through `EpochFacts` alongside the three real ops — no EpochFacts const op exists and nothing const-related is consumed pre-flip. Reword the fourth item as pool-side flip-prep so flip work doesn't hunt for a nonexistent epoch twin.
2. `DurableConstSource` doc says returning `None` "poisons that const mint" — true for `function_name` (`MissingConstCallable`), deliberately false for `constant` (`MissingConst`, not poisoned). The behavior is right; the doc overstates it.
3. Optional one-line note at the index: it indexes every arena `ConstDecl` while the epoch's authority is the shell-bound candidate set; these coincide today and the durable-record join makes a stray handle inert (span-only), same accepted-risk shape as the 6e593b2d named-method precedent.

Two operational notes: (a) this branch will conflict in `body_identity.rs` once #1953 (slice r6c) merges — rebase over trunk after that lands; (b) when undrafting, make sure the LAST push happens while the PR is non-draft — draft-guarded CI jobs skip otherwise and `ready_for_review` alone triggers nothing, which is the dead zone that evicted #1946 from the merge queue. Suite claims remain CI-verifiable; they'll be confirmed on that post-undraft push.

---

PR #1955, opened 2026-07-24
RUE-1091: add provider module registry flip prep

steveklabnik:
## Summary

- add a body-local module registry backed by durable module identities
- answer endpoint, call-resolution, and aggregate module-definition consults from registered facts
- compare each provider projection with the LIVE epoch and pin deduplication plus conflict refusal
- document flip-time provenance for every provider overlay registration entry point and retain honest STOPs where durable declaration truth is still absent

## Scope

This is inert flip preparation and does not switch production body analysis to the provider path. It is independent of #1954.

Refs RUE-1091.

## Validation

- `./buck2 test //:type-architecture-inventory-validation //:body-analysis-capability-inventory-validation`
- `scripts/rue quick`
- `./clippy.sh`
- `scripts/rue fmt`

### DIFF
--- a/crates/rue-air/src/sema/aggregate_resolution.rs
+++ b/crates/rue-air/src/sema/aggregate_resolution.rs
@@ -13,6 +13,7 @@ use rue_span::FileId;
 
 use super::body_identity::{BodyIdentityPool, DurableNominalSource};
 use super::context::{ConstValue, LocalVar};
+use super::provider_module_registry::ProviderModuleRegistry;
 use super::{ConstInfo, DeclarationPhase, Sema};
 use crate::intern_pool::TypeInternPool;
 use crate::types::{EnumId, ModuleId, StructId, Type};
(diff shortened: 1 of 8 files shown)


steveklabnik:
Reviewed (inline, contract-level vs trunk `dac91232`): **approve** — no blockers.

Verified: the registry dedups on the durable module handle and fails closed on both conflict directions (mismatched file/path for a known durable, and a second durable claiming a registered file); all three driver projections (`module_endpoint`/`module_id_for_file`, `module_def`, `AggregateFacts::module`) are compared against the LIVE epoch registry in the differential with dedup and conflict refusal pinned; the change is genuinely inert — every `register_module` sits on a Provider\* driver whose construction sites are all cfg(test), and `ProviderModuleRegistry` is `pub(in crate::sema)` with no production caller. The fill-source provenance section satisfies the plan obligation: one entry per `register_*` entry point with its flip-time source and the pinning test, and the STOPs (the `value_const`/`module_binding` composition deferred to the flip; `source_path` explicitly not claimed as declaration truth) are honest and match the code.

Two L3 nits, neither blocking:

1. `ProviderModuleRegistry::register` uses `self.defs.get(id.index() as usize)?` on the dedup path, so an internal invariant violation (a durable key mapped to a missing def) surfaces as an ordinary conflict refusal. It fails closed either way, but the two conditions are worth distinguishing if this ever grows a diagnostic.
2. The differential pins the durable→conflicting-file refusal; the mirror case (a second durable handle claiming an already-registered file, the `by_file` guard) isn't exercised. Cheap to add alongside the existing assertion.

Operational note: when you undraft, make sure the **last push happens while the PR is non-draft** — draft-guarded jobs skip, and `ready_for_review` alone triggers nothing, which is the dead zone that evicted #1946 from the queue. Also expect a small rebase over #1956 (side-B/rFinal) if that lands first: it moves the shared `DurableDeclSource` adapter and render helpers into a cfg(test) `test_support` module in `revisioned_query_database.rs`, which is the same region your differential edits.

---

PR #1958, opened 2026-07-24
RUE-1091: close provider flip prerequisites

steveklabnik:
## Summary

This is the preparatory change immediately before the RUE-1091 production cutover.

- give the endpoint, call, and aggregate provider drivers one shared body-local type/module identity context
- assemble value constants and module bindings once and install them into the call and aggregate fact overlays
- classify associated functions by nominal ownership instead of `self`
- carry named destructor metadata into the provider pool, derive the canonical destructor symbol, and verify needs-drop after containment finalization
- add an instance-keyed producer-facts operation so empty-argument specialization wrappers are preserved
- tighten the rFinal harness so the three named blocker rows are constructed and backed by live tests
- remove the stale declaration-shell and body-produced-anonymous edge exceptions

The production body analyzer still uses the epoch path in this PR. That keeps the following cutover focused on wiring, guard tightening, and epoch deletion instead of discovering identity gaps during the flip.

Refs RUE-1091.

## Blocker closures

- `instance_keyed_producer_body_facts`: added the instance-keyed operation. This was chosen over canonicalizing the family filter because the provider must query the exact producer instance already carried by the declaration projection.
- `const_candidate_facts`: the endpoint driver composes exact RIR declaration metadata with the shared module registry; call and aggregate drivers accept the resulting body-local `ConstInfo`.
- `pool_drop_metadata_parity`: durable named nominals carry destructor presence, the pool derives the canonical qualified `.__drop` symbol, copyability is disabled, and rFinal verifies both the symbol and finalized needs-drop result.

Associated functions now use the same nominal-member path as receiver-taking methods, while preserving `has_self = false`.

STOPs: none.

## Structural gate

The gate intentionally still reads `ACTIVATE RUE-1091`, because this preparatory PR does not change the production analyzer:

```text
normal:
projection exact_ratio_slope_numerator=+9180900
install exact_ratio_slope_numerator=+9180900
endpoints exact_ratio_slope_numerator=+18361800

large:
projection exact_ratio_slope_numerator=+100989900
install exact_ratio_slope_numerator=+100989900
endpoints exact_ratio_slope_numerator=+201979800
```

The next PR must make all six rows exactly flat as part of the production cutover.

## Validation

- `scripts/rue quick` — 36 targets passed; both differential corpora reported zero disagreements
- `./buck2 test //crates/rue-air:rue-air-test //crates/rue-compiler:rue-compiler-test //:type-architecture-inventory-validation //:body-analysis-capability-inventory-validation`
- focused rFinal, const, aggregate-selection, edge-set, and containment-finalization tests
- `scripts/rue fmt` (unrelated `rue-test-runner` reformat excluded)
- normal and large RUE-1090 structural matrices


steveklabnik:
Adversarial review (diff + base-source inspection vs `6c1edaa0`; no buck2 run from this checkout, so all suite/gate claims are CI-verifiable). **Verdict: needs-fix-round — 2×L1, 5×L2.** The architecture is right and the riskiest claim is genuinely true; what's missing is proof and ledger bookkeeping, which is exactly what this PR exists to get right before the cutover.

**Verified clean, end-to-end:** inertness — every new/changed path traces to a `cfg(test)` boundary (`CompilerBodyFactProvider` impls at `revisioned_query_database.rs:11648,11887`; `test_support` at `:13408`; `ProviderTypeFacts` `:12402`; `SignatureFacts` `:13050`), and `DurableNominal` has **no** production constructor (only `body_identity.rs:772` destructuring, two rue-air unit tests, and cfg(test) `test_support:13587`), so the destructor/copyability logic is unreachable in a non-test build. Destructor symbol derivation is byte-identical to the epoch's *final* form (`body_identity.rs:845` vs `typeck.rs:1860-1862`, after `requalify_colliding_destructor_symbols` at `:1808-1826`; the pool republishes symbol paths at `:899-912` before minting, so `struct_symbol_name` is already final). The associated-function rule `*has_self || key.owner().is_some()` is equivalent to `kind().requires_owner()` and the owner-shape invariant is enforced two-way at admission (`bound_definitions.rs:1166`, `binding_manifest.rs:1802`) — a `Function` key structurally cannot carry an owner, so no misclassification is reachable. The r4b-3 zero-edges invariant survives (`ProviderAggregateFacts` still holds no provider handle). The edge-exception removals are earned: the difference loop still panics on any unrecorded family difference and the exact-set non-vacuity assertion at `rfinal_harness.rs:2400-2464` is intact. Gate rows are internally consistent (endpoints = 2× projection, projection == install, both matrices).

**Mandatory fixes:**

1. **L1-A — the destructor symbol assertion is a tautology, and the gap row points at the wrong test.** `rfinal_harness.rs:1420-1434` recomputes `format!("{}.__drop", pool.struct_symbol_name(...))` on the same pool and same `StructId` the pool used at `body_identity.rs:845` — it can only fail if `set_struct_destructor` is deleted outright; it cannot detect a wrong spelling, which is exactly what the retired gap text called out. Compounding: `CLOSED_GAP_LIVE_TESTS["pool_drop_metadata_parity"]` → `side_b_finalize_containment_metadata_freeze_hook` (`:459-462`), a test unchanged by this PR that uses the `method` corpus case (no `__drop`) and would pass with the entire destructor closure reverted. Failure scenario: someone simplifies `:845` to the un-requalified `format!("{}.__drop", name)` spelling from `declarations.rs:1272` — both this assertion and the `needs_drop` check still pass, and the flip ships symbols that don't link against drop glue (`drop_glue_names.rs:18`). **Fix:** compare against `BoundSema`'s live `StructDef::destructor` for a destructor-bearing nominal (the endpoint-const-info test at `:22926+` shows the LIVE-differential pattern), and repoint the gap row at a test that actually mints a destructor.

2. **L1-B — "STOPs: none" contradicts the repo's own ledger.** `docs/notes/rue-1091-analyzer-rewire-plan.md` is untouched — the only slice in this arc that doesn't amend it. Its rider still states verbatim that `value_const` / `module_binding` are "**not yet registered** into the aggregate/call drivers" and that the aggregate const-arm differential "continues to pin provider `Absent` versus epoch `Const`" — both made false by this PR. The same rider opens with "**Every** public `register_*` entry point … has one source" and enumerates each with its pinning differential; this PR adds four (`aggregate_resolution.rs:445,452`, `call_resolution.rs:382,397`) with no inventory entry and no named pinning differential. **Fix:** retire the false STOP, add the four entry points to the fill-source inventory, and record the L2-E coupling below.

3. **L2-D — no LIVE epoch differential for associated functions.** `provider_call_facts_method_info_matches_epoch` (`:21569-21643`) is a genuine cross-path proof for a `self`-taking method; nothing analogous was added for an associated function, though the PR removes the `associated_function_pool_identity` exception row and asserts "preserving `has_self = false`". The only new coverage (`rfinal_harness.rs:1574-1598`) pushes `has_self` into a fact string compared forward-vs-reverse **against itself** (`:2359-2363`) — permutation stability, not epoch parity. **Fix:** add a `Counter.make`-shaped sibling test.

4. **L2-C — the aggregate const arm is pinned on tag only, over a hand-fabricated value.** `provider_aggregate_facts_selection_order_matches_epoch` (`:23260-23268`) registers a literal `ConstInfo` (including a `Span::with_file(file, 0, 0)` that is not the epoch's span) and then asserts only `assert_eq!(ml_tag, el_tag)`. A driver storing `is_pub`/`span` wrongly passes. The endpoint side *is* properly pinned (`:23074` vs `epoch_module_binding_info`), so the composition is real — it's the sink that's unpinned. **Fix:** source the `ConstInfo` from `ProviderEndpointFacts::const_info` over the shared identity (as `run_body_replay` does at `:1650-1665`) and compare the render, not the tag.

5. **L2-G — the one non-tautological destructor assertion still uses the loose match this PR fixed elsewhere.** The predicate was correctly tightened from name-only to module+kind+name in two places (`rfinal_harness.rs:1410-1416`, `revisioned_query_database.rs:13598-13605`) but left at `owner.name() == display` in the third (`rfinal_harness.rs:1956`) — the only place that actually catches a destructor regression. Three copies, two rules; on a multi-module corpus a same-named nominal without a destructor produces a spurious failure, and the match works only because `minted` happens to store `key.name()`. **Fix:** one helper, module+kind+name.

6. **L2-F — the shared pool converts three local invariants into one unenforced global one.** (i) `with_type_pool` holds a `Ref` across the caller's closure (`body_endpoint.rs:836`, `aggregate_resolution.rs:570`, `call_resolution.rs:362`), so any `pool_mut()` reached from *another* driver inside it — `aggregate.struct_in_file`, `calls.function_info`, `endpoint.mint_anonymous` — is now an `already borrowed` panic. This was structurally impossible at base (three separate `RefCell`s; note `function_info`/`method_info` moved `&mut self` → `&self` at `call_resolution.rs:441,452`), and `with_type_pool` is `pub` with a caller-supplied closure, so it's externally reachable. The doc at `body_endpoint.rs:487-494` still asserts the old per-driver property. (ii) `finalize_containment_metadata` sets no frozen flag, and `set_struct_destructor` is the only caller of `invalidate_containment_metadata` (`intern_pool.rs:2347`, `:648-650`) — so a lazy mint through the aggregate or call driver *after* the endpoint's freeze silently wipes every containment fact or leaves the new type unanswerable. Both fail closed (`try_type_needs_drop` → `None`), so not unsound, but the r4a-2a freeze contract is now cross-driver caller discipline with no enforcement and no test. **Fix:** a `frozen: Cell<bool>` that refuses post-freeze mints, or an explicit recorded obligation plus a test that mints through a second driver after freeze; and correct the stale comment.

**L2-E, recommended (record, don't necessarily re-engineer):** two canonicalization conventions now coexist. The pool collapses on entry (`with_collapsed_empty_specializations`, recursive) while the harness **re-inflates unconditionally** at `rfinal_harness.rs:1707-1712`, including when `producer_instances` already carried a bare `Definition`. I verified the inflated shape matches what production mints (`semantic_identity.rs:136-142`), so it's correct today — but the gap text documenting the coupling was deleted and nothing replaced it, and the re-inflation adds exactly one level where the collapse is recursive. Also: the dedup at `:1722` is *definition*-keyed while the op is *instance*-keyed, so a specialized producer with two argument sets gets zero coverage of the capability the op was added for.

**L3:** stale `declaration-shell` justification comment at `:496-498`; prefer `kind().requires_owner()` over open-coding at `:13628` (`semantic_identity.rs:439-441` exists to be the one source); the side-B call-driver oracle silently widened to the endpoint's `.with_anonymous_nominals(...)` source (`:1993-2004`) — probably desirable, unremarked; `CLOSED_GAP_LIVE_TESTS` is a string table with no compile-check, which is what let L1-A through; `SIDE_B_FACT_EXCEPTIONS` has no non-vacuity check, so 1 of 2 rows is permanently un-hit.

Hygiene clean: single trailer-free commit; `ProviderIdentityContext` needs no type-architecture allowlist entry; `api_inventory` visibility requirements respected.

---

PR #1959, opened 2026-07-24
RUE-1091: discover provider body demand from RIR

steveklabnik:
## Summary

- add a rue-air body-analysis entry that walks only the requested RIR body and emits demand through a provider-owned context
- drive calls, types, nominals, methods, operators, constants, and anonymous producers through keyed provider operations and the body-local overlay
- repoint rFinal side B at that entry, keeping production transactions comparison-only
- assert discovered definition-root parity, exact edge-family accounting, schedule invariance, cancellation/eviction safety, and immunity to unrelated declaration growth

## Split boundary

This is Flip-A only. Production body analysis remains epoch-backed, `PROVIDER_ONLY_MODE` remains false, and the RUE-1090 gate remains `ACTIVATE`. Flip-B will wire this entry into production and remove the epoch path.

Refs RUE-1091.

## Validation

- `scripts/rue quick` (36 targets passed)
- `//crates/rue-air:rue-air-test` (628 passed)
- `//crates/rue-compiler:rue-compiler-test` (828 passed)
- body-analysis capability inventory validation and tool tests (zero unaccounted authority)

### DIFF
--- a/crates/rue-air/src/lib.rs
+++ b/crates/rue-air/src/lib.rs
@@ -24,6 +24,7 @@ pub mod layout;
 mod module_registry;
 mod param_arena;
 mod path_norm;
+mod provider_body_analysis;
 mod runtime_call;
 mod scope;
 mod sema;
@@ -68,6 +69,7 @@ pub use layout::{Layout, LayoutKind, PaddingRange, SLOT_BYTES};
 pub use module_registry::ModuleRegistry;
 pub use param_arena::{ParamArena, ParamRange};
 pub use path_norm::{mangle_symbol_component, normalize_module_path, unmangle_symbol_component};
+pub use provider_body_analysis::{ProviderBodyAnalysisContext, analyze_provider_body};
 pub use runtime_call::{
     OptionVariant, RuntimeAirArgument, RuntimeAirType, RuntimeCallActivation, RuntimeCallKind,
     RuntimeOperandOrigin,
--- a/crates/rue-air/src/provider_body_analysis.rs
+++ b/crates/rue-air/src/provider_body_analysis.rs
@@ -1,0 +1,365 @@
+//! Body-local demand discovery for provider-driven semantic analysis.
+//!
+//! This is the pre-cutover entry used by the RUE-1091 differential harness.
+//! Unlike a replay of an already-published body transaction, it starts from one
+//! RIR body root and discovers semantic facts while walking that body.  The
+//! caller owns the provider/overlay context and materializes each demand through
+//! the keyed [`ProviderBodyAnalysisContext`] callbacks.
+
+use std::collections::HashSet;
+
+use lasso::ThreadedRodeo;
+use rue_rir::{InstData, InstRef, RepeatCount, Rir, RirPatternView};
+use rue_span::Span;
+
+use crate::OperatorName;
+
+/// Provider/overlay operations demanded by a body-local RIR walk.
+///
+/// The context, rather than the walker, owns the provider's durable identity
+/// types. This keeps the entry in `rue-air` while allowing the compiler adapter
+/// to materialize exact declaration signatures, nominals, methods, constants,
+/// and imports into its task-owned overlay.
+pub trait ProviderBodyAnalysisContext {
+    /// A value-position name was referenced by the body.
+    fn demand_value(&mut self, name: &str, span: Span);
+
+    /// A call head and its body-local argument expressions were referenced.
+    fn demand_call(&mut self, name: &str, arguments: &[InstRef], span: Span);
+
+    /// A type-syntax spelling was referenced by the body.
+    fn demand_type(&mut self, syntax: &str, span: Span);
+
+    /// A struct literal head was referenced. `module` is the body-local RIR
+    /// expression that names a qualified module, when present.
+    fn demand_struct(&mut self, module: Option<InstRef>, type_name: &str, span: Span);
+
+    /// An enum variant head was referenced.
+    fn demand_enum(&mut self, module: Option<InstRef>, type_name: &str, span: Span);
+
+    /// A method was referenced on a body-local receiver expression.
+    fn demand_method(&mut self, receiver: InstRef, method: &str, span: Span);
+
+    /// A user-overloadable operator was referenced on a body-local receiver.
+    fn demand_operator(&mut self, receiver: InstRef, operator: OperatorName, span: Span);
+
+    /// A named array repeat count was referenced.
+    fn demand_const_value(&mut self, name: &str, span: Span);
+
+    /// An anonymous nominal syntax node was reached in this body.
+    fn demand_anonymous_nominal(&mut self, instruction: InstRef, span: Span);
+}
+
+/// Walk exactly one RIR body and materialize every semantic demand originating
+/// in that body.
+///
+/// The walk follows only instruction edges reachable from `body`; it never
+/// enumerates declarations, module indexes, or an in-scope symbol universe.
+pub fn analyze_provider_body(
+    context: &mut impl ProviderBodyAnalysisContext,
+    rir: &Rir,
+    interner: &ThreadedRodeo,
+    body: InstRef,
+) {
+    struct Walker<'a, C> {
+        context: &'a mut C,
+        rir: &'a Rir,
+        interner: &'a ThreadedRodeo,
+        visited: HashSet<InstRef>,
+    }
+
+    impl<C: ProviderBodyAnalysisContext> Walker<'_, C> {
+        fn symbol(&self, symbol: lasso::Spur) -> String {
+            self.interner.resolve(&symbol).to_owned()
+        }
+
+        fn visit(&mut self, reference: InstRef) {
+            if !self.visited.insert(reference) {
+                return;
+            }
+            let inst = self.rir.get(reference);
+            let span = inst.span;
+            macro_rules! visit {
+                ($($child:expr),* $(,)?) => {
+                    {
+                        $(self.visit($child);)*
+                    }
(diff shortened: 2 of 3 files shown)


steveklabnik:
Heavy adversarial review (diff + base-source inspection vs `1863a50f`; no buck2 from this checkout, so suite claims are CI-verifiable). **Verdict: needs-fix-round, and I recommend an approach change first — see "Architectural redirect" at the end, which I think makes most of the L1 list disappear rather than requiring it to be built.**

Green CI is not evidence here: the assertions pass **because** the corpus avoids the shapes the walk cannot handle. Confirmed-good first: the entry is layered correctly (lives in `rue-air`, takes only `&Rir`/`&ThreadedRodeo`/`InstRef`/caller context, no epoch type in its signature, reads no `Sema`); the `InstData` match is exhaustive with no wildcard, so a new RIR variant breaks the build; operator coverage is right against the real `OperatorName` surface (`Ne`→`Eq` correct, `Mod`/`Lt`/`Neg` genuinely absent from the boundary); the edge-difference loop still panics in both directions with exact-set non-vacuity; schedule invariance permutes against a warm database (real, not self-comparison) and cancellation/eviction now run through the new entry; freeze ordering is correct (`exercise_finalize_seam` after all mints) so the mint-after-freeze hazard isn't reached; inertness verified (`canonical_semantic.rs:1947,2467` still call `analyze_one_body_instance`, `PROVIDER_ONLY_MODE` false, no production caller); commit hygiene clean.

**L1-1 — the entry is given the body block, not the declaration.** `rfinal_harness.rs:2465-2478` unwraps `body` out of `FnDecl`/`DropFnDecl` and passes that, so the walker's `FnDecl` arm (`provider_body_analysis.rs:180-193`) — the only place parameter and return types are demanded — is unreachable for the requested body. `struct Point { x: i64 } fn f(p: Point) -> i64 { 0 }` discovers nothing while production references `Point`. The corpus survives because its only nominal-typed parameter case is primitive (`make(value: i32)`) and `get(self)`'s owner comes from the pre-seed at `:2459-2474`.

**L1-2 — no lexical scope model, violating the ratified r0 precedence.** `provider_body_analysis.rs:95-98` calls `demand_value` for *every* `VarRef` including locals and comptime params; the context resolves the name in module scope as a `Const` (`:2092-2105`). The walker sees `Alloc { name }` and `FnDecl { params }` but tracks no bindings. Base plan `:880-883` ratifies the opposite ("comptime value parameters win over file-level constants") and `:906-909` forbids exactly this ("no fuzzy 'const named X' abstraction"). `const value: i32 = 5; fn main() -> i32 { let value = 1; value + 1 }` over-discovers; a `comptime N` against a module `const N` resolves the wrong one. This reintroduces the r0 global-uniqueness hazard inside the flip's own analyzer.

**L1-3 — receiver types are guessed by name.** `rfinal_harness.rs:1995-2043` handles a literal `StructInit`, bare `self`, and a `VarRef` *whose name is looked up as a struct*; everything else returns `None` and `demand_method`/`demand_operator` then return with no fact, no demand, no exception row. `let c = C { v: 1 }; c.get()` never discovers `C.get`. The same name-guess also over-discovers: `struct v {…}` plus `let v = 1; v + 2` mints and records `Type(v)`. There is no type inference in the entry, so no expression-typed receiver is discoverable at all.

**L1-4 — root parity is a lossy projection satisfiable by near-zero discovery, and it is masking a live divergence today.** The operator is a genuine exact set equality (`:2812`), but `reference_definition_roots` (`:933-1004`) collapses `Nominal(Anonymous(identity))` to its producer root and drops primitives/array lengths/specialization arguments. On the arc's keystone case `anonymous_producer::main`, side B discovers `{Callable(Definition(Pair))}` and misses the anonymous nominal entirely — production has the nominal plus specialization plus `i32` — yet both project to `{Pair}` and the assertion passes. Also, a body with all-primitive demand compares ∅ == ∅.

**L1-5 — the anonymous-identity arms are now dead on side B and the exception row was deleted as if fixed.** `discovered_references` can only contain `Type(Nominal(Named))`, `Callable(Definition)`, `Definition`, so `replay_type`'s anonymous/slice/generic/module arms and `replay_producer` (`:1737-1810`) are unreachable for the whole corpus. That is why `comptime_call_reduces_to_anonymous_nominal` stopped being hit and was removed from `SIDE_B_FACT_EXCEPTIONS` (`:479`) and `EXPECTED_FACT_EXCEPTION_HITS` (`:2719`). Presented as a tightening, this is coverage loss, and it regresses recorded carry-forward #3 (base plan `:355-359`: enumerate every `TypeInstanceKey` arm, "never rely on incidental corpus coverage, or an arm can be vacuously green"). `identity_variant_inventory_is_exhaustive_and_constructible` doesn't substitute — it pattern-guards constructed values and never pushes them through a driver.

**L1-6 — `producer_instance_body_facts(body_instance)` is an unconditional prefetch of the body's own production output, and the call is impossible in production.** `:2379-2385`, first statement of `run_body_replay`, before any demand exists. Its evaluator returns production's `BodyTransaction::Success { produced_anonymous_nominals }` verbatim when memoized and only reconstructs from declaration shells on a memo miss — the harness's fresh database always misses, which is exactly why the `compiler.declaration-shell` ProviderOnly row appeared (`:514-521`). So the leak is latent, not live, but nothing asserts that, and the file header (`:16-19`) and `ProductionDemandPeer` doc (`:568-575`) claim otherwise unconditionally. Worse for Flip-B: production guards this query against its own instance as a self-cycle (`revisioned_query_database.rs:9457-9462`), so the flip cannot make this call at all.

**L2s (condensed):** comptime producer instances are mis-keyed with `CanonicalArguments::default()` even when the call had real arguments, and the result is discarded (`:2210-2219`) — the opposite-conventions hazard resolved the wrong way; the immunity test adds exactly **one** unrelated declaration in a *different* module, doesn't compare `discovered_references`, and can't see O(U) *work* — `BodyRirIndex::new` builds a whole-program `RirDeclarationIndex` and is constructed twice per body (`:2386`, `:2392`) plus a linear `decls` scan at `:2588-2591`, i.e. edge-immune while work stays linear in the declaration universe, which is precisely what keeps the gate non-flat; the absent-lookup assertion lost its type-namespace half (base `:2029-2046` → substring check at `:2543-2551`); `!body.facts.is_empty()` is vacuous because a walk always pushes one summary string (`:2484-2487`); silent no-ops throughout the context impl (`:2225`, `:2232`, `:2250`, `:2270`, `:2303`, `:2317-2320`) take no exception row, so the table's no-silent-tolerance property doesn't hold for the discovery layer; whole demand classes have no arm and no STOP (`Try`/`?` never calls `language_item` so r6c's obligation is untested, `StringConst` no-ops despite r6b's `Str(N)`, `FieldGet`/`FieldSet`/`OffsetOf`/`Match` variant demand nothing so `ProviderAggregateFacts` gets zero body-derived demand, `Intrinsic` ignored, `resolve_import` never called, no drop points); module qualification is dropped (`_module` ignored, `lookup_qualified` never called, associated-call owner resolved only as `Struct`, never `Enum`); no plan rider (first slice in the arc not to amend it); all nine context methods bind `_span`, so the span-source obligation is undischarged; and `provider_body_analysis.rs` — the file that becomes the production analyzer — is outside the capability guard's inventory list (`validate-body-analysis-capabilities.py:144-156`), where adding it is trivially green today and would lock the property in.

---

## Architectural redirect (please read before doing the fix list)

L1-2, L1-3, and much of L2-6 all have the same root cause: this PR builds a **second, simplified body walker** that re-derives demand by name-guessing, alongside the real one. It can't have a scope model or receiver typing without reimplementing analysis — and it shouldn't have to.

Production already has a body analyzer that does all of this correctly: `analyze_one_body_instance` (`crates/rue-air/src/sema/one_body.rs:990`) walks the body with real scope and typing over a `BodySema`, and consults declaration facts **through the seam traits that slices r1a/r1b/r1c hoisted for exactly this purpose** (`BodyEndpointProvider`/`endpoint_facts` at `one_body.rs:19`, plus the call-resolution and aggregate seams). The whole point of those slices was that the fact *source* becomes swappable while the analysis logic stays byte-identical: `EpochFacts` reads the epoch, the `Provider*` drivers read the provider.

So demand discovery is not missing — it already exists, in the real analyzer, correct by construction, because every seam consult *is* a demand. What's missing is a `BodySema` whose declaration-fact backing is the provider + `BodySemanticOverlay` instead of the whole-program epoch. That is what "the analyzer consumes only provider + overlay" has meant in the plan from the start.

Concretely, I'd suggest Flip-A become: construct the body-local semantic context over provider+overlay, run **the existing** `analyze_one_body_instance` against it, and repoint rFinal side B at *that*. Then L1-1 (signature demand), L1-2 (scope/precedence), L1-3 (receiver typing), L1-5 (anonymous arms), and the missing arms in L2-6 are all inherited from production logic rather than rebuilt — and side-B/side-A parity becomes a genuinely strong claim, because both sides run the same analysis code over different fact sources. L1-4's projection can then tighten to reference-level equality without a mountain of new discovery work, and L1-6's prefetch can go away since the real analyzer demands producer facts only for reached producers.

If there's a concrete reason that swap can't work — e.g. `BodySema` is structurally welded to `Sema` in a way that can't be satisfied by the overlay — that's a genuine finding worth reporting as a STOP with file:line evidence, and it would change the plan. But please establish that before building a parallel walker further: a second analyzer is a permanent correctness liability, and Flip-B deletes the only oracle that could reveal a divergence between them.

`provider_body_analysis.rs` is still valuable as the demand-*recording* shim (the context/trait surface is good); it's the walking-and-guessing part that should be replaced by the real analyzer.

---

PR #1960, opened 2026-07-25
Incrementality value-audit instrumentation (fail-closed audit)

steveklabnik:
## Summary

This draft PR is an incrementality value-audit and instrumentation stack. It records current production behavior and fail-closed evidence; it does not enable a provider analyzer, flip production behavior, or claim flip readiness.

Refs RUE-1091, RUE-1090, RUE-1092.

The integrated stack is:

- `b336a008` — existing Clippy iterator-clone correction retained at the fork head
- `741322a9` — O(1) exact-key retained-body counter probe and lifecycle coverage
- `8cf3a129` — complete warm/fresh retained-body identity oracle
- `6c0b7efe` — schema-v2 runner, manifest-authoritative evidence gates, and adversarial tests
- `2964791c` — narrow integration Clippy cleanup in the merged oracle

The production body hot path calls the exact-key `contains_retained_key` probe. The complete retained-key scan is cfg(test)-only and supplies the snapshot oracle, including successful, failed, deleted, renamed, visibility, and imported-body cases.

## Protocol and gates

- Release-profile protocol with one unrecorded warmup and seven alternating paired samples.
- Required evidence is fail-closed: malformed, partial, missing, unsupported, or cross-protocol evidence cannot become a pass.
- Schema v2, exact parity, output-presence checks, production locality counters, exact-identity availability, RSS rollup, Darwin RSS parsing, RUST_LOG hygiene, cached baseline loading, robust partial-row handling, explicit role/source/binary provenance, same-binary classification, Caldera timeout headroom, and the Buck test target are covered.
- Thresholds and locality upper bounds are loaded from and validated against the checked-in manifest.
- A claimed improvement requires the precommitted threshold and three times the larger MAD.
- Cold wall/RSS regression and optional repeated-edit memory evidence remain separately gated.

## Current-production findings

The bounded representative same-binary smoke used one release compiler and one release session-benchmark binary for all three roles. It produced a mechanically valid schema-v2 report classified as `same_binary_protocol_smoke`, with exact parity passing.

- Cold and warm no-op observations passed.
- Warm leaf-body locality failed honestly: production computed 6 semantic bodies, 4 CFGs, and 90 durable-source records against the manifest’s one-unit bounds; the report also retained the corresponding invalidation counters.
- Warm signature-fanout locality failed honestly: production computed 6 semantic bodies, 4 CFGs, and 90 durable-source records against the manifest’s bounded allowance.
- Stable exact identity sets are unavailable in the session-benchmark protocol and are explicitly explained in the evidence.
- Repeated-edit RSS is explicitly unsupported because the reused benchmark has no bounded persistent-session protocol.
- Synthetic and Caldera workloads were not run in this bounded smoke.

The smoke exited nonzero because those current-production value gates failed, not because evidence was malformed. This is protocol/evidence validation only and does not establish a historical/current/candidate value claim.

## Test evidence

- `scripts/rue fmt` completed; unrelated formatter output in `crates/rue-test-runner/src/lib.rs` was removed.
- `bash ./clippy.sh`: passed with no lint violations.
- `scripts/rue unit query exact_retained_key_probe --nocapture`: 1 passed.
- `scripts/rue unit compiler scaling_harness --nocapture`: 21 passed, including correctness, invalidation, scaling, counter lifecycle, failure/deletion/rename/visibility/import oracles.
- `scripts/rue unit compiler-session-bench --nocapture`: 6 passed.
- `./buck2 test //:value-audit-tool-tests`: 20 tests passed.
- `scripts/rue quick`: 36 targets passed, including 836 compiler tests, 839 scaling-matrix tests, 88 oracle-diff tests, and frontend differential coverage.
- Bounded representative same-binary smoke: schema v2 and same-binary classification validated; exact parity passed; value verdict `fail` for the documented locality findings.

### DIFF
--- a/crates/rue-compiler-session-bench/src/main.rs
+++ b/crates/rue-compiler-session-bench/src/main.rs
@@ -840,7 +840,32 @@ where
 
 fn with_parity(mut scenario: Value, evidence: Value) -> Value {
     scenario["differential_parity"] = evidence;
+    scenario["required_vs_reused_work"] = work_projection(&scenario);
     scenario
+}
+
+/// Stable black-box work projection consumed by the value-audit runner.
+/// Keep this schema independent of the detailed counter ledger: the runner
+/// needs only required-versus-reused artifact counts for locality gates.
+fn work_projection(scenario: &Value) -> Value {
+    json!({
+        "modules": {
+            "required": count(scenario, &["parse", "modules_reparsed"]),
+            "reused": count(scenario, &["parse", "modules_reused"]),
+        },
+        "semantic_bodies": {
+            "required": count(scenario, &["semantic_work", "bodies_attempted"]),
+            "reused": count(scenario, &["semantic_work", "durable_bodies", "reused_bodies"]),
+        },
+        "cfgs": {
+            "required": count(scenario, &["semantic_work", "cfg", "builds_attempted"]),
+            "reused": count(scenario, &["semantic_work", "cfg", "reuses"]),
+        },
+        "semantic_queries": {
+            "required": count(scenario, &["queries", "semantic_executions"]),
+            "reused": count(scenario, &["queries", "semantic_reuses"]),
+        },
+    })
 }
 
 fn completion_workloads(functions: usize) -> Vec<Value> {
@@ -1480,6 +1505,18 @@ fn assert_completion_structure(scenarios: &[Value], functions: usize) {
     assert_declaration_epoch_work(cold, 1, 1, 1, 0, 0);
 
     let noop = get("completion_exact_noop");
+    for name in [
+        "completion_exact_noop",
+        "completion_unrelated_edit",
+        "completion_changed_reachable_body",
+        "completion_reverse_closure",
+    ] {
+        let projection = &get(name)["required_vs_reused_work"];
+        for artifact in ["modules", "semantic_bodies", "cfgs", "semantic_queries"] {
+            assert!(projection[artifact]["required"].is_u64());
+            assert!(projection[artifact]["reused"].is_u64());
+        }
+    }
     assert_eq!(count(noop, &["queries", "semantic_executions"]), 0);
     assert_eq!(count(noop, &["queries", "semantic_reuses"]), 1);
 
--- a/crates/rue-compiler-session-bench/src/main.rs
+++ b/crates/rue-compiler-session-bench/src/main.rs
@@ -821,6 +821,10 @@ where
 
     let executable_sha256 = format!("{:x}", Sha256::digest(&reused_executable.elf));
     json!({
+        "schema_version": 1,
+        "status": "pass",
+        "comparison": "cold_vs_reused_in_process",
+        "comparison_completed": true,
         "public_semantic_cfg_artifacts": "exact",
         "public_type_universe": "exact_ordered_entries",
         "durable_specialized_body_payloads": "exact",
@@ -839,31 +843,81 @@ where
 }
 
 fn with_parity(mut scenario: Value, evidence: Value) -> Value {
+    scenario["evidence_schema"] = json!({
+        "version": 2,
+        "differential_parity_version": 1,
+        "locality_work_version": 2,
+    });
     scenario["differential_parity"] = evidence;
     scenario["required_vs_reused_work"] = work_projection(&scenario);
     scenario
 }
 
-/// Stable black-box work projection consumed by the value-audit runner.
-/// Keep this schema independent of the detailed counter ledger: the runner
-/// needs only required-versus-reused artifact counts for locality gates.
+/// Stable production-counter projection consumed by the value-audit runner.
+/// The detailed counter ledger remains available beside this projection, while
+/// these fields give external tooling one typed vocabulary for computed,
+/// reused, and invalidated work.  Identity sets are currently not exposed by
+/// this benchmark, so that absence is explicit rather than inferred.
 fn work_projection(scenario: &Value) -> Value {
+    macro_rules! required_count {
+        ($path:expr) => {
+            match try_count(scenario, $path) {
+                Some(value) => value,
+                None => {
+                    return json!({
+                        "schema_version": 2,
+                        "status": "unsupported",
+                        "reason": "production counter path was absent",
+                    });
+                }
+            }
+        };
+    }
     json!({
+        "schema_version": 2,
+        "counter_source": "production_metrics",
+        "exact_identity_sets": {
+            "available": false,
+            "reason": "CompilerSession does not expose stable recompute/reuse identity sets in this protocol"
+        },
         "modules": {
-            "required": count(scenario, &["parse", "modules_reparsed"]),
-            "reused": count(scenario, &["parse", "modules_reused"]),
+            "computed": required_count!(&["parse", "modules_reparsed"]),
+            "invalidated": required_count!(&["parse", "modules_reparsed"]),
+            "required": required_count!(&["parse", "modules_reparsed"]),
+            "reused": required_count!(&["parse", "modules_reused"]),
+        },
+        "rir": {
+            "computed": required_count!(&["queries", "rir_executions"]),
+            "invalidated": required_count!(&["queries", "rir_executions"]),
+            "reused": required_count!(&["queries", "rir_reuses"]),
+        },
+        "declarations": {
+            "computed": required_count!(&["semantic_work", "declaration_resolution_invocations"]),
+            "invalidated": required_count!(&["queries", "definition_entries_invalidated"]),
+            "reused": required_count!(&["semantic_work", "declaration_reuse", "durable_records_reused"]),
+        },
+        "durable_source": {
+            "computed": required_count!(&["semantic_work", "per_body_declaration_context", "durable_source_records_copied"]),
+            "invalidated": required_count!(&["semantic_work", "body_analyses_invalidated"]),
+            "reused": required_count!(&["semantic_work", "durable_bodies", "reused_bodies"]),
         },
         "semantic_bodies": {
-            "required": count(scenario, &["semantic_work", "bodies_attempted"]),
-            "reused": count(scenario, &["semantic_work", "durable_bodies", "reused_bodies"]),
+            "computed": required_count!(&["semantic_work", "body_analyses_computed"]),
+            "invalidated": required_count!(&["semantic_work", "body_analyses_invalidated"]),
+            "required": required_count!(&["semantic_work", "bodies_attempted"]),
+            "reused": required_count!(&["semantic_work", "durable_bodies", "reused_bodies"]),
         },
         "cfgs": {
-            "required": count(scenario, &["semantic_work", "cfg", "builds_attempted"]),
-            "reused": count(scenario, &["semantic_work", "cfg", "reuses"]),
+            "computed": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "invalidated": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "required": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "reused": required_count!(&["semantic_work", "cfg", "reuses"]),
         },
         "semantic_queries": {
-            "required": count(scenario, &["queries", "semantic_executions"]),
-            "reused": count(scenario, &["queries", "semantic_reuses"]),
+            "computed": required_count!(&["queries", "semantic_executions"]),
+            "invalidated": required_count!(&["queries", "semantic_entries_invalidated"]),
+            "required": required_count!(&["queries", "semantic_executions"]),
+            "reused": required_count!(&["queries", "semantic_reuses"]),
         },
     })
 }
@@ -1145,6 +1199,12 @@ fn count(value: &Value, path: &[&str]) -> u64 {
         .fold(value, |value, key| &value[*key])
         .as_u64()
         .unwrap()
+}
+
+fn try_count(value: &Value, path: &[&str]) -> Option<u64> {
+    path.iter()
+        .try_fold(value, |value, key| value.get(*key))
+        .and_then(Value::as_u64)
 }
 
 fn assert_structure(scenarios: &[Value], modules: usize) {
@@ -1511,10 +1571,21 @@ fn assert_completion_structure(scenarios: &[Value], functions: usize) {
         "completion_changed_reachable_body",
         "completion_reverse_closure",
     ] {
+        assert_eq!(get(name)["differential_parity"]["schema_version"], 1);
+        assert_eq!(get(name)["differential_parity"]["status"], "pass");
+        assert_eq!(
+            get(name)["differential_parity"]["comparison_completed"],
+            true
+        );
         let projection = &get(name)["required_vs_reused_work"];
+        assert_eq!(projection["schema_version"], 2);
+        assert_eq!(projection["counter_source"], "production_metrics");
+        assert!(projection["exact_identity_sets"]["available"].is_boolean());
         for artifact in ["modules", "semantic_bodies", "cfgs", "semantic_queries"] {
             assert!(projection[artifact]["required"].is_u64());
             assert!(projection[artifact]["reused"].is_u64());
+            assert!(projection[artifact]["computed"].is_u64());
+            assert!(projection[artifact]["invalidated"].is_u64());
         }
     }
     assert_eq!(count(noop, &["queries", "semantic_executions"]), 0);
@@ -1883,7 +1954,7 @@ fn main() {
         println!(
             "{}",
             json!({
-                "schema_version": 1,
+                "schema_version": 2,
                 "workload": "representative_compiler_session_scenarios",
                 "fixture": "benchmarks/scenarios/representative/main.rue",
                 "measurement_scope": "compiler_build_and_query_performance_not_runtime",
@@ -1910,7 +1981,7 @@ fn main() {
     println!(
         "{}",
         json!({
-            "schema_version": 11,
+            "schema_version": 12,
             "workload": "compiler_session_invalidation",
             "configuration": {
                 "modules": config.modules,
--- a/crates/rue-compiler-session-bench/src/main.rs
+++ b/crates/rue-compiler-session-bench/src/main.rs
@@ -26,6 +26,97 @@ const DEFAULT_MODULES: usize = 128;
 const DEFAULT_WARMUP: usize = 3;
 const DEFAULT_ITERATIONS: usize = 10;
 
+const AVAILABLE_CONTEXT_COUNTERS: &[(&str, &[&str])] = &[
+    (
+        "cold_body_preparations",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "cold_body_preparations",
+        ],
+    ),
+    (
+        "declarations_inspected",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "declarations_inspected",
+        ],
+    ),
+    (
+        "modules_registered",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "modules_registered",
+        ],
+    ),
+    (
+        "rir_indexes_constructed",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "rir_indexes_constructed",
+        ],
+    ),
+    (
+        "rir_instructions_visited",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "rir_instructions_visited",
+        ],
+    ),
+    (
+        "durable_source_records_inspected",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "durable_source_records_inspected",
+        ],
+    ),
+    (
+        "durable_source_records_copied",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "durable_source_records_copied",
+        ],
+    ),
+    (
+        "shells_prepared",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "shells_prepared",
+        ],
+    ),
+    (
+        "projections_performed",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "projections_performed",
+        ],
+    ),
+    (
+        "semantics_installed",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "semantics_installed",
+        ],
+    ),
+    (
+        "endpoints_installed",
+        &[
+            "semantic_work",
+            "per_body_declaration_context",
+            "endpoints_installed",
+        ],
+    ),
+];
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct Config {
     modules: usize,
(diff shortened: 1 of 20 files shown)


crates/rue-compiler-session-bench/src/main.rs
```diff
@@ -839,10 +843,85 @@ where
...
+            "invalidated": required_count!(&["semantic_work", "body_analyses_invalidated"]),
+            "required": required_count!(&["semantic_work", "bodies_attempted"]),
+            "reused": required_count!(&["semantic_work", "durable_bodies", "reused_bodies"]),
+        },
+        "cfgs": {
+            "computed": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "invalidated": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "required": required_count!(&["semantic_work", "cfg", "builds_attempted"]),
+            "reused": required_count!(&["semantic_work", "cfg", "reuses"]),
+        },
```
steveklabnik, line 1019: Three of the seven locality counters report `invalidated` as a verbatim duplicate of `computed`, and `durable_source` borrows two of its three fields from unrelated counters:

- `modules.invalidated` == `modules.computed` == `modules.required` == `parse.modules_reparsed`
- `rir.invalidated` == `rir.computed` == `queries.rir_executions`
- `cfgs.invalidated` == `cfgs.computed` == `cfgs.required` == `cfg.builds_attempted`
- `durable_source.invalidated` = `semantic_work.body_analyses_invalidated` (body-analysis invalidation, not durable-source-record invalidation)
- `durable_source.reused` = `durable_bodies.reused_bodies` — the exact same value as `semantic_bodies.reused`

This matters because of the new gate in `locality_check`:

```python
if values["computed"] > maximum or values["invalidated"] > maximum:
```

whose comment says checking both is what lets the gate "detect a full-program recompute." For `modules`, `rir`, and `cfgs` the second half of that disjunction is arithmetically identical to the first, so it contributes no detection power — the gate is exactly as strong as `computed > maximum` alone. The schema reads as though seven artifacts each carry an independent invalidation signal when only `declarations`, `semantic_bodies`, and `semantic_queries` actually do.

`durable_source` is the sharper problem: someone reading a published audit report will reasonably take `durable_source.invalidated` as durable-source-record invalidation. It isn't.

The pattern to reuse is the one already applied correctly a few lines up:

```rust
"exact_identity_sets": {
    "available": false,
    "reason": "CompilerSession does not expose stable recompute/reuse identity sets in this protocol"
},
```

Emit the fields that have real production counters, and mark the rest explicitly unavailable with a reason rather than aliasing a nearby counter. `locality_check` can then skip absent counters instead of applying a bound to a value that cannot fail independently.

Minor, same function: `required` is present on `modules`, `semantic_bodies`, `cfgs`, and `semantic_queries` but absent on `rir`, `declarations`, and `durable_source`. Worth making uniform while the schema is at `schema_version: 2`.

---

crates/rue-compiler/src/scaling_harness.rs
```diff
@@ -1562,6 +2153,316 @@ fn invalidation_negative_lookup_becoming_positive_invalidates_consumers() {
...
+        .update(&unrelated_module_snapshot(0))
+        .into_result()
+        .and_then(|_| unrelated_session.canonical_semantic(&options))
+        .expect("deletion lifecycle fixture compiles");
+    let deleted_body = deleted_output.work().body_analysis;
+    assert_eq!(deleted_body.body_analyses_computed, 1);
+    assert_eq!(deleted_body.body_analyses_reused, 0);
+    assert_eq!(
+        deleted_body.body_analyses_invalidated, 0,
+        "deletion reaches a fresh root key, so it has no retained predecessor"
```
steveklabnik: Question about the asymmetry this test pins, and whether the explanatory comment is right.

Within one session (`unrelated_session`), the same `main.rue` text goes through:

- `0 -> 1` module: `computed == 0`, `reused == cold_computed` — full reuse, as expected.
- `1 -> 0` module: `computed == 1`, `reused == 0`, `invalidated == 0`.

So adding an unrelated module reuses `main`'s body analysis, but removing that same module recomputes it *and* reports no retained predecessor. `invalidated == 0` means `has_retained_body_key` returned false for `main`'s key at that point.

I chased the obvious explanation and it doesn't hold: `StableDefinitionKey` is `{module, namespace, kind, name, owner}` (`bound_definitions.rs:108`), and `main.rue` keeps `FileId::DEFAULT` across all three snapshots, so the key ought to be identical in all of them. If that's right, "deletion reaches a fresh root key" isn't what's happening, and the two candidate explanations are both worth knowing about:

1. The body key genuinely does vary with the unrelated-module universe — which is precisely the locality property this stack exists to measure, and would deserve its own gate rather than an assertion recording it.
2. `main`'s memo node was evicted between revisions (retention limit), in which case `body_analyses_invalidated` under-reports whenever eviction beats the probe, and the counter's meaning is "had a retained predecessor *that survived eviction*."

Either way the current comment asserts a mechanism that I don't think the key structure supports. If it's (2), that's worth a sentence in the `has_retained_body_key` doc comment too, since it bounds what the counter can be used to prove.

I may well be missing something — happy to be corrected with the actual mechanism.

---

crates/rue-compiler/src/scaling_harness.rs
```diff
@@ -442,23 +534,171 @@ fn render_diagnostics(errors: &CompileErrors) -> String {
...
+    // having no fresh-side retained key; that is cache reachability, not a
+    // semantic result. Identities retained by both revisions are compared for
+    // exact presence/absence and payload equality. The union is still retained
+    // in the snapshots so deletion tests can explicitly prove that a removed
+    // identity is not observable, rather than silently dropping it here.
+    let identities = warm_bodies
+        .keys()
+        .filter(|identity| fresh_bodies.contains_key(*identity))
+        .cloned()
+        .collect::<BTreeSet<_>>();
```
steveklabnik: Comparing only the intersection of warm and fresh identities is well justified for the failure case — the comment above explains it precisely, and a failed rooted query legitimately stops before an unchanged helper is requested.

That excuse doesn't exist in the `(Ok, Ok)` arm, though. When both sides compiled successfully, a warm-only or fresh-only retained identity is a real divergence, and the intersection filter silently drops it. `assert_successful_output_body_presence` narrows the gap by requiring every emitted function to have exactly one observable transaction, but it only walks `output.functions()` — a retained identity that isn't an emitted function (a stale key one side kept and the other never created) still escapes.

Consider asserting exact key-set equality when both results are `Ok`, keeping the intersection filter only for the `(Err, Err)` arm where the early-stop rationale actually applies.

---

steveklabnik (commented):
## Round 2 review — `2964791` (was `08120a9`)

Reviewed the five new commits as an incremental diff against the previous head. This is a strong revision: every blocking item from the previous round is addressed, and several fixes went further than what was asked.

### Previously blocking — all resolved

**O(n²) probe on the production body loop.** Fixed correctly. `QueryFamily::contains_retained_key` (`rue-query/src/lib.rs:1880`) does `lock(&self.inner.nodes).contains_key(key)` against the existing `HashMap`, and `has_retained_body_key` now calls it instead of enumerating the family per reached body.

I checked that this is semantics-preserving rather than just faster. `body_transactions` is `QueryFamily<BodyQueryKey, BodyTransaction>` — no `CompatibilityKey` wrapper on that family — and `CompatibilityKey`'s `PartialEq`/`Hash` delegate to `key` alone anyway, so the hash probe and the old `candidate == key` predicate select the same node. No epoch or revision discriminator sneaks into the lookup.

`exact_retained_key_probe_uses_hash_lookup_not_predicate_enumeration` is a nice touch: counting `Hash` invocations via `CountingKey` pins the complexity itself rather than just the return value, so a regression to enumeration fails the test instead of silently passing.

**Cold RSS gate was inert.** Fixed — `rss["status"] == "fail"` now propagates into the pair status, with `unsupported`/`indeterminate` handled distinctly.

**macOS RSS parsing never matched.** Fixed — `parse_rss` tries both the `label: value` and `value  label` forms, and the test covers both.

**`manifest.toml` thresholds were decorative.** Fixed, and then some. Schema bumped to 2; `load_audit_manifest` now validates every threshold against `PRECOMMITTED_THRESHOLDS` and rejects drift, and the manifest gained `scenario_policy` and per-scenario `locality` bound tables that the runner actually reads. `pair_verdict` refuses to run without manifest-loaded thresholds. The duplicate source of truth is gone.

**`body_analyses_reused` never asserted non-zero.** Fixed by `counter_lifecycle_covers_cold_unrelated_edit_changed_body_and_deletion`, which pins the full lifecycle: cold (`reused == 0`), unrelated edit (`reused == cold_computed`, and `per_body_declaration_context == default()`), changed body (`computed + reused == cold_computed`). The partition assertion is stronger than a bare `> 0` — it proves the counters account for every cold body.

**Weakened equality assertion.** Fixed — the `body_analyses_invalidated = 0` normalization is replaced by explicit per-field assertions including `invalidated == fresh + 1` with a stated reason. Reads much better than the normalize-then-compare it replaced.

**`test-value-audit.py` not in CI.** Fixed — `value-audit-tool-tests` in `BUCK` with all imported scripts and manifests declared as resources.

### Also fixed from the non-blocking list

Cold-path warmup now runs and is driven by the manifest value rather than hardcoded; cross-protocol pairs fail closed (and single-role protocol mixing marks `indeterminate`); Caldera timeout gets manifest-declared headroom over the 300s budget; `merge_role_sample` no longer erases collected samples on a late `unsupported`; `locality_check` fully type-validates before comparing, so no more `TypeError` on a missing key; `load_perf_baseline` is `lru_cache`d; `source_provenance` uses `tracked_files` instead of hashing untracked build output; `measurement_env` centralizes `RUST_LOG` stripping; `scaling_sample` lost its unused parameter.

Two upgrades beyond the ask are worth calling out. `parity_check` replaced the `parity is not None` presence test with real validation — schema version, `comparison_completed`, per-field exactness against `PARITY_EXACT_FIELDS`, and a 64-hex digest check. And `locality_check` moved from lower-bound presence checks ("some CFG was reused") to manifest-declared upper bounds on computed work, which is the direction that can actually detect a full-program recompute.

### New findings

Three, all inline. Only the first is something I'd want addressed before merge:

1. **`work_projection` aliases counters that don't exist** (`session-bench/src/main.rs`) — `modules`, `rir`, and `cfgs` report `invalidated` as a verbatim duplicate of `computed`, making the invalidated half of the new upper-bound gate arithmetically redundant. `durable_source` borrows `invalidated` and `reused` from the body counters outright. The `exact_identity_sets: {available: false, reason: ...}` pattern already in this function is the right fix.
2. **Deletion-path asymmetry** (`scaling_harness.rs`) — a question, not a defect claim. The test pins that removing an unrelated module recomputes `main` with no retained predecessor, while adding one reuses it. The comment's stated mechanism doesn't match what I can see of the key structure; details inline.
3. **Intersection-only identity comparison** (`scaling_harness.rs`) — justified for the `(Err, Err)` arm, but the same filter applies to `(Ok, Ok)`, where a one-sided retained identity is a genuine divergence.

### CI

24 checks green including `clippy`, `fmt`, `test (macos)`, `compile`, `scaling matrix`, and `compiler reproducibility`; 5 still running; no failures.

Item 1 is the one I'd fix before this leaves draft — it's a schema that claims more evidence than the counters behind it can support, in a report intended to be published as audit evidence. Items 2 and 3 are fine to resolve with a reply if the reasoning already accounts for them.

---

crates/rue-compiler/src/scaling_harness.rs
```diff
@@ -1562,6 +2165,289 @@ fn invalidation_negative_lookup_becoming_positive_invalidates_consumers() {
...
+#[should_panic(expected = "successful warm/fresh body-key sets differ")]
+fn correctness_oracle_rejects_successful_unreachable_warm_key() {
+    let _ = run_source_edit(
+        "reached body becomes unreachable",
+        "fn hidden() -> i32 { 1 }\nfn main() -> i32 { hidden() }\n",
+        "fn hidden() -> i32 { 1 }\nfn main() -> i32 { 0 }\n",
+        &["hidden", "main"],
+    );
+}
+
```
steveklabnik, line 2552: **My round-2 comment was over-broad and this is the fallout — please don't take it as-is.**

I asked for exact key-set equality on `(Ok, Ok)` on the reasoning that a one-sided retained identity is a real divergence. That's wrong for the retained-key *universe*, and the previous revision already said so in a comment I should have weighted more heavily:

```rust
// The source declaration remains valid, so a warm session may retain its
// unchanged transaction even though a fresh rooted compile never reaches
// it. Keep that retained-key case visible; deletion below requires the
// stronger non-observability assertion.
```

That's correct. A warm session retaining `helper`/`unused`/`hidden` after a rename, deletion, or reachability loss is ordinary cache reachability, not a semantic divergence — the retained transaction is still *right*, just no longer *needed*. Tightening the comparison to the whole retained family made three legitimate edit shapes fail the oracle.

Two things follow, and I think both are regressions worth reverting:

**Real coverage was deleted.** `correctness_oracle_rename_delete_and_visibility_edits` asserted concrete behavior — that `main` refreshes after a rename, that `helper` becomes non-observable, that `unused` is non-observable in *both* warm and fresh after deletion, and the precise three-way `before`/`warm`/`fresh` state for the unreachable case. All three replacements assert only that a panic string appears. Nothing now checks what actually happens to a renamed, deleted, or unreachable body.

**The shared oracle got narrower, not stronger.** `assert_warm_fresh_parity` is documented as "the one shared warm-vs-fresh oracle every edit row uses," and `run_source_edit`'s comment says the oracle is "deliberately shared by all of these cases so a new edit shape cannot accidentally fall back to comparing only a root semantic projection." It can no longer be applied to any edit that removes or renames a declaration. A future edit-shape test covering deletion has to route around the oracle entirely — exactly the failure mode that comment warns about.

The scope I should have asked for is the *observable* set, not the retained set. Some candidates, roughly in order of how much I'd trust them:

1. Compare key sets restricted to entries whose value is `Some`. Fixes rename and deletion cleanly (warm's stale `helper`/`unused` are `None` — the old assertions prove it). Does **not** fix the unreachable case, where warm legitimately holds an observable `hidden` that fresh never reaches.
2. Compare the identities reachable from the semantic output on each side, plus deterministic failures — i.e. extend `assert_successful_output_body_presence` into a two-sided set comparison over `output.functions()` rather than over the raw family. This is the one that matches "same semantic result" and handles all three shapes.
3. Keep the intersection for payload equality, and add a one-sided check only for identities that are observable *and* reachable.

Whichever you pick, please restore the three deleted tests' behavioral assertions. `correctness_oracle_rejects_asymmetric_successful_body_key_sets` (the synthetic two-map test) is a fine way to keep the new assertion covered without spending three real edit fixtures on it.

Sorry for the churn — the underlying observation (a warm-only *observable, reachable* identity is a divergence) still holds, but I stated it as the whole retained family and that isn't what it should have said.

---

crates/rue-compiler/src/scaling_harness.rs
```diff
@@ -1562,6 +2165,289 @@ fn invalidation_negative_lookup_becoming_positive_invalidates_consumers() {
...
+        ))
+        .into_result()
+        .and_then(|_| unrelated_session.canonical_semantic(&options))
+        .expect("deletion lifecycle fixture compiles");
+    let deleted_body = deleted_output.work().body_analysis;
+    assert_eq!(deleted_body.body_analyses_computed, 1);
+    assert_eq!(deleted_body.body_analyses_reused, 0);
+    assert_eq!(
+        deleted_body.body_analyses_invalidated, 1,
+        "the forced new request must invalidate the retained main body"
```
steveklabnik: Thanks for chasing this down — the actual mechanism was neither of the two I guessed, and it's more interesting than both. Returning to `unrelated_module_snapshot(0)` produced a snapshot byte-identical to the original cold one, so the top-level semantic query hit its own memo and handed back the *cold request's* work record. The `computed == 1, reused == 0, invalidated == 0` I was puzzling over were literally the cold run's numbers replayed. Good catch.

Two follow-ups.

**The fixture now conflates two edits.** Forcing a new request by changing `main` to `0 + 0` means this case is simultaneously a module deletion *and* a root-body edit, while the test name and position still present it as the deletion case. `invalidated == 1` is attributable to the body edit alone; the module removal contributes nothing observable. If the intent is to cover deletion, consider perturbing an *unrelated* module's content instead of `main`'s body so the deletion remains the only thing touching `main`'s lifecycle — or rename the case to reflect that it's a combined edit.

**The general hazard deserves a note at the source.** I see the runner-side guard added for this:

```python
# A non-zero body/context counter with no current semantic execution is a
# historical cached-result field, not current-request recomputation. Do not
# let an exact-cycle cache hit satisfy a locality claim.
```

That's a good generalization and it covers the audit path. But it's a downstream sniff test for a property of `CanonicalSemanticOutput::work()` itself: on an exact-cycle snapshot, `work()` describes a previous request. Anything reading `output.work()` inherits that, and the only current defense lives in a Python heuristic two layers away. A sentence on `work()` (or on `canonical_semantic`) saying the returned work record belongs to the request that *computed* the memoized result, not necessarily the caller's request, would stop the next consumer from rediscovering this the hard way. The counter-lifecycle test found it only because the numbers happened to look wrong.

---

steveklabnik (commented):
## Round 3 review — `c8794d3` (was `2964791`)

One commit, reviewed as an incremental diff. Two of my three round-2 items are cleanly resolved, the third was answered with a root cause better than either hypothesis I offered — but the fix for the remaining item overshot, and that's on me.

### Resolved

**Aliased locality counters.** Fixed thoroughly. Every `invalidated` field that was a copy of `computed` is now `{available: false, reason: ...}` with a specific reason, `durable_source.reused` is likewise marked unavailable with an accurate explanation, and `required` is uniform across all artifacts. Two things went beyond the report:

- `semantic_bodies.reused` was switched from `durable_bodies.reused_bodies` to `body_analyses_reused`. I'd only flagged that it duplicated `durable_source.reused`; using the counter that actually means body-analysis reuse is the right call.
- Eleven per-body context counters are now projected, with manifest bounds for each and `COLD_SEMANTIC_WORK_REQUIRED_PATHS` pinning their presence.

The Python side keeps pace: `counter_value`/`unavailable_counter` distinguish a real counter from an explicit absence, and the gate skips unavailable dimensions rather than treating them as zero. That's the shape I was asking for.

**Exact-cycle stale work.** The answer — a byte-identical snapshot hits the top-level semantic memo and returns the *previous* request's work record — is the real mechanism, and neither of my guesses was close. Generalizing it into a runner guard (`semantic_executions == 0` with non-zero computed work ⇒ unsupported) is a good move; it protects the audit path, not just the one test. Two follow-ups inline, both minor.

### Needs another pass

**The `(Ok, Ok)` key-set tightening broke three edit shapes, and my round-2 comment is why.** I asked for exact key-set equality over the retained identity universe. That was the wrong scope. A warm session retaining `helper`/`unused`/`hidden` after a rename, deletion, or reachability loss is ordinary cache reachability — the previous revision's own comment said exactly this and I under-weighted it.

The consequence is that `correctness_oracle_rename_delete_and_visibility_edits` became three `#[should_panic]` tests. Concretely:

- Assertions that `main` refreshes after a rename, that `helper` stops being observable, that `unused` is non-observable in both warm and fresh, and the three-way `before`/`warm`/`fresh` state for the unreachable case — all deleted. The replacements assert only that a panic string appears.
- `assert_warm_fresh_parity`, documented as "the one shared warm-vs-fresh oracle every edit row uses," can no longer be applied to any edit that renames or removes a declaration. `run_source_edit`'s comment warns that a new edit shape must not "accidentally fall back to comparing only a root semantic projection" — a future deletion test now has no choice but to route around the oracle.

The observation still holds in narrower form: a warm-only *observable and reachable* identity is a genuine divergence. Options for scoping it that way are inline; the one I'd favor is comparing identities reachable from `output.functions()` on each side plus deterministic failures, which handles all three shapes. `correctness_oracle_rejects_asymmetric_successful_body_key_sets` already covers the new assertion synthetically, so the three real fixtures can go back to asserting behavior.

**Minor:** the lifecycle "deletion" case now changes `main`'s body *and* removes a module in the same step, so `invalidated == 1` is attributable to the body edit alone and the deletion contributes nothing observable. Details inline.

### CI

Running on the new head: 12 green so far including `fmt`, `compile`, `asan`, `scaling matrix`, `compiler reproducibility`, and `release`; the rest in progress; no failures.

Nothing here touches the audit runner or the production hot path — it's confined to test scope in `scaling_harness.rs`. Restoring the three tests' assertions under a narrower comparison should be the last thing between this and leaving draft.

---

steveklabnik (commented):
## Round 4 — `9d67f6b` "refine warm parity oracle"

All three round-3 items resolved. Not merge-blocking; posting for the record since this is queued.

**Coverage regression — fixed via the option I'd have picked.** `reachable_successful_body_identities` walks from `output.functions()` transitively through each transaction's callable reference edges, and the `(Ok, Ok)` arm compares those reachable sets instead of the raw retained family. Stale unreachable keys no longer trip the oracle, and `correctness_oracle_rename_delete_and_visibility_edits` is back as a single test with every original assertion restored — `helper` non-observable after rename while `renamed` is, `unused` non-observable on both sides after deletion, and the three-way `before`/`warm`/`fresh` split for the unreachable case where warm legitimately keeps an observable `hidden`.

The synthetic assertion test survived and got better: it now builds its fixture from a real retained transaction and a reachable-plus-warm-only pair rather than two bare `None` maps. `correctness_oracle_rejects_missing_reachable_transaction_on_both_sides` is a good addition — it covers the new panic path in the reachability walk, which would otherwise have been the one untested branch in the new helper.

**Deletion fixture — fixed more thoroughly than I asked.** Renaming to `..._and_unrelated_deletion` and perturbing the *unrelated* module's content (trailing `\n`) while keeping `main` byte-identical is the right disentangling. The assertions flipped to `computed == 0, reused == 1, invalidated == 0` plus `per_body_declaration_context == default()`, which now proves the actual locality property — deleting an unreachable module doesn't enter the body coordinator at all — rather than recording a number. Stronger than the version it replaces.

**`work()` doc — fixed.** The note that the record belongs to the request that computed the memoized output is exactly the guardrail the next consumer needs.

### Two observations, neither blocking

**The `(Err, Err)` failure-set comparison was loosened in the same commit.** It went from exact key-set equality to intersection-only iteration. That reverses a deliberate earlier decision whose comment argued the opposite case:

> These are current-revision failure terminals, not intentionally retained last-good artifacts... a fresh-side failure that disappears from the warm cache is a correctness failure, not a cache hit.

The new early-stop rationale is reasonable, and the ordered-diagnostics equality asserted just above is a strong backstop for anything user-visible, so I don't think this is wrong. But it's the one place in this commit that moves toward weaker checking, and it silently overturns a documented choice — worth a sentence in the comment saying why the earlier reasoning no longer applies, so the next reader doesn't re-tighten it.

**Minor:** `reachable_successful_body_identities` is called with hardcoded `"warm"`/`"fresh"` rather than the scenario `label`, so its panic message loses which edit fixture failed — every other assertion in `assert_warm_fresh_parity` carries `{label}`. Also `warm_reachable_states`/`fresh_reachable_states` use `.cloned().unwrap_or(None)`, but the walk has already panicked on any reachable identity without a transaction, so that fallback is unreachable.

### CI

31 of 33 green on `9d67f6b` including `clippy`, `fmt`, `asan`, `valgrind`, `test (macos)`, `test (linux-arm64)`, `scaling matrix`, and `compiler reproducibility`; `compile` still running; 1 skipped; no failures.

---

PR #1961, opened 2026-07-25
Gate the regressed large examples, and isolate what decides warm body locality

steveklabnik:
## Scope

Two independent changes, both follow-ups to the value-audit work. Rebased onto
trunk now that #1960 has merged, so this is just the two commits.

Refs RUE-1091, RUE-1083.

## 1. Gate the regressed large examples

`rill`, `mosaic`, `harbor`, `lattice`, and `meridian` are `--skip`ped from
`//:cli-tests`, so their compile cost is not part of the required corpus.

The large-example compile guard (now un-stubbed on trunk) answers a different
question: it proves the six programs still *compile*, under a 3600s timeout,
with a Caldera timeout downgraded to a warning because pure slowness is a
tracked performance follow-up. It makes no performance claim. These rows are
where cold *cost* gets gated.

- Add all five as cold-only audit workloads with absolute per-role wall budgets.
- Check in the pre-cutover and post-cutover figures from
  `rue-1083-closure-evidence.md` as a `[historical_reference]` table, so the
  numbers the project has to beat live in the manifest rather than in prose.
- Generalize the Caldera-only cold budget into per-workload
  `cold_wall_seconds` / `cold_timeout_seconds`. Caldera restates the existing
  threshold and headroom, and the loader rejects drift between them.

The absolute budget is checked **per role, without pairing**, so these rows
return a real verdict in a same-binary protocol smoke where every pair is
`unsupported`. That matters: it means the cold question gets an answer before a
pre-cutover baseline binary exists. Each workload's process timeout sits above
its own budget so an over-budget compile is recorded as a gate failure rather
than raising out of the run.

The loader also rejects a reference that claims to be role evidence, a gate
detached from its declared multiple of the reference, and a gate that fails to
separate the reference from the recorded regression.

Cost: these run under the same 1-warmup/7-paired-sample protocol as every other
row, and `meridian` currently compiles in minutes. The full matrix is an
hours-long run, in the same class as the existing Caldera row. Noted in the
README; `--workloads` selects a subset.

## 2. Isolate what decides warm body locality

Two checked-in workloads reported incompatible results. The N=128 completion
scenarios reanalyze exactly the edited leaf. The seven-file representative
fixture reanalyzes **all six** of its bodies after a leaf edit — including three
that are provably outside the edit's reverse closure (`worker.compute`,
`model.weight`, `std.math.abs` are unaffected by editing `labels.adjustment`).

Those fixtures differ in body count, module count, std use, and discovery
protocol simultaneously, so neither result could be attributed to a cause. The
natural hypothesis was that the representative program is too small for locality
to be visible. That hypothesis is wrong.

`rue-compiler-session-bench --module-axis` holds the call graph, reached body
count, and edit shape fixed and varies two things independently: module count,
and the protocol that commits import discovery.

| modules | discovery | leaf edit: computed | reused | durable source records copied |
|---:|---|---:|---:|---:|
| 1 | staged | 1 | 127 | 129 |
| 1 | rooted_demand | 128 | 0 | 16512 |
| 8 | rooted_demand | 128 | 0 | 17408 |
| 128 | rooted_demand | 128 | 0 | 32768 |

(128 reached bodies. The cold row copies 16512 records at one module.)

- **Module count changes nothing.** One module and 128 are identical.
- **The discovery protocol changes everything.** Same program, same module
  count, same edit: one body reanalyzed under staged discovery, all 128 under
  rooted demand. A one-line leaf edit under rooted demand performs *exactly* the
  declaration-context work of its own cold build.
- Durable CFG reuse is unaffected in every row — a leaf edit rebuilds 1 CFG and
  reuses 127 under both protocols. Only body-query reuse collapses.

Staged discovery is not available to any program with an import: the observation
ledger is incomplete and the attempted revision refuses to close. The axis
asserts this by construction, which is why staged only runs at one module. So
the exact-reverse-invalidation evidence rests on a shortcut only single-file
programs can take, and every multi-module program takes the other path.

This does not contradict any assertion in the completion workload — those are
accurate about the corpus they measure. It qualifies what that corpus is
evidence *for*, and the qualification is now recorded next to the claim in
`body-analysis-cfg-incrementality-audit.md`.

The workload pins the staged locality result, since that is what the existing
claim rests on, and deliberately leaves the rooted-demand numbers unasserted so
a repair does not read as a failure.

### Why this is worth knowing before the flip

The per-body context repair targets the O(bodies × declarations) cost *inside*
`analyze_body_query` — what a body pays once chosen. The behavior measured here
decides *how many* bodies get chosen. Those are different layers. If the flip
lands and multi-module warm rebuilds still recompute every body, the cause is
here rather than in the repaired path.

## Test evidence

Run before the rebase onto post-#1960 trunk; CI covers the current base.

- `scripts/rue fmt`: clean. Unrelated formatter churn in
  `crates/rue-test-runner/src/lib.rs` was reverted, not committed.
- `bash ./clippy.sh`: 66 targets, no violations.
- `scripts/rue unit compiler-session-bench`: 10 passed (6 existing + 4 new).
- `python3 scripts/test-value-audit.py` and `./buck2 test //:value-audit-tool-tests`:
  37 passed (28 existing + 9 new).
- `scripts/rue quick`: 36 targets passed.
- Module axis run at `--bodies 16 --module-counts 1,2,4,8` and
  `--bodies 128 --module-counts 1,8,128`; both tables above are from those runs.

The value-audit runner itself was not executed end-to-end — it needs three
release binaries with distinct source provenance, which this environment does
not have. The new workloads are covered by the loader validation and unit tests
only.

### DIFF
--- a/benchmarks/value-audit/README.md
+++ b/benchmarks/value-audit/README.md
@@ -34,6 +34,36 @@ treated as zero work. A run using one binary for all three roles is classified
 as `same_binary_protocol_smoke`; it validates protocol and fail-closed gates,
 not historical/current/candidate value.
 
+## Regressed-example cold workloads
+
+`rill`, `mosaic`, `harbor`, `lattice`, and `meridian` are the five large example
+programs the RUE-1026/RUE-1027 query cutover regressed. All five are `--skip`ped
+from `//:cli-tests` and the large-example compile guard is stubbed, so this audit
+is currently the only place their cold cost is gated.
+
+Each carries an absolute `cold_wall_seconds` gate derived from the checked-in
+`[historical_reference]` table, which transcribes the pre-cutover and
+post-cutover figures from
+[`docs/notes/rue-1083-closure-evidence.md`](../../docs/notes/rue-1083-closure-evidence.md).
+That reference is prose evidence from a local release-build comparison on the
+maintainer's host — it is not a run of this protocol, it is not attributable to
+a role binary, and the runner refuses to let it be declared role evidence.
+
+Two properties make these rows useful before a historical baseline binary
+exists:
+
+- The gate is **per-role and absolute**, so it produces a real verdict even in a
+  `same_binary_protocol_smoke` where every pair is `unsupported`. Role-vs-role
+  cold comparison still needs the pre-cutover binary; the gate does not.
+- `cold_timeout_seconds` is deliberately far above the gate, so an over-budget
+  compile is recorded as a gate failure instead of raising out of the run.
+
+Cost: these are cold-driver workloads under the same one-warmup, seven-paired-
+sample protocol as every other row, and `meridian` currently compiles in minutes
+rather than seconds. Expect the full matrix to run for hours on one host, in the
+same class as the existing Caldera row. Use `--workloads` to select a subset
+while iterating.
+
 The existing session benchmark emits versioned full-parity evidence and
 production computed/reused/invalidated counters. Required warm locality gates
 use manifest-authoritative exact or upper bounds; full-program reparsing,
--- a/benchmarks/value-audit/README.md
+++ b/benchmarks/value-audit/README.md
@@ -38,8 +38,15 @@ not historical/current/candidate value.
 
 `rill`, `mosaic`, `harbor`, `lattice`, and `meridian` are the five large example
 programs the RUE-1026/RUE-1027 query cutover regressed. All five are `--skip`ped
-from `//:cli-tests` and the large-example compile guard is stubbed, so this audit
-is currently the only place their cold cost is gated.
+from `//:cli-tests`, so their compile cost is not part of the required corpus.
+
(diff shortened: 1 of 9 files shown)


steveklabnik:
Closing: already merged, via #1963.

This PR was the bottom of a three-PR stack, so #1963's branch contained every commit in it. When #1963 merged, all nine commits landed together. The `MERGE_CONFLICT` that removed this from the queue is the queue finding nothing left to apply, not a real conflict.

All four commits from this PR are on trunk:

- `a925b25d` Gate the regressed large examples in the incrementality value audit
- `80921720` Isolate what decides warm body locality with a module/discovery axis
- `983f34fb` Correct the large-example guard description in the audit README
- `ed1417e7` Record the mechanism behind the rooted-demand locality loss

No action needed and nothing to re-queue.

---

PR #1962, opened 2026-07-25
Value-audit follow-ups: scale the leaf bound, bound the gauntlet's claim, attribute the skips

steveklabnik:
## Scope

Four independent follow-ups from the assessment. **Stacked on #1961** — its four
commits appear here until it merges. Review the last four commits only.

Refs RUE-1091, RUE-1121, RUE-1083.

## 1. Scale the warm leaf-body locality bound by its reverse closure

`max_*_computed` was one unit for every counter, which would **fail a correct
implementation**: exact reverse invalidation recomputes the edited body *and its
direct reverse callers*, so a leaf edit does not cost one unit. The bound
survived only because production currently recomputes the whole program, which
fails at any bound.

The bounds are now stated per recomputed unit, and each workload declares the
reverse closure of the leaf it edits — 2 for the synthetic corpus (`f000`,
`main`), 3 for the representative fixture (`labels.adjustment`, `report.adjust`,
`main`). The runner multiplies.

Safety properties, all tested: a zero per-unit bound stays zero at any scale, so
the no-op and unrelated-edit gates cannot be loosened; a non-positive or
non-integer scale is `unsupported`, never a pass; and the loader rejects a
workload that declares a closure without supporting the scenario or vice versa.

## 2. Record what the post-flip gauntlet cannot observe

Following the mechanism found in #1961: retained work is validated by a
compatibility token that import discovery mints fresh per request, so anything
going through discovery cannot validate what it retained.

I checked every locality-asserting row in the gauntlet. All of them run on the
constant-token path:

- the RUE-1121 exact-recompute-set and exact-flat rows (stage 3) drive
  `session.update` + `canonical_semantic` with no discovery;
- stage 7/8 use `rue-scaling-bench`, same;
- stage 9 is a cold compile.

The one harness row that *does* drive rooted demand
(`correctness_oracle_import_edit_compares_imported_body_and_linked_bytes`)
asserts parity and consumer refresh — never locality. That is why it passes
today, and it is consistent with the measurements in #1961.

Two consequences, now written next to the stage ledger:

1. **No stage is gated on the epoch-reset decision.** The flip need not wait, and
   a failing stage must be explained by the flip rather than by this.
2. **A green sweep is not evidence that a driver-shaped warm rebuild is fast.**
   It measures an in-process host managing its own snapshots. Extending stage 8
   to `rue main.rue` on a developer's machine is unsupported by the gauntlet as
   written; claiming it needs one added row driving a warm edit through the
   rooted-demand protocol.

## 3. Attribute the large-example CLI skips

The `--skip` list read as an unowned permanent exclusion. It now says what it is:
a performance deferral owned by RUE-1083, with compile *correctness* covered by
the un-stubbed large-example guard and compile *cost* gated by the
`regressed_example` audit workloads. Restoring a name needs the compile time
back, not more coverage.

## 4. Ignore Python bytecode

`scripts/__pycache__` is easy to sweep into a commit by accident — I did exactly
that in #1961 and force-pushed the removal. Ignoring it prevents a repeat.

## Test evidence

- `python3 scripts/test-value-audit.py` and `./buck2 test //:value-audit-tool-tests`:
  42 passed (37 before, +5 for the scaled bound).
- `bash ./clippy.sh`: 66 targets, no violations.
- No Rust changed in this PR; the compiler suites are unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_012LsHBgTP9eKtFWf2hPgcQ8)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ third-party/.cargo/.global-cache/
 
 # BuildBuddy remote cache local config (RUE-316) — holds the API key, never committed
 .buckconfig.local
+
+# Python bytecode from repository tooling scripts.
+__pycache__/
+*.pyc
--- a/BUCK
+++ b/BUCK
@@ -203,11 +203,24 @@ sh_test(
     },
 )
 
-# RUE-1083: these four large example programs exceed their compiler budgets
-# after the query cutover. Keep the CLI targets running the other 1,667 cases
-# while RUE-1083 restores their automatic and explicit example coverage;
-# Caldera remains isolated below. Shared verbatim by //:cli-tests and its shards
-# so a slice runs exactly the same cases the monolithic target would.
+# RUE-1083: these large example programs exceed their compiler budgets after the
+# query cutover. Keep the CLI targets running the other 1,667 cases while
+# RUE-1083 restores their automatic and explicit example coverage; Caldera
+# remains isolated below. Shared verbatim by //:cli-tests and its shards so a
+# slice runs exactly the same cases the monolithic target would.
+#
+# These skips are a PERFORMANCE deferral, not a coverage decision, and they are
+# owned by RUE-1083 rather than left unattributed:
+#
+# * that these programs still COMPILE is covered by the large-example guard
+#   (.github/workflows/large-examples.yml), which is no longer stubbed;
+# * that their compile COST is bounded is gated by the regressed_example
+#   workloads in benchmarks/value-audit/manifest.toml, whose budgets come from
+#   the pre-cutover figures in docs/notes/rue-1083-closure-evidence.md.
+#
+# Restoring them here requires the compile time, not more coverage. Remove a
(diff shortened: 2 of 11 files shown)


steveklabnik:
Closing: already merged, via #1963.

This PR sat in the middle of a three-PR stack, so #1963's branch contained every commit in it. When #1963 merged, all nine commits landed together. The `MERGE_CONFLICT` that removed this from the queue is the queue finding nothing left to apply, not a real conflict.

All four of this PR's own commits are on trunk:

- `1b34e1c3` Scale the warm leaf-body locality bound by its reverse closure
- `31939d7a` Record what the post-flip gauntlet cannot observe
- `53cd470c` Attribute the large-example CLI skips to their owning issue
- `28710b0d` Ignore Python bytecode from the repository tooling scripts

No action needed and nothing to re-queue.

---

PR #1966, opened 2026-07-25
RUE-1132: share the declaration projection across a request's bodies

DorianListens:
## Summary

`analyze_body_query` projected the whole durable declaration universe once per reached body. Every input to that projection is revision-scoped — the merged program, the prepared shell records and their provisional definition universe, and the query-owned durable declarations — and none carries a body key, so each body recomputed a bit-for-bit identical value.

This projects once per rooted attempt and shares the resulting `Arc<[SemanticDeclarationExport]>` with every body of the request.

Refs RUE-1132, RUE-1091, RUE-1083.

## What the memo is, and is not

It is not a dependency authority. It holds no lease, observes no terminal, is created inside one rooted attempt and dropped with it, and pins the revision and configuration it was filled for. A request presenting a different revision or configuration fails closed rather than reprojecting into a slot owned by another revision, so no body can observe a projection that is not its own. A failed projection leaves the memo empty, so the deterministic failure reaches every body rather than one body failing and the rest reusing a half-filled slot. Invalidation semantics are unchanged.

The anonymous-nominal projection is body-varying and deliberately stays per body. The projection lock is never held across the projection itself, mirroring `recipe_cache`'s claim discipline.

Projection work counters are charged only when a body actually projects, at the stage's own source, so a reusing body drops the counter rather than having it subtracted afterwards.

## Structural effect

Exact, deterministic, reproduced by the scaling harness on every run:

| bodies × decls | `per_body_project` | `projections_total` |
|---|---:|---:|
| 100 × 100 | 201 → **1** | 20,301 → **201** |
| 1000 × 100 | 1101 → **1** | 1,102,101 → **1,101** |
| 100 × 1000 | 1101 → **1** | 111,201 → **1,101** |

## Cold wall time (value-audit protocol)

`scripts/rue-value-audit.py` under its standard protocol — 1 warmup, 7 paired alternating samples, medians and MADs — with pre-change trunk `fb54de6` as `current_production` and this revision as `candidate`. Both binaries built at `//platforms:release` on one host (4-core Xeon 2.1GHz), distinct SHA-256 verified before measuring.

| workload | base ms (MAD) | candidate ms (MAD) | delta | 3×MAD bar | significant |
|---|---:|---:|---:|---:|:--:|
| synthetic (1000 bodies) | 9465.8 (79.5) | 6960.3 (178.8) | **−26.5%** | 536.4 | yes |
| representative_multi_module | 281.3 (5.9) | 239.0 (2.6) | **−15.1%** | 17.8 | yes |
| medium_ordinary_rue | 21.0 (0.8) | 19.5 (0.6) | −7.1% | 2.5 | no |

`medium_ordinary_rue` (quicksort, 73 lines) does not clear the bar and is reported as **no measurable change**. That is the expected shape rather than a disappointment: the saving scales with reached bodies, and quicksort has too few to share a projection across.

Cold peak RSS is unchanged on every workload, far inside the audit's budget (largest delta 24 KiB against a 1.5 MiB allowance).

**Scope of the claim.** The `current_production_vs_candidate` cold pair **passes** on all three workloads. The audit's *overall* verdict is `indeterminate`, for two structural reasons that are not about this change: its historical-baseline pair is a deliberate same-binary smoke (no pre-cutover binary was supplied — that is #1963's concern), and the warm scenarios are `unsupported` without per-role session-benchmark binaries. **Warm behavior is not claimed here.** `caldera` was not run; it is cold-only at ~45 min per sample, so seven paired samples is an hours-long run.

A second, independent protocol run gave synthetic −22.8% and representative −4.6%. Both runs agree on direction everywhere. Synthetic is robust across both (−23% to −27%); the representative workload sits closer to its noise floor and cleared the bar in one run of two. The table above is the run with complete wall+RSS evidence.

## What this does not do

Compile remains O(declarations × bodies): prepare and install are still paid per body, and together they are the larger share. This removes the ×bodies factor from one of the four stages. It does not move the RUE-1090 gate, which correctly still reads `ACTIVATE RUE-1091` — the projection slope drops ~100× without flattening.

## Harness rows

The RUE-1121/RUE-1090 rows record the new charging shape rather than being loosened. The projection row remains a tracked RUE-1091 expected-failure — one whole-universe projection is still not exactly flat in unrelated declarations — but its witness moves from `universe × bodies` to `universe`, so a regression back to per-body projection blows past the band and panics. The growing-bodies row becomes a hard total-work invariant on total projection work, strictly stronger than the per-body linear row it replaces.

## Test evidence

- `./test.sh`: PASSED (exit 0).
- `scripts/rue quick`: 36 targets.
- `//crates/rue-compiler:rue-compiler-test`: 842 passed, including 4 new memo tests — reuse asserted on `Arc` identity, fail-closed on foreign revision and on foreign configuration, and no caching of a failed projection.
- `//crates/rue-compiler:scaling-matrix-test`: 841 passed.
- `bash ./clippy.sh`: 66 targets, no violations. `scripts/rue fmt`: clean (unrelated formatter churn in `crates/rue-test-runner/src/lib.rs` was reverted, not committed).

### DIFF
--- a/crates/rue-compiler/src/body_overlay.rs
+++ b/crates/rue-compiler/src/body_overlay.rs
@@ -693,6 +693,11 @@ mod tests {
             cancellation: &rue_query::CancellationToken,
         ) -> Result<crate::body_query::BodyTransaction, rue_query::QueryAbort> {
             let mut work = rue_air::PerBodyDeclarationContextWork::default();
+            // A fresh memo per call: this differential compares one production
+            // body against one overlay body, so each side projects for itself
+            // exactly as it did before RUE-1132 shared the projection across the
+            // bodies of a request.
+            let shared_projection = crate::canonical_semantic::SharedDeclarationProjection::new();
             crate::canonical_semantic::analyze_body_query(
                 &self.merged,
                 &self.rir,
@@ -704,6 +709,7 @@ mod tests {
                 &crate::body_query::WellKnownOptionResolution::default(),
                 key,
                 cancellation,
+                &shared_projection,
                 &mut work,
             )
         }
--- a/crates/rue-compiler/src/canonical_semantic.rs
+++ b/crates/rue-compiler/src/canonical_semantic.rs
@@ -12,6 +12,118 @@ use rue_air::{
 };
 use tracing::info_span;
 
+/// Request-scoped memo for the body-invariant durable declaration projection
+/// (RUE-1132).
+///
+/// `project_durable_declaration_semantics` is a pure function of inputs that are
+/// all revision-scoped — the merged program, the prepared shell records and
+/// their provisional definition universe, and the query-owned durable
+/// declarations. None of them carries a body key, so every reached body used to
+/// recompute a bit-for-bit identical value. This memo computes it once per
+/// semantic request and hands the shared `Arc` to each body.
+///
+/// It is **not** a dependency authority and holds no lease: it is created inside
+/// one rooted attempt, dropped with it, and pins the revision and configuration
+/// it was filled for. A request presenting a different revision/configuration is
+/// a programming error, not a cache miss — the body fails closed rather than
+/// recomputing into a slot that belongs to another revision. Because the value
+/// is owned data (`Arc<[SemanticDeclarationExport]>`) rather than a semantic
+/// epoch, sharing it retains no mutable declaration state across bodies.
+///
+/// The projection lock is never held across the projection itself: a miss
+/// releases the lock, projects, then re-locks to publish. Two concurrent misses
+/// may therefore both project — deterministically producing equal values — and
+/// the first writer wins, mirroring `recipe_cache`'s claim discipline.
+pub(crate) struct SharedDeclarationProjection {
+    filled: std::sync::Mutex<Option<FilledDeclarationProjection>>,
+}
+
+struct FilledDeclarationProjection {
+    revision: crate::SourceRevision,
+    configuration: crate::semantic_query_nucleus::SemanticQueryConfiguration,
+    exports: Arc<[rue_air::SemanticDeclarationExport]>,
+}
+
+/// Why a body's projection request did not reuse the shared value. The stage
+/// charges its projection work only on a `Computed` outcome, so a reused
+/// projection drops the per-body counter at the stage's own source rather than
+/// being subtracted afterwards.
+enum SharedProjectionOutcome {
+    Computed(
+        Arc<[rue_air::SemanticDeclarationExport]>,
+        crate::durable_semantics::DurableSemanticProjectionWork,
+    ),
+    Reused(Arc<[rue_air::SemanticDeclarationExport]>),
+}
+
+impl SharedDeclarationProjection {
+    pub(crate) fn new() -> Self {
+        Self {
+            filled: std::sync::Mutex::new(None),
+        }
+    }
+
+    /// The declaration projection for this request, computing it on first
+    /// demand. `project` runs outside the lock.
+    fn get_or_project(
+        &self,
+        revision: &crate::SourceRevision,
+        configuration: &crate::semantic_query_nucleus::SemanticQueryConfiguration,
+        project: impl FnOnce() -> Result<
+            (
+                Arc<[rue_air::SemanticDeclarationExport]>,
+                crate::durable_semantics::DurableSemanticProjectionWork,
+            ),
+            crate::durable_semantics::DurableSemanticProjectionFailure,
+        >,
+    ) -> Result<SharedProjectionOutcome, SharedProjectionFailure> {
+        {
+            let filled = self
+                .filled
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            if let Some(filled) = filled.as_ref() {
+                if filled.revision != *revision || filled.configuration != *configuration {
+                    return Err(SharedProjectionFailure::ForeignRequest);
+                }
+                return Ok(SharedProjectionOutcome::Reused(filled.exports.clone()));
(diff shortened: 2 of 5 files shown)


DorianListens:
## `./test.sh`: PASSED

The full suite finished after this PR was opened, exit code 0, terminating in `=== TEST SUITE: PASSED ===`. From the captured tail: `//:spec-tests` 2178 passed / 0 failed / 15 ignored, `//:ui-tests` 209 passed / 0 failed. The `//:cli-tests` stage completed earlier in the run and scrolled out of the captured window, but the suite gate is a single pass/fail and it passed.

## Correction: the wall-time figure in the description is overstated

The description reports mosaic at a 14.87 s → 11.97 s median. I no longer think that number should be relied on, and I would rather say so here than leave it standing.

Two measurements of the same change disagree by about 2×:

| measurement | delta | basis |
|---|---:|---|
| `Body_project_declarations` span | **−1.6 s** | within-run, directly attributed |
| end-to-end wall median | −2.9 s | 3 samples per side, taken ~40 min apart |

The end-to-end delta is nearly double the stage delta, which the change itself cannot explain — it can only remove the projection work. The two sample sets were collected on a shared 4-core VM with several heavy builds in between, so host drift is the likely difference.

**The defensible claim is −1.4 to −1.6 s on mosaic, about 11% of compile**, and that is what should be quoted. The 2.9 s figure is an upper bound pending a properly paired, alternating-sample run under the value-audit protocol.

The structural evidence in the description is unaffected — the counter table is exact, deterministic, and reproduced by the scaling harness on every run.

Nothing else in the description changes. In particular the RUE-1090 verdict remains `ACTIVATE RUE-1091`: this PR reduces the projection slope 100× without flattening it, and does not move the gate.

---

steveklabnik (commented):
Adversarial review: I did not find a code-level correctness defect in this change.

I traced the projection inputs and payload ownership, every `analyze_body_query` call site, the rooted-attempt lifetime, warm-body behavior, failure/retry behavior, and the scaling-counter changes. The shared value contains owned stable identities rather than request-local AIR handles; the anonymous-nominal projection remains per-body; and the memo is created inside the one production rooted attempt. The changed scaling rows still fail closed on a return to per-body projection.

I also independently ran the focused four memo tests and `scripts/rue quick` from the PR head in an isolated worktree. That passed all 36 quick targets, including 842 compiler tests, 845 scaling-matrix tests, the differential oracle, and both corpus oracle audits.

One acceptance/evidence issue remains before this should be represented as completing RUE-1132: the issue requires the cold-wall improvement to be recorded with the usual median/MAD value-audit protocol, but the PR description still claims a 14.87 s → 11.97 s median and says it clears the significance bar. The follow-up comment explicitly retracts that claim as unreliable, notes the samples were separated by ~40 minutes on a shared VM, and says a paired alternating run is still pending. Please either run the required protocol and replace the table/conclusion with defensible results, or remove the unsupported end-to-end claim and explicitly leave that acceptance criterion open. A correction in the conversation does not make the PR description accurate.

Housekeeping: the generated-by footer in the PR description is also inconsistent with this repository tool-neutral contribution-text rule and should be removed.

No inline code findings.

DorianListens:
Thanks — both points addressed. Force-pushed `a0440b8` → `634b23e`; the code is unchanged, the commit message and description are not.

**1. Acceptance evidence.** You were right that a retraction in a comment does not make the description accurate. I ran the required protocol rather than deleting the claim.

`scripts/rue-value-audit.py`, standard protocol (1 warmup, 7 paired alternating samples, medians/MADs), pre-change trunk `fb54de6` as `current_production` against this revision as `candidate`, both at `//platforms:release` on one host:

| workload | base ms (MAD) | candidate ms (MAD) | delta | 3×MAD bar | significant |
|---|---:|---:|---:|---:|:--:|
| synthetic (1000 bodies) | 9465.8 (79.5) | 6960.3 (178.8) | −26.5% | 536.4 | yes |
| representative_multi_module | 281.3 (5.9) | 239.0 (2.6) | −15.1% | 17.8 | yes |
| medium_ordinary_rue | 21.0 (0.8) | 19.5 (0.6) | −7.1% | 2.5 | **no** |

The `current_production_vs_candidate` cold pair passes on all three. The description now states the scope precisely: the audit's *overall* verdict is still `indeterminate` because the historical-baseline pair is a deliberate same-binary smoke and the warm scenarios are `unsupported` without per-role session-bench binaries, so **warm is not claimed**. `caldera` was not run — cold-only at ~45 min per sample.

Three things worth flagging rather than burying:

- **`medium_ordinary_rue` shows no measurable change.** It is in the table as a non-result. Quicksort has too few reached bodies for a shared projection to matter, which is the shape the mechanism predicts.
- **A second independent protocol run gave synthetic −22.8% and representative −4.6%.** Direction agrees everywhere; synthetic is robust across both runs, the representative workload sits near its noise floor and cleared the bar in one run of two. Both runs are disclosed in the description rather than only the better one.
- **The candidate binary I first measured was stale.** It predated my last two commits by six minutes. I caught it on an mtime check, rebuilt from the committed tree, and verified distinct SHA-256 for both roles before sampling. The old numbers are gone from the record.

**2. Housekeeping.** Footer removed. `AGENTS.md:265` also forbids co-author trailers in *commit* text, and the commit carried `Co-Authored-By:` and `Claude-Session:` trailers — you did not flag those, but they are the same rule, so they are gone too. The amended message also replaces the retracted wall-time paragraph, so the durable record no longer carries the unsupported claim.

RUE-1132's acceptance criteria now read: byte-identical corpus output ✓, counters drop to one unit per revision ✓, cold-wall improvement under the median/MAD protocol ✓, invalidation semantics unchanged ✓.


---

PR #1987, opened 2026-07-26
Restore the reasonably-budgeted large-program CI gates

steveklabnik:
Part of RUE-1083.

Restores every temporary CI contingency whose check fits a reasonable budget at the current performance level, keeps the two programs that don't (Caldera, Meridian) deferred with measured evidence recorded inline, and recalibrates the borderline contracts to measured cold CI cost. All required CI is green at `bec3599`.

**Restored (real coverage):**
- **CLI skip filters removed** for the automatic and explicit Harbor, Lattice, and Mosaic families and Rill's temporary contract. Rill returns from the 60s `heavyweight_medium` (retired) to the 30s `heavyweight` contract; `std_collections_m2_smoke` returns to the ordinary case budget; the frontend-diff child-process ceiling returns to 30s; `heavyweight_extra_long.compile_timeout_ms` returns from 900,000 to 300,000.
- **Contracts recalibrated to measured cold CI cost:** Lattice → 120s `heavyweight_long` (killed at 30.014s/30s on linux-x64), Harbor → 120s `heavyweight_long` (cases killed at 30.009s–30.014s/30s on linux-arm64 and linux-x64).
- **Valgrind corpus restored** (minus Caldera): the required step runs `scripts/run-sanitizer.sh` again, now ~6.5 minutes end to end. Restoring it immediately caught a stale curated program the stub had been hiding: `san_fallible_initializer.rue` predates RUE-1112's identity-based `?` and defined its own structurally Option-shaped enum; it now returns the trusted `std.option.Option`, preserving the uninitialized-slot cleanup scenario it memchecks.
- **Guard workflow deleted:** the temporary non-required `large-examples.yml` compile guard is gone.

**Recalibrated outer margin:** with real heavyweight coverage back, linux-arm64 shard 1 hit the test executor's 600s default at exactly 10:00 with zero case failures, so `scripts/ci-heavy-suite` gives the CLI shards a 1200s and the monolithic corpus an 1800s explicit outer margin. The per-case contract budgets remain the honest gates; the margin re-tightens with them.

**Still deferred (RUE-1083 stays open):**
- **Caldera:** killed at its 300s contract on linux-arm64 and measured ~31 minutes end to end (non-release compile + memcheck) in the Valgrind corpus on linux-x64 (run 30211132077). `//:cli-tests-caldera` keeps its transparent success stub and the Valgrind driver defers only the Caldera root.
- **Meridian:** killed at 120.025s against the 120s long contract on linux-x64 and at 300.022s even against the widened 300s extra-long contract on linux-arm64 — no reasonable per-case budget fits. The ordinary corpus skips its automatic and explicit cases next to Caldera's isolation skip; the extra-long contract still bounds filtered local runs, and Meridian remains real in the Valgrind corpus.

Local validation: wrapper-script tests (83/83) and build-sharing tests (30/30) pass; TOML/YAML/bash parse clean; rustfmt clean.


steveklabnik:
Gonna keep the full coverage but let the times go up a bit, since they're not oppressive, even if they're longer than before.

steveklabnik:
Done in 5dbda9c: Lattice moves to the 120s `heavyweight_long` contract and Meridian to the 300s `heavyweight_extra_long` contract, with the shard-2 measurements (killed at 30.014s and 120.025s respectively) recorded in the section comments. All coverage stays real.

Two things this round of CI still has to measure, since the first round's fail-fast cancelled them: Caldera against its restored 300s budget (left unchanged so we get a real number), and the restored Valgrind corpus's total runtime. Will adjust if either comes back over.

---

PR #1998, opened 2026-07-27
Remove two test-only paths the compiler never takes

steveklabnik:
Two `#[cfg(test)]` mechanisms let tests validate behavior production never
executes. Each is removed by making the test take the production path instead.

---

## 1. The durable reuse baseline was chosen twice (RUE-1143)

`semantic_attempt` chose its durable reuse baseline twice. Production read the
last-good semantic record; two `#[cfg(test)]` fields mirrored that record and,
under test, shadowed it:

```rust
let previous_cfg_cache = durable_baseline
    .as_ref()
    .and_then(|entry| entry.successful_cfg_cache.clone())
    .unwrap_or_else(|| Arc::from([]));
#[cfg(test)]
let previous_cfg_cache = self.last_successful_cfg_cache.clone().unwrap_or(previous_cfg_cache);
```

The same shape applied to `durable_declaration_cache`. So the durable-reuse
tests validated a baseline-selection order that production never executes — and
they sit underneath the differential oracle, which is the defense for exactly
that class of bug. The mirrors also had to be silently repopulated after every
successful attempt to stay current, so the test-visible baseline depended on
bookkeeping production did not perform.

**Change.** One unconditional slot, `durable_baseline_override`, and one
selection order compiled identically in both configurations: an explicit
override when supplied, otherwise the last-good record. Ordinary compiles never
set the override, so this reduces to the last-good record.

The slot is deliberately **not** `#[cfg(test)]` — that is what keeps the
algorithm single-path. Only the injection helpers are test-gated, which is an
injection point rather than a divergent algorithm. This follows how the crate
already handles `shared_declaration_base_differential` and the recipe/overlay
meters: production-inert state carrying a comment that says so.

Reuse tests now read the produced baseline through accessors over the last-good
record, perturb a copy, and inject it — exercising the production lookup order
rather than a parallel branch. The CFG-schema test drops the override after the
attempt it was meant to affect, making explicit what the old mirror did
implicitly by repopulating.

## 2. The fabricated test import graph (RUE-1149)

`CompilerSession::supplied_test_import_graph` (+ `adopt_test_import_graph`) let
a test hand the session a pre-built `CanonicalImportGraph`, bypassing discovery
entirely. The graph came from `test_support::test_fixture_import_graph`, which
resolved imports by **string-matching specifiers against loaded module
spellings**. Its own doc comment conceded the point:

> Build explicit canonical records for in-memory fixtures whose import spellings
> name one loaded logical or physical module spelling. This is deliberately a
> test-data convention, **not a path-resolution implementation**.

So it was a second, weaker resolver. Production resolves through the discovery
protocol with candidate precedence, canonical identity, provenance, typed
observations, and read policy (ADR-0051); around 28 session tests exercised none
of it. Same defect class as RUE-1143, one layer up.

**Change.** `test_support::TestDiscoveryHost` is an in-memory host for the real
protocol, modelled on the loop `rue-oracle-diff` already runs:

```text
begin_import_input_request
  -> { stage_import_discovery -> import_demand_frontier_for_roots
       -> serve each frontier request as accepted|absent -> add_plan_reads
       -> publish_import_observation_batch }
  -> close_import_discovery
```

What makes this a swap rather than a reimplementation: **the host resolves
nothing.** The compiler computes each probe and hands it over as a frontier
request carrying a `requested_path`; the fixture answers only whether that exact
path exists and what its bytes are. Precedence, canonical identity, provenance,
and read policy all stay in production code.

A fixture's *logical* module paths describe a virtual project tree — each module
served at exactly the path production would request for that module identity,
under a synthetic project root. Its *physical* paths carry over as canonical
identity, which is their production role: a canonical path that differs from the
requested one is the relocated-or-linked file case, not a second spelling the
resolver may match against.

Both choke points — `test_support::publish_test_snapshot` and `session.rs`'s
`publish_with_test_imports` — route through the host, so the ~28 call sites are
unchanged. Import-free fixtures take `import_graph::import_free_canonical_graph`,
which is now the single construction the session's own semantic path uses for a
revision with no import directives.

Deleted, not merely unused: `supplied_test_import_graph`,
`adopt_test_import_graph`, `test_fixture_import_graph`,
`test_fixture_import_graph_parts`, and `bound_definitions::test_fixture_import_graph`
(which existed only to call `_parts`). No production or test path constructs a
`CanonicalImportGraph` outside discovery or the import-free constructor; the
remaining direct `from_supplied` calls are the validator's own unit tests.

### Fixtures that only resolved under string matching

These are the ones the old matcher accepted and real candidate precedence does
not. Each was corrected rather than routed around, and each names behavior that
was previously untested.

- **`pipeline_tests::canonical_std_strbuf_identity_survives_qualified_and_aliased_lookup`**
  and **`trusted_std_presence_does_not_create_a_bare_strbuf_prelude_name`** —
  imported the std entry point with the absolute specifier `"/sdk/_std.rue"`,
  which pins the fixture to one host's std layout. Now `"std/_std.rue"`, which
  the resolver joins onto the importer and root anchors like any other
  specifier.

- **`pipeline_tests::caller_logical_std_spelling_cannot_spoof_strbuf_language_item`**
  → **`caller_std_shaped_specifier_cannot_spoof_strbuf_language_item`** — the
  old fixture handed `SourceMetadata` a caller-authored logical path of
  `\0rue-std/strbuf.rue` for a file spelled `spoof.rue`, and the string matcher
  resolved `@import("spoof.rue")` to it. Discovery derives the logical identity
  from the requested path, so that spoofing vector is not expressible at all.
  The test now asserts the property that *is* reachable: a project module
  reached through an `std/`-shaped specifier stays an ordinary module, so its
  `StrBuf` is not the language item. Trust comes from the captured std root,
  never from a spelling.

- **`canonical_semantic::codegen_input_tracks_root_paths_and_options_but_not_linker`**
  — designated `helper.rue` as the root of an import-bearing two-module fixture,
  leaving `main.rue` (the only importer) unreachable. A program's graph covers
  only what discovery reaches from its root, so the root axis moved to an
  import-free pair. The other axes (options, opt level, relocation) are
  unchanged.

- **`session::semantic_diagnostic_identity_includes_the_accepted_import_graph`**
  — retargeted `@import("choice.rue")` by substituting two hand-built graphs
  pointing at `a.rue` and `b.rue`, which no resolver would ever produce for one
  specifier. It now retargets the way production does: the importer-relative
  candidate outranks the project-root one, so adding a sibling `choice.rue` next
  to the importer moves the import off the root-level module without changing a
  single spelling in the importer. That exercises candidate precedence, which
  the old form asserted nothing about.

- **`session::dependency_manifest_keeps_canonical_module_edges_across_physical_relocation`**
  — unchanged fixture, changed expectation. Relocation now moves only the
  canonical read provenance behind an unchanged requested path, so module
  identity, source text, and the published revision are all untouched and the
  dependency manifest is *reused* rather than recomputed (`executions` 2 → 1).
  Stronger than the original claim: the edge is stable without recomputing it.

- **`pipeline_tests::canonical_batch_reports_one_pass_structural_work`** — same
  kind of correction. An import-bearing program is parsed by its discovery
  epoch, and the batch pass that follows reuses that parse. The one-pass claim
  splits accordingly: discovery lexes and parses each module exactly once, and
  the batch adds zero parse work. Every merge/lower/semantic assertion is
  unchanged.

- **`session::supplied_test_graph_is_consumed_without_resolution_fallback`** →
  **`committed_discovery_graph_is_consumed_without_resolution_fallback`** — the
  injected graph it named no longer exists; the property (a committed graph is
  consumed with `imports.executions == 0`, never re-resolved) is now asserted
  against a real discovery close.

## Remaining test-only state

Test-only fields on `CompilerSession` drop from **seven to four**. The remaining
four inject an *input* or an *event* at a deterministic point and select no
alternative algorithm; each carries a comment saying why it cannot go through
the production path:

- `cancel_merge_before_commit`, `cancel_semantic_after_dependency`,
  `cancel_semantic_before_publication` — production cancellation arrives
  asynchronously through a `CancellationToken` and cannot be deterministically
  landed between two chosen steps. The branch each triggers is the same
  cancellation path a real token drives.
- `recipe_cache_slot` — already documented; the test-only overlay path.

## Validation

`rue-compiler-test` 873 pass, `rue-query-test` 55 pass, differential-oracle,
public-API and scaling-matrix targets green, `scripts/rue quick` 36/36 with the
oracle agreeing on every modeled eligible case.

Not run locally: the full `scripts/rue test` suite and non-x86-64 targets. CI is
the full-platform authority.

Fixes RUE-1143

Fixes RUE-1149

### DIFF
--- a/crates/rue-compiler/src/body_overlay.rs
+++ b/crates/rue-compiler/src/body_overlay.rs
@@ -629,13 +629,14 @@ mod tests {
             let merged = crate::merge_parsed_modules(&parsed).unwrap();
             let rir = crate::lower_canonical_rir(&merged).unwrap();
             let options = CompileOptions::default();
-            let imports = crate::bound_definitions::test_fixture_import_graph(&merged).unwrap();
+            let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
             let (definitions, query_declarations, _) =
                 crate::bound_definitions::bind_canonical_declaration_semantics(
                     &merged,
                     &rir,
                     options.preview_features.clone(),
                     options.target,
+                    &imports,
                 )
                 .unwrap();
             let query_shells = crate::canonical_semantic::query_owned_declaration_shells_for_test(
--- a/crates/rue-compiler/src/bound_definitions.rs
+++ b/crates/rue-compiler/src/bound_definitions.rs
@@ -555,30 +555,9 @@ pub(crate) fn bind_canonical_definitions(
     preview_features: PreviewFeatures,
     target: Target,
 ) -> MultiErrorResult<BoundDefinitionSet> {
-    let imports = test_fixture_import_graph(merged)?;
+    let imports = crate::import_graph::import_free_canonical_graph(merged.ast())?;
     bind_canonical_definitions_with_work(merged, rir, preview_features, target, &imports)
         .map(|(definitions, _)| definitions)
-}
-
-#[cfg(test)]
-pub(crate) fn test_fixture_import_graph(
-    merged: &CanonicalMergedProgram,
-) -> MultiErrorResult<crate::CanonicalImportGraph> {
-    crate::test_support::test_fixture_import_graph_parts(
-        merged.ast().root(),
-        merged
-            .ast()
-            .modules()
-            .iter()
-            .map(|module| {
-                (
-                    module.module_id().clone(),
-                    Arc::from(module.physical_path()),
-                )
-            })
-            .collect(),
-        merged.ast().import_directives(),
-    )
 }
 
 /// Bind once and export stable, request-independent declaration semantics
@@ -590,18 +569,18 @@ pub(crate) fn bind_canonical_declaration_semantics(
     rir: &CanonicalRirOutput,
     preview_features: PreviewFeatures,
     target: Target,
+    imports: &crate::CanonicalImportGraph,
 ) -> MultiErrorResult<(
     BoundDefinitionSet,
     Arc<[crate::DurableDeclarationSemantic]>,
     rue_air::SemanticDeclarationExportWork,
 )> {
-    let imports = test_fixture_import_graph(merged)?;
     let bound = crate::canonical_semantic::bind_query_owned_declarations_for_test(
         merged,
         rir,
         preview_features,
         target,
-        &imports,
+        imports,
     )?;
     let manifest = bound.binding_manifest();
     let definitions = issue_bound_definitions(
@@ -649,7 +628,7 @@ pub(crate) fn compare_canonical_durable_declaration_install(
     preview_features: PreviewFeatures,
     target: Target,
 ) -> MultiErrorResult<(crate::DurableSemanticProjectionWork, DeclarationBindingWork)> {
-    let imports = test_fixture_import_graph(merged)?;
+    let imports = crate::import_graph::import_free_canonical_graph(merged.ast())?;
     let ordinary = crate::canonical_semantic::bind_query_owned_declarations_for_test(
         merged,
         rir,
@@ -1252,11 +1231,13 @@ mod tests {
         let parsed = parse_source_snapshot_modules(snapshot).unwrap();
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         bind_canonical_declaration_semantics(
             &merged,
             &rir,
             PreviewFeatures::new(),
             Target::default(),
+            &imports,
         )
         .unwrap()
     }
@@ -1339,7 +1320,7 @@ mod tests {
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
         let current_definitions = bind(&relocated);
-        let imports = test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         let shells = crate::canonical_semantic::query_owned_declaration_shells_for_test(
             &merged,
             &rir,
@@ -1387,7 +1368,7 @@ mod tests {
         let parsed = parse_source_snapshot_modules(&input).unwrap();
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
-        let imports = test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         let shells = crate::canonical_semantic::query_owned_declaration_shells_for_test(
             &merged,
             &rir,
@@ -1420,7 +1401,7 @@ mod tests {
         let parsed = parse_source_snapshot_modules(&input).unwrap();
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
-        let imports = test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         let shells = crate::canonical_semantic::query_owned_declaration_shells_for_test(
             &merged,
             &rir,
@@ -1479,7 +1460,7 @@ mod tests {
         )
         .unwrap();
         // A fresh oracle epoch remains valid after the query-owned install.
-        let imports = test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         crate::canonical_semantic::bind_query_owned_declarations_for_test(
             &merged,
             &rir,
--- a/crates/rue-compiler/src/canonical_semantic.rs
+++ b/crates/rue-compiler/src/canonical_semantic.rs
@@ -3784,7 +3784,7 @@ mod tests {
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
         let options = CompileOptions::default();
-        let imports = crate::bound_definitions::test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
 
         let ordinary = match rue_air::Sema::new_synthetic(
             rir.rir(),
@@ -3825,7 +3825,7 @@ mod tests {
         let parsed = parse_source_snapshot_modules(&source).unwrap();
         let merged = merge_parsed_modules(&parsed).unwrap();
         let rir = lower_canonical_rir(&merged).unwrap();
-        let imports = crate::bound_definitions::test_fixture_import_graph(&merged).unwrap();
+        let imports = crate::import_graph::import_free_canonical_graph(merged.ast()).unwrap();
         crate::bound_definitions::configure_canonical_sema(
             &merged,
             &rir,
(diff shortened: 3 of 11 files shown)


steveklabnik:
Status on the two things that have knocked this out of the merge queue. Neither originates in this PR.

## 1. `clippy` is red on trunk, not on this branch

The `clippy` gate fails on `3d53a52d` with `needless_borrow` errors. It reproduces on plain trunk:

```
$ git checkout --detach origin/trunk   # 00f7cb24
$ ./clippy.sh
error: this expression creates a reference which is immediately dereferenced by the compiler
  = help: ...rust-clippy/rust-1.92.0/index.html#needless_borrow
clippy gate FAILED
```

291 errors, entirely within `crates/rue-air/src/sema/`:

| count | file |
|---|---|
| 63 | `sema/control_flow.rs` |
| 54 | `sema/analysis/ownership.rs` |
| 42 | `sema/aggregates.rs` |
| 36 | `sema/analysis/type_inference.rs` |
| 36 | `sema/analysis/intrinsics.rs` |
| 18 | `sema/analysis/calls.rs` |
| 15 | `sema/analysis/builtin_ops.rs` |
| 9 | `sema/analyze_ops.rs` |
| 6 | `sema/typeck.rs`, `sema/analysis/instructions.rs`, `sema/analysis.rs` |
| 3 | `sema/comptime_eval.rs` |

This branch cannot be the cause — it modifies nothing outside `crates/rue-compiler`:

```
$ git diff origin/trunk HEAD --stat -- . ':!crates/rue-compiler'
(empty)
```

The lint URL pins `rust-1.92.0`, and trunk's own queue run passed `clippy` earlier, so this looks like a toolchain bump reaching CI after that run rather than a code change. The gate's own message points at the two options: fix the lints, or adjust `deny_lints`/`allow_lints` in `toolchains/rust/BUCK`. Both are repo-wide calls affecting every open PR, so I have not made either here — a 291-lint sweep across `rue-air` does not belong in a PR about test-only compiler paths. Happy to put one up separately.

## 2. The earlier queue removal was a borderline timeout, also not this PR

The first dequeue was `cli.examples::mosaic::main` on `macos-cli-shard-3`:

```
TIMEOUT: compiler phase timed out (elapsed=30009 ms, budget=30000 ms)
test result: FAILED. 449 passed; 1 failed
```

9 ms over a 30 s budget — 0.03%. Mosaic is a 12,360-line program pinned to the `heavyweight` (30 s) tier in `execution_contracts.toml`, while `harbor` and `lattice` sit on `heavyweight_long` (120 s).

I measured it rather than assuming, building both binaries and interleaving runs to cancel load drift:

| pair | trunk | this branch | Δ |
|---|---|---|---|
| 1 | 39726 | 44680 | +4954 |
| 2 | 43325 | 42129 | −1196 |
| 3 | 41405 | 39936 | −1469 |
| 4 | 37060 | 39581 | +2521 |
| 5 | 39308 | 40628 | +1320 |

The paired differences flip sign, and trunk alone spans 37.1 s–43.3 s on an identical binary — within-binary noise is roughly 5× any between-binary difference (paired t ≈ 1.05). No detectable regression. There is also no mechanism for one: the production change here is an `Option` check per semantic *attempt*, not per node.

Worth noting separately that mosaic sits at its tier boundary and will keep flaking regardless of this PR; that is a real issue but not one to resolve by widening a gate to land a change.

## Where this leaves the PR

`3d53a52d` merges current trunk in. Locally on that commit: `rue-compiler-test` 873 pass, `rue-query-test` 55, differential-oracle 4, `scripts/rue quick` 36/36 with the oracle agreeing on every modeled eligible case, `fmt` clean. The 31 non-clippy checks that completed on the previous head were all green.

Re-queueing will keep failing on `clippy` until trunk recovers. I will re-run this once it does.

---

steveklabnik:
Correcting my previous comment: the claim that `clippy` is broken on trunk was wrong, and so was the toolchain-bump explanation.

**What I got wrong.** I ran `./clippy.sh` on detached trunk, saw `needless_borrow` errors, and concluded trunk was red. That inference doesn't hold. The same check at `73252e83` — where CI's `clippy` job was **green** at 02:52 — produces a byte-identical result locally:

| commit | CI `clippy` | local `rue-air[clippy.txt]` |
|---|---|---|
| `73252e83` | success | 64547 bytes, 99 `^error` lines |
| `3d53a52d` | failure | 64547 bytes, 99 `^error` lines |

Since my local run flags a commit CI passes, it is not evidence about CI's verdict, and I shouldn't have reported it as such.

The toolchain claim was also wrong. `RUST_VERSION = "1.92.0"` landed in `11f9e930` on 2026-07-24, days before this branch's base, and that commit is an ancestor of both trunk and this branch. Lint policy is likewise unchanged — it only ever added `clippy::needless_borrows_for_generic_args`, a different lint.

**What remains true.** `rue-air`'s source is identical across the green and red runs: this branch modifies nothing outside `crates/rue-compiler` (`git diff origin/trunk HEAD -- . ':!crates/rue-compiler'` is empty), and neither `21df7ac4` nor `00f7cb24` touches `crates/rue-air` at all. So whatever flipped the gate at 04:03, it was not a change to the code the failure names.

**What I can't explain.** Why the gate's verdict changed on unchanged input. The `clippy` job reads the BuildBuddy remote action cache, so a cache miss recomputing those diagnostics fresh is my leading candidate — that would make the findings real but normally masked — though I could not confirm it, and re-running the job returns 403 for me. Someone with permission re-running `clippy` on this run would settle it: green on unchanged content confirms the gate is cache-dependent.

The other 31 checks on `3d53a52d` are green, including all of `mosaic` this time.

---

steveklabnik:
Resolving the `clippy` failure on this PR: the cause is real lint debt on trunk, and the fix is #2001. Correcting two things I got wrong above along the way.

**What actually happened.** `crates/rue-air/src/sema/` carries 98 clippy findings (97 `needless_borrow`, 1 `get_first`). They arrived with RUE-1092 (`14edd527`). This branch forked at `094d15af`, *before* that commit, so its own CI ran against a `rue-air` that predated them and `clippy` was legitimately green. Merging trunk in at `3d53a52d` pulled them in, and `clippy` went red on a branch that never touched `rue-air`.

**Correction 1.** I claimed earlier that `rue-air` was identical between the green and red runs. It is not — 11 files, 622 insertions. I had diffed `14edd527..00f7cb24`, the two newest trunk commits, when the relevant range started at this branch's actual merge base. That error is what made the failure look inexplicable.

**Correction 2.** I said trunk's `clippy` was broken and blamed a toolchain bump. The toolchain claim was wrong — `RUST_VERSION = "1.92.0"` landed in `11f9e930` on 2026-07-24, days before this branch's base. And trunk's CI is green, exactly as reported. What is true is narrower and worse: trunk's *code* carries the findings while trunk's *checks* pass, because the gates were not reading it.

**Why the gates missed it** (detail in #2001, tracked as RUE-1152):

- `fmt` checked 1 file of 281. `//:fmt-check` took its file list from a root-package `glob(["crates/**/*.rs"])`, and Buck globs do not descend into subpackages. All 30 crates owning a `BUCK` file were invisible; only `crates/rue-runtime-asan/` — the one crate without one — was ever checked.
- `clippy` skipped diagnostics it could not read: `[ -s "$ARTIFACT" ] || continue` is false both for an empty file (clean) and a missing one (never read). A remote cache hit can satisfy the build without materializing that file.
- `fmt.sh` exited **0** when discovery found no sources at all.

So my local runs and CI were both right and were measuring different things, which is what I misread as environment noise.

**Sequencing.** #2001 fixes the gates and the 98 findings, and is green across all 32 checks with `fmt` now covering 281 files and `clippy` asserting it read all 66 targets. Once it lands, this branch needs a trunk merge to pick up the `rue-air` fixes, after which `clippy` here goes green for the right reason. No change is needed to the RUE-1143/RUE-1149 work itself — the other 31 checks on `3d53a52d` were already passing.

---

PR #1999, opened 2026-07-27
RUE-786: attribute whole-pipeline and backend subphase timing

steveklabnik:
Closes RUE-786.

## Problem

The timing surface folded the driver's source loading, most of the semantic query path, and all of code generation and linking into a handful of coarse spans. Measured on trunk, 25–45% of the compiler's root span was claimed by no leaf pass at all, so latency work could not be aimed at a phase and a warm-rebuild regression could hide entirely inside the residual.

## What changed

Spans were added across the four dark regions:

- **Driver** — `source_loading`, the per-round discovery frontier (`import_revision_inputs`, `import_stage`, `import_parse_staging`, `parse_query_key` / `parse_program` / `parse_query_commit`, `import_plan_build`, `import_plan_publish`, `import_frontier`, `import_read`), `import_discovery_close`, and `toolchain_acquisition`.
- **Pipeline** — `module_index_projection`, `canonical_merge`, `module_rir_lowering`, `rir_projection`, the declaration shell and nucleus projections, the per-body query and coordinator phases, `declaration_reuse`, and `semantic_finalization`.
- **Backend** — `mir_lowering`, `register_allocation`, `mir_peephole`, `mir_scheduling`, `mir_verification`, `string_table_compaction`, `machine_emission`, and `object_serialization`.
- **Linker** — `link_parse_objects`, `link_archive_resolve`, and `link_layout` / `link_relocate` / `link_emit` on both the ELF and Mach-O paths.

The full resulting tree is documented in `docs/process/perf-baseline.md`.

### Rayon span parentage

Backend subphases execute inside a `par_iter`, and Rayon workers do not inherit the calling thread's current span. Left alone, every subphase would have been reported as an independent timing root with nondeterministic parentage (work stealing means the calling thread gets correct nesting and the workers do not). The `codegen` span is now held by value and re-entered as the first statement inside the closure. A regression test asserts `codegen -> mir_lowering` and friends in the batch edge set, and a `--jobs 0` run confirms `compile` is still the sole root with zero overlapping leaf work.

### Driver phases

Executable publication deliberately runs after the compiler's timing root closes, so it can be neither nested under `compile` (mis-attribution) nor made a second root (`total_ms` is consumed as "compiler time" by the dashboard, scaling bench, and scenario tooling). A span may now declare `driver_phase = true`; the timing layer routes it — and its children — into a separate `driver_phases` array instead of `passes`, and excludes it from the root-duration union. It is a breakdown of `process - total_ms`, never of `total_ms`. `compile` remains the sole timing root and schema v2 is unchanged apart from the new optional field, which is omitted entirely when a run measures no driver phase.

### Harness bound on unattributed time

`scripts/perf-baseline.py` now computes the unattributed share per sample (`max(root - leaf work, 0) / root`, paired before the median, consistent with the existing accounting), reports it in text and JSON, and fails when a workload's median exceeds `--max-unattributed-percent`. Parallel leaf spans can overlap and inflate leaf work, which only makes the check more permissive, so it cannot fail spuriously under `--jobs > 1`.

## Measured effect

Release compiler, `-O0`, `--jobs 1`, via `scripts/perf-baseline.py`:

| workload | unattributed share |
|---|---|
| hello | 13.0% |
| fibonacci | 13.2% |
| quicksort | 14.3% |
| multi_file | 16.0% |
| structs | 18.6% |
| deep_nesting | 23.9% |
| array_heavy | 25.0% |
| register_pressure | 25.4% |
| arithmetic_heavy | 28.1% |
| many_functions | 29.2% |
| large_structs | 39.9% |

The default bound is 50, leaving headroom over the worst workload. Verified in both directions: the corpus passes at the default, and a tighter bound exits 1 naming the offending workloads and their measured shares.

## Where the remaining residual is

It is **not** spread across the compiler any more. Every remaining large gap has the same shape — an aggregate whose children are the compute work, and whose residual is rue-query's own per-request cost (key hashing, memo lookup, dependency recording, terminal publication in `request_impl`):

| aggregate | what its residual measures |
| --- | --- |
| `body_transaction` | one `body_transaction` request per reached body |
| `declaration_shell_projection` | one `declaration_occurrence_indexes` request per module |
| `declaration_nucleus` | one `semantic_nucleus` request per declaration |
| `parse_program` | one `parse_modules` request per module |

Attributing that needs spans inside the query runtime's request path, which is the accounting boundary RUE-817 owns. I deliberately left it out of this PR rather than reaching into `rue-query`'s core — happy to fold it in here instead if reviewers would rather see it done in one go.

Two things worth noting about the numbers. Merging trunk moved the profile on its own: the recent query-attempt work raised exactly the request-machinery share this instrumentation now makes visible. And `codegen` and `linker` used to be **leaves**, so their internals counted as fully attributed work; they are aggregates now and only their leaf children count, which makes a pre-change leaf sum non-comparable to a post-change one. Both are called out in `docs/process/perf-baseline.md`.

Refs RUE-817.

## Testing

- `scripts/rue quick`; full spec suite (2178 passed).
- Unit tests for `rue`, `rue-compiler`, `rue-codegen`, `rue-linker`, `rue-query`, `rue-air`, `rue-cfg`, `rue-rir`, `rue-parser`.
- `scripts/test-benchmark-tools.py` (117 passed), including new coverage for driver-phase aggregation, driver-phase shape/membership validation, the paired unattributed share, and the bound itself.
- Existing nesting tests asserted the old shallower tree and were updated to assert the new chains explicitly rather than loosened.
- Manual: `--time-passes` text output, `--benchmark-json`, `--jobs 0`, and every `--emit` stage.
- `scripts/perf-baseline.py` end to end against a release compiler.

### DIFF
--- a/crates/rue-codegen/BUCK
+++ b/crates/rue-codegen/BUCK
@@ -30,6 +30,7 @@ rue_crate(
         "//crates/rue-target:rue-target",
         "//third-party:fixedbitset",
         "//third-party:lasso",
+        "//third-party:tracing",
     ],
     test_deps = [
         "//crates/rue-lexer:rue-lexer",
--- a/crates/rue-codegen/src/aarch64/mod.rs
+++ b/crates/rue-codegen/src/aarch64/mod.rs
@@ -31,6 +31,7 @@ use rue_air::FrozenTypeInternPool;
 use rue_cfg::ValidatedCfg;
 use rue_error::CompileResult;
 use rue_target::Target;
+use tracing::info_span;
 
 use crate::MachineCode;
 // Re-export from parent
@@ -133,17 +134,26 @@ fn generate_inner(
             | Aarch64Inst::StringConstCap { string_id, .. } => Some(*string_id),
             _ => None,
         });
-    let (local_strings, string_id_remap) =
-        crate::compact_string_table(strings, atoms, referenced_strings, require_complete_atoms)?;
-    for inst in prepared.mir.instructions_vec_mut() {
-        let string_id = match inst {
-            Aarch64Inst::StringConstPtr { string_id, .. }
-            | Aarch64Inst::StringConstLen { string_id, .. }
-            | Aarch64Inst::StringConstCap { string_id, .. } => string_id,
-            _ => continue,
-        };
-        *string_id = string_id_remap[string_id];
-    }
+    let local_strings = {
+        let _span = info_span!("string_table_compaction").entered();
+        let (local_strings, string_id_remap) = crate::compact_string_table(
+            strings,
+            atoms,
+            referenced_strings,
+            require_complete_atoms,
+        )?;
+        for inst in prepared.mir.instructions_vec_mut() {
+            let string_id = match inst {
+                Aarch64Inst::StringConstPtr { string_id, .. }
+                | Aarch64Inst::StringConstLen { string_id, .. }
+                | Aarch64Inst::StringConstCap { string_id, .. } => string_id,
+                _ => continue,
+            };
+            *string_id = string_id_remap[string_id];
+        }
+        local_strings
+    };
+    let _emission_span = info_span!("machine_emission").entered();
     let emitter = Emitter::new(
         &prepared.mir,
         prepared.total_locals,
--- a/crates/rue-codegen/src/codegen_pipeline.rs
+++ b/crates/rue-codegen/src/codegen_pipeline.rs
@@ -8,6 +8,7 @@
 use rue_air::{ArgClass, ArgConvention, FrozenTypeInternPool, NativeCallAbi, ReturnClass};
 use rue_cfg::{Cfg, CfgArgMode, CfgInstData, CfgValue};
 use rue_error::{CompileError, CompileResult, ErrorKind};
+use tracing::info_span;
 
 use crate::frame_layout::{FrameLayout, SavedRegScheme};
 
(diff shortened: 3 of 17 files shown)


DorianListens:
@steveklabnik if I understand the table correctly, we still have 9-34% unattributed time after this change, is that correct? That seems high to me.

steveklabnik:
Correct, and worth digging into — thanks for pushing on it. I broke the residual down per owning aggregate rather than leaving it as one number, and the answer changed what I did.

Two of the gaps turned out to be regions I had reclassified but not actually attributed, and those are now fixed (pushed in 314e512): the parse query's own keying and terminal publication (`parse_query_key` / `parse_program` / `parse_query_commit`), and the declaration shell epoch (`declaration_shell_projection` / `declaration_shell_prepare`). On `large_structs` that alone was ~17% of the root.

What's left is not spread across the compiler. It is one thing wearing four hats — rue-query's own per-request cost:

| aggregate | residual on `large_structs` | what it measures |
| --- | --- | --- |
| `body_transaction` | 7.5% of root | one request per reached body |
| `declaration_shell_projection` | 9.3% | one `declaration_occurrence_indexes` request per module |
| `declaration_nucleus` | — | one `semantic_nucleus` request per declaration |
| `parse_program` | 7.8% | one `parse_modules` request per module |

Each is an aggregate whose children are the compute work and whose residual is `request_impl` itself: key hashing, memo lookup, dependency recording, terminal publication. So the honest read of the remaining number is "the query runtime costs this much", not "we don't know where the time goes" — which is the thing the issue was actually asking for, since previously it was genuinely unknown.

Attributing it means putting spans inside `rue_query::Runtime::request_impl`. That is the accounting boundary RUE-817 owns, and it is a different subsystem from anything else in this PR, so I left it out rather than reaching into the query core unasked. Happy to fold it in here if you'd rather see it done in one go — it is a contained change, just a scope call that isn't mine to make.

One caveat on comparing against the numbers in the issue: `codegen` and `linker` used to be *leaves*, so their internals counted as fully attributed. They are aggregates now and only their leaf children count, so the metric got strictly stricter across this change. Merging trunk also moved the profile on its own — the recent query-attempt work raised exactly the request-machinery share this instrumentation now exposes. That is why the table in the PR body reads 13–40% rather than the 9–34% you saw.

The harness bound is set at 50% to keep headroom over the worst workload today, and both the docs and the bound's comment name the query runtime as the ratchet to drive it down.

---

steveklabnik:
The merge queue bounced this with `MERGE_CONFLICT`, which was misleading — the PR was `mergeable_state: clean` against `trunk` the whole time, and `trunk` had not moved (still `21df7ac`, the commit this branch already had merged in).

The cause was the shape of the branch, not its content: it carried a merge commit (`0e05433`, trunk merged in after the backend refactor landed), and replaying that through the queue's temp branch reintroduced the original conflicts.

Fixed by linearizing — the branch is now a single commit on top of `trunk` (`14e7461`). The tree is bit-for-bit identical to what CI already validated at `314e512`; `git diff 314e512 14e7461` is empty. That did require a force-push to my own branch, with no content change.

Re-running CI now. Auto-merge (squash) is still enabled, so it should re-queue on its own once checks are green.

---

PR #2000, opened 2026-07-27
Run the heavy corpora as cacheable build actions

steveklabnik:
## Problem

buck2 re-executes every `buck2 test` invocation. Test executions are not actions: they are handed to the test executor, they never appear in buck2's `Commands: N (cached: ...)` accounting, and OSS buck2 ships no test-result cache. The merge queue therefore re-ran every corpus in full on each merge, even when the merge commit's tree was byte-identical to the tree the PR run had just validated.

Measured on three CI runs of one such tree (`1c9fe424`) — PR runs [30232087197](https://github.com/rue-language/rue/actions/runs/30232087197) and [30232994034](https://github.com/rue-language/rue/actions/runs/30232994034), merge_group run [30233477187](https://github.com/rue-language/rue/actions/runs/30233477187):

| lane | PR run 1 | PR run 2 | merge_group |
| --- | ---: | ---: | ---: |
| linux-x64 cli-shard-1 | 10:27.1 | 11:28.9 | 11:18.2 |
| linux-arm64 cli-shard-1 | 11:35.7 | 11:36.3 | 11:21.6 |
| linux-x64 spec | 1:30.1 | 1:40.5 | 1:41.2 |
| build actions (corpus jobs) | 463–465 / 465 | **465 / 465** | **465 / 465** |

The build converges to a perfect hit rate; the corpora re-execute every time. The slowest merge_group job was 744s — 7s setup, 59s of analysis and CAS materialization for a fully cached build, and **678s (91%) of test execution**.

## What changed

`cached_corpus_suite` (`corpus.bzl`) splits each heavy corpus in two:

- a **genrule** that runs the harness via `scripts/corpus-action` and writes a stamp on success — an ordinary action keyed on its declared inputs, so the PR run uploads it and the merge_group run reads it back;
- a thin **`sh_test`** asserting the stamp, which keeps the suite's name, labels, and `Pass: root//:NAME (time)` line. `scripts/ci-heavy-suite`'s result assertion and `test.sh`'s RUE-924 corpus-omission audit therefore work unchanged.

Converted: the CLI corpus and its four shards, Caldera, spec, UI, release smoke.

Not converted: `//:reproducible-programs` and `//:oracle-diff-generated-smoke`. Their harnesses are shell scripts that read repository paths directly rather than through declared env inputs, so that contract has to be established first or caching them would be a false pass rather than a win. Both are off the merge queue's critical path. Marked inline.

Per-target executor timeouts move from `ci-heavy-suite` to each suite's `timeout_seconds` — the test executor now only asserts a sub-second stamp, so a `--timeout` there would bound the wrong thing while reading as if the corpus were still bounded. Per-case budgets in `execution_contracts.toml` remain the honest gates.

## Two contracts this makes load-bearing

Documented in `corpus.bzl` and `docs/process/build-cache.md`:

1. **Declared inputs now decide correctness.** Under a plain `sh_test` an undeclared input was merely untracked — the suite re-ran regardless. Here it is a false pass: change a file the action does not name and the corpus reports success against the previous tree's result.
2. **Paths must be absolute.** `$(location ...)` expands absolute in an `sh_test` (which runs from the project root) but relative to the action's working directory in a genrule. The harnesses spawn the compiler with a case temp dir as cwd, and `find_dir` falls back to relative defaults, so a relative value can resolve to a real but wrong directory and quietly test nothing. `corpus-action` resolves everything named in a suite's `absolutize` list immediately before the harness runs; it cannot be done in BUCK because a genrule cmd has no shell command substitution — `$(...)` is Buck's own macro syntax.

## Testing

Mechanism, against the pinned buck2 with byte-identical inputs in one daemon:

| | unchanged inputs |
| --- | --- |
| `sh_test` | re-runs (5.0s → 5.0s) |
| `genrule` | **not re-run** (4.1s → 0.098s) |

`//:spec-tests` end to end:

- ran the real corpus as a build action — 2177 passed, 1 failed against a deliberately broken case;
- unchanged inputs: **0.4s**, down from ~70s cold;
- edited spec case: re-ran in 64s, so invalidation tracks inputs;
- failing case: build error naming the case, **no stamp written**, nothing cached;
- `buck2 test //:spec-tests` → `✓ Pass: root//:spec-tests (0.0s)`.

`//:cli-tests-shard-0` ran as a build action in ~3.5 min, matching its 3:39 CI time, so it executed the real slice rather than no-opping.

Also passing: `//:corpus-action-tests` (new, 12 checks over the stamp-only-on-success, absolutization, fail-closed, plumbing-isolation and timeout contracts), `//:wrapper-script-tests`, `//:cli-shard-coverage-validation`. Heavy-suite label discovery still returns all ten targets, and a `--exclude rue_heavy_suite --always-exclude` pass does not build the corpus actions (0.2s against an invalidated suite), so the broad pass stays fast.

**Not verifiable locally:** actual cross-run reuse needs the shared cache, which this checkout has no key for — OSS buck2 has no persistent local action cache, so a locally reverted edit re-runs rather than hitting the earlier result. Cross-run and cross-platform behavior is what CI on this PR will show. The per-platform digest separation rests on each action depending on the host-built `rue` and harness binaries; worth confirming against the first merge_group run that macOS and Linux lanes do not share entries.

Refs RUE-1118.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Apefw5j746c6nptMFptfBT)_

### DIFF
--- a/BUCK
+++ b/BUCK
@@ -1,8 +1,8 @@
 # Repo-root test-suite targets (RUE-144 / RUE-132).
 #
-# Each sh_test ties a test-harness binary to the rue compiler and the on-disk
-# inputs the suite actually reads (cases/, std/, docs/), so Buck owns the binary
-# handoff and caches each suite against its real inputs:
+# Each suite ties a test-harness binary to the rue compiler and the on-disk
+# inputs it actually reads (cases/, std/, docs/), so Buck owns the binary
+# handoff and keys each suite on its real inputs:
 #
 #   buck2 test //...        # runs unit tests + spec/UI/CLI suites + repo gates
 #
@@ -10,17 +10,41 @@
 # under std/ re-runs the CLI and spec suites (std/ MUST be a declared input here
 # or either suite could get false cache hits on standard-library changes).
 #
+# RUE-1118: the heavy corpora run through `cached_corpus_suite` rather than a
+# bare sh_test. buck2 re-executes every test invocation — test executions are
+# not actions and never reach the action cache — so a plain sh_test re-ran the
+# whole corpus on each merge even when the merge commit's tree was byte-identical
+# to the tree the PR run had just validated. The macro moves the harness into a
+# cacheable build action and leaves a thin sh_test asserting its stamp, so the
+# suite keeps its name, labels, and result line. See corpus.bzl for the input
+# contract this makes load-bearing.
+#
 # Mechanics: the harness binaries already locate everything via env vars
-# (rue-test-runner's find_rue_binary / find_dir), `$(exe_target ...)` /
-# `$(location ...)` expand to absolute paths, and sh_test runs from the
-# project root — so the harness binary itself can be the `test` command and
-# no wrapper script is needed. A filegroup's output directory is named after
-# the rule and contains its srcs at package-relative paths, hence the
-# `$(location ...)/cases` shape.
+# (rue-test-runner's find_rue_binary / find_dir), and a filegroup's output
+# directory is named after the rule and contains its srcs at package-relative
+# paths, hence the `$(location ...)/cases` shape. In an sh_test those macros
+# expand to absolute paths and the test runs from the project root; in a
+# genrule they are relative to the action's working directory, which is why
+# cached_corpus_suite routes them through scripts/corpus-action.
 #
 # These suites live at the repo root rather than in the harness crates' BUCK
 # files so that `buck2 test //crates/...` (quick-test.sh, test.sh's filtered
 # path) still means "unit tests only".
+
+load(":corpus.bzl", "cached_corpus_suite")
+
+# The two halves of a cached corpus suite: the action wrapper that runs a
+# harness and writes its stamp, and the thin check that asserts the stamp.
+sh_binary(
+    name = "corpus-action",
+    main = "scripts/corpus-action",
+)
+
+sh_binary(
+    name = "corpus-stamp-check",
+    main = "scripts/corpus-stamp-check",
+    visibility = ["PUBLIC"],
+)
 
 # The formatting gate lives in fmt.sh, not here. A `fmt-check` sh_test used to
 # sit at this spot, taking its file list from `glob(["crates/**/*.rs"])`. A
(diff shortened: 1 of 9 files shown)


steveklabnik:
## Blocked on a pre-existing trunk failure, not on this PR

`clippy` is red on `21db8fa`. It is not caused by this change and I am not fixing it here without a decision, because both available fixes are substantially larger than this PR and one of them turns trunk red.

**This PR changes zero Rust files.** The diff is 9 files: `BUCK`, `corpus.bzl`, four shell scripts, one shell test, and a doc. The 98 reported violations are all `needless_borrow` in `crates/rue-air/**`.

### The clippy gate is fail-open

The same branch passed clippy twice and then failed with identical Rust code, and trunk has not moved (still `00f7cb2`). The passing run explains itself:

```
Running clippy on 66 targets...
./clippy.sh: line 64: printf: write error: Broken pipe
./clippy.sh: line 64: printf: write error: Broken pipe
./clippy.sh: line 64: printf: write error: Broken pipe
clippy gate passed: no lint violations.
```

Line 64 is the detection itself:

```bash
if printf '%s\n' "$CLEANED" | grep -qE '^error'; then
```

`clippy.sh` runs under `set -euo pipefail`. `grep -q` exits the moment it matches, closing the pipe; `printf` then fails with EPIPE, and `pipefail` propagates *printf's* non-zero status as the pipeline's. The `if` reads false and the violation is discarded.

The failure mode is inverted: **the broken-pipe message only appears when grep did match.** Those three lines are three targets whose violations were found and silently dropped, and the run reported "no lint violations." Whether the race is lost depends on output size and scheduling, which is why the same tree passes and fails nondeterministically.

Verified locally on this branch: `./clippy.sh` exits 1 with `error: aborting due to 98 previous errors`. Same 98, no Rust changes — they are on trunk today.

### Why I stopped rather than fixing it

Two separable problems, neither in this PR's scope:

1. **`clippy.sh` fail-open race** — a small, confident fix, but fixing it alone makes clippy correctly fail on trunk and every open PR.
2. **98 `needless_borrow` sites in rue-air** — mechanical, but a large unrelated diff that would bury a corpus-caching change.

They have to land together to avoid a red trunk, and together they are their own PR. Fixing them here would make this PR about three unrelated things.

Worth noting the theme: a cached-or-swallowed result reporting success without doing the work is the same failure this PR is about. The gate has been reporting green while finding violations; this PR's first CI run surfaced a UI-corpus input that had been passing against an undeclared `std/` for the same reason. That is an argument for fixing the gate properly rather than quickly.

Nothing to do on this branch until that is sorted. Every other check is green, including all 18 platform-corpus jobs across linux-x64, linux-arm64 and macOS.

---

steveklabnik:
Superseded by #2044, which is this branch rebased onto current trunk. GitHub refuses to reopen a PR whose branch was force-pushed, so the rebase had to become a new PR rather than a revival of this one.

For the record on why this is being picked back up: the blocker described in the comment above — the fail-open clippy gate — was fixed by #2001 (merged 07-27 06:21) and hardened by #2039. This PR was closed at 17:29 the same day, eleven hours after that cleared, with no closing comment. Nothing was wrong with the approach, and all 18 platform-corpus jobs were green here.

The rebase was not a replay: trunk gained the RUE-1157 test tiers, replaced `//:cli-tests-caldera` with the `large-example-*` family, deleted the benchmark infrastructure this branch still referenced, and added the RUE-1159 timeout policy and flake-detection repetitions. #2044 describes each resolution, and flags two decisions the conversion now forces that this PR never faced.

---

PR #2001, opened 2026-07-27
Make the clippy and fmt gates fail instead of silently skipping

steveklabnik:
Both lint gates could report a pass for code they never examined. Each failed open. This fixes the gates, then fixes the 98 findings that reached trunk underneath them.

## The fmt gate checked 1 file of 281

The CI `fmt` job ran `buck2 test //:fmt-check`, whose file list came from a root-package glob:

```python
_RUST_SOURCES = glob(["crates/**/*.rs"])
```

A Buck glob does not descend into subpackages, and all 30 crates that own a `BUCK` file are subpackages. So the list resolved to exactly one entry:

```console
$ ./buck2 cquery 'root//:fmt-check' --output-attribute args
"args": ["--edition", "2024", "--check", "crates/rue-runtime-asan/src/main.rs"]
```

`crates/rue-runtime-asan/` is the only crate without a `BUCK` file. Every other Rust source was invisible to the gate, which reported success for all 280 of them.

CI now calls `./fmt.sh check`, whose `find`-based discovery covers every source:

```console
$ ./fmt.sh check
All 281 files formatted correctly!
```

The `//:fmt-check` target is removed rather than left certifying one file. Restoring cache-aware checking inside Buck means enumerating sources per crate instead of one root glob — a real change, not a glob tweak, so it is follow-up work and the BUCK comment says so.

`fmt.sh` carried a matching hazard: empty discovery printed "No Rust files found" and exited **0**, so a broken checkout or a bad discovery change read as a clean tree. It now fails, and reports the file count in both modes so coverage is visible in the log.

## The clippy gate scored unread artifacts as clean

`clippy.sh` builds a `[clippy.txt]` subtarget per crate and greps each for `^error`. The loop skipped anything it could not read:

```bash
[ -s "$ARTIFACT" ] || continue
```

`-s` is false both for a file that is empty (genuinely clean) and for one that does not exist. Those are different claims, and collapsing them means a target whose diagnostics were never materialized counts as passing. A remote cache hit can satisfy the build without writing that file to local disk.

Three changes:

- `--materializations=all`, so a cache hit still lands the diagnostics on disk. This is load-bearing, not a tuning knob.
- A missing artifact is an error, not a skip.
- Files-read must equal targets-enumerated, so partial coverage can no longer pass as full coverage.

```console
$ ./clippy.sh
Running clippy on 66 targets...
clippy gate passed: no lint violations.
```

## The 98 findings this was hiding

97 `needless_borrow` and 1 `get_first`, across 12 files in `crates/rue-air/src/sema/`:

| count | file |
|---|---|
| 21 | `sema/control_flow.rs` |
| 18 | `sema/analysis/ownership.rs` |
| 14 | `sema/aggregates.rs` |
| 12 | `sema/analysis/intrinsics.rs` |
| 12 | `sema/analysis/type_inference.rs` |
| 6 | `sema/analysis/calls.rs` |
| 5 | `sema/analysis/builtin_ops.rs` |
| 3 | `sema/analyze_ops.rs` |
| 2 | `sema/analysis.rs`, `sema/analysis/instructions.rs`, `sema/typeck.rs` |
| 1 | `sema/comptime_eval.rs` |

That total matches the `aborting due to 98 previous errors` seen when a run did evaluate them. They entered with RUE-1092 (`14edd527`) and surfaced only when a PR run happened to read the artifacts — which presents as "this PR broke clippy" on a branch that never touched `rue-air`, which is what prompted this.

Every edit is the lint's own machine-applicable suggestion: drop a borrow the compiler immediately dereferences, and `args.get(0)` → `args.first()`. No control-flow or signature changes. Applied programmatically from the diagnostics with each target character verified before editing (98 applied, 0 skipped).

## Validation

`rue-air` 76 pass, `rue-compiler-test` 873, `rue-query-test` 654, differential-oracle 4. `./clippy.sh` green over 66 targets with `rue-air`'s `clippy.txt` at 0 bytes. `./fmt.sh check` green over 281 files.

Not run locally: the full `scripts/rue test` suite and non-x86-64 targets. CI is the full-platform authority — and for this PR in particular, CI is now checking something.

Fixes RUE-1152

---
_Generated by [Claude Code](https://claude.ai/code/session_01XbGPNzfcXMgL4p44rVXuSu)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,12 @@ jobs:
         uses: facebook/install-dotslash@v2  # pinned: the moving `latest` branch broke macOS runners (sha256sum flags) on 2026-06-18
 
       - name: Check formatting
-        # Buck declares all Rust sources as inputs and applies the hermetic
-        # rustfmt target's platform-specific runtime environment.
-        run: ./buck2 test //:fmt-check
+        # Runs the hermetic rustfmt over every Rust source. This used to be
+        # `buck2 test //:fmt-check`, whose file list came from a root-package
+        # `glob(["crates/**/*.rs"])` -- and Buck globs do not descend into
+        # subpackages, so the 30 crates owning a BUCK file were all invisible.
+        # That gate checked 1 file of 281 and passed regardless (RUE-1152).
+        run: ./fmt.sh check
 
       - name: Check production debug assertions
         run: scripts/validate-debug-assert-policy.py
--- a/BUCK
+++ b/BUCK
@@ -22,23 +22,17 @@
 # files so that `buck2 test //crates/...` (quick-test.sh, test.sh's filtered
 # path) still means "unit tests only".
 
-# Formatting is a first-class repository check. The source files are both
-# passed to rustfmt and declared as resources, so Buck invalidates the cached
-# result when any Rust source changes. The rustfmt target's RunInfo owns host
-# selection and dynamic-library setup; write-mode formatting remains in
-# fmt.sh because build/test actions must not mutate the source tree.
-_RUST_SOURCES = glob(["crates/**/*.rs"])
-
-sh_test(
-    name = "fmt-check",
-    test = "toolchains//rust:rustfmt",
-    args = [
-        "--edition",
-        "2024",
-        "--check",
-    ] + _RUST_SOURCES,
-    resources = _RUST_SOURCES,
-)
+# The formatting gate lives in fmt.sh, not here. A `fmt-check` sh_test used to
+# sit at this spot, taking its file list from `glob(["crates/**/*.rs"])`. A
+# Buck glob does not descend into subpackages, and all 30 crates that own a
+# BUCK file are subpackages -- so the list resolved to the single source under
+# crates/rue-runtime-asan/ (the one crate without a BUCK file). The gate ran
+# rustfmt over 1 file of 281 and reported a pass for the other 280 (RUE-1152).
+#
+# Restoring cache-aware checking here means enumerating sources per crate,
+# which is a real change rather than a glob tweak; until then CI calls
+# `./fmt.sh check`, whose `find`-based discovery covers every source and now
+# fails rather than exits 0 when it finds none.
 
 # The std library sources are runtime inputs to CLI integration tests and spec
 # cases that opt into the real std (compiled programs `@import` them via
--- a/clippy.sh
+++ b/clippy.sh
@@ -41,7 +41,11 @@ echo "Running clippy on ${#CLIPPY_TARGETS[@]} targets..."
 clippy_err="$(mktemp)"
 trap 'rm -f "$clippy_err"' EXIT
 clippy_status=0
-SHOW_OUTPUT="$(./buck2 build "${CLIPPY_TARGETS[@]}" --show-output 2>"$clippy_err")" \
+# `--materializations=all` is load-bearing, not a tuning knob. A remote cache
+# hit satisfies the build without ever writing the diagnostics file to local
+# disk, and the loop below can only read files that exist -- so without this the
+# gate scores cached targets as clean (RUE-1152).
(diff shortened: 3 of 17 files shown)


steveklabnik:
From #2000, which is the PR this surfaced on — one more fail-open path in `clippy.sh` that this PR does not close.

The coverage fix here is right and is a real bug. But there is a **second, independent** way the gate drops a finding it did read, in the detection line itself (line 80 after this PR, unchanged):

```bash
if printf '%s\n' "$CLEANED" | grep -qE '^error'; then
```

`clippy.sh` runs under `set -euo pipefail`. `grep -q` exits the moment it matches, closing the pipe; `printf` then fails with EPIPE, and `pipefail` propagates *printf's* non-zero status as the pipeline's status. The `if` reads false and the finding is discarded.

The tell is inverted, which is what makes it nasty: **the broken-pipe message only appears when grep did match.** From the run on #2000 that reported a pass:

```
Running clippy on 66 targets...
./clippy.sh: line 64: printf: write error: Broken pipe
./clippy.sh: line 64: printf: write error: Broken pipe
./clippy.sh: line 64: printf: write error: Broken pipe
clippy gate passed: no lint violations.
```

Three targets' findings located and dropped, reported as "no lint violations." Whether the race is lost depends on artifact size and scheduling — the same tree passed twice and failed once.

Deterministic repro:

```bash
set -euo pipefail
CLEANED="$(printf 'error: boom\n'; for i in $(seq 1 200000); do echo "filler line $i"; done)"
if printf '%s\n' "$CLEANED" | grep -qE '^error'; then
    echo "DETECTED"
else
    echo "MISSED"      # <- this branch
fi
```

This PR's green run cannot exercise it: with 0 findings `grep` never matches, so `printf` never takes EPIPE. The path stays live for the next violation anyone introduces, and `CHECKED`/`MISSING` do not catch it — coverage is complete, the artifact is read, the finding is just thrown away.

Fix is a one-liner — drop the pipe, e.g. `if grep -qE '^error' <<<"$CLEANED"; then`, or use `case`/`[[ ]]`. Worth folding in here rather than leaving a known fail-open in a PR whose point is that these gates must fail rather than silently skip.

---

steveklabnik:
Dequeued on a `test (linux-x64)` failure that I believe is infrastructure, not this change. Evidence, and what I could not establish.

**The failing test passes on identical content.** This branch is based on `00f7cb24`, so the queue ref `gh-readonly-queue/trunk/pr-2001-00f7cb24` is a fast-forward — the same tree as `2246b692`. That same `test (linux-x64)` job passed on the PR-head run at 05:19–05:28 and failed at 05:44 on byte-identical code.

**The failure has no output.** The only trace anywhere in the 5732-line log is the batch summary:

```
Tests finished: Pass 62. Fail 1. ...
1 TESTS FAILED
  ✗ root//crates/rue-frontend-diff:frontend-diff-test
```

No `✗ Fail:` block, no assertion, no diff — buck2 normally prints a failing test's output. The surrounding batch reported `Cache hits: 100%`, `Commands: 5 (cached: 5)`, `Down: 876MiB`, so nearly everything came from the remote cache.

**It passes locally on this branch:**

```
✓ Pass: root//crates/rue-frontend-diff:frontend-diff-test (50.3s)
frontend token + structural differential: 1443/1443 corpus files and 3 syntax probes agree
```

**No mechanism.** `rue-frontend-diff` compares token and structural frontend output. This PR's only compiler-source change is in `crates/rue-air/src/sema/`, which runs after the frontend, and every edit is a borrow removal or `args.get(0)` → `args.first()` — no control-flow or signature changes.

**What I cannot show.** I have no failure message to analyse, so I cannot prove the negative; I can only show that identical content passed twice and failed once, with no diagnostic. `rerun_failed_jobs` returns 403 for me, so I cannot retry it myself.

Re-queueing is the next step — if it fails again, and especially if it fails with actual output, that changes the picture and I will treat it as real. Worth noting this is the second infrastructure-shaped failure in this queue tonight; the earlier one was `cli.examples::mosaic::main` exceeding a 30 s budget by 9 ms.

---

PR #2006, opened 2026-07-27
RUE-1092: flip the production body path onto the provider host

steveklabnik:
## What this PR is

This is the flip: the body path stops reaching a declaration-wide epoch per body and runs on a provider-backed receiver instead. It merges when the per-body counters go flat and the large-program CI gates come back — not before. It is deliberately one PR rather than a stack, because the receiver work is inert on its own and merging it in pieces reproduces the non-closing-prerequisite pattern that has kept this issue open across 20+ slices.

## Progress

- [x] Declare the body-local overlay and pin the host worklist
- [x] Settle the identity-export direction (memoized consulted endpoints)
- [x] Provider export host: type and nominal identity, `canonical_type_instance`, `body_function_identity`
- [x] The host struct proper: `ProviderBodyHost`, and `SemanticBodyExportHost` on it
- [x] The receiver's own body RIR (`body_rir_ref`)
- [x] Back the `Configuration` rows
- [x] Narrow the host contract's supertrait, with a real guard and a compile-time witness
- [x] Give the inference lift (`type_to_infer_type`) one definition
- [ ] **The file table** — the receiver needs a two-column body-local file table (**blocks everything below**)
- [ ] `InferenceFactSource` on the receiver (first attempt reverted; see below)
- [ ] `BodyAnalysisHost`: the four fact facades
- [ ] `TypeSyntaxHost` + `OrdinaryBodyAnalysisHost` (these land together)
- [ ] Generify the one-body driver over the host
- [ ] Provider-backed body-owner-token resolution
- [ ] The artifact differential: side B runs the engine and publishes a `BodyTransaction`
- [ ] `Rc` → `Arc`; register a real `compiler.body-transaction` evaluator
- [ ] Delete the coordinator closure and `analyze_body_query`
- [ ] Fix the two `EpochFacts` adapters that over-bind; end-state cleanup (`#[allow(dead_code)]`, `NeedsProviderWiring`)
- [ ] RUE-1090 counters flat; large-program gates restored; `RUE_CALDERA_SUCCESS_STUB` deleted

## The file table is the blocking problem

Two rounds of adversarial review landed here, and the second round refuted the fix proposed by the first. What follows is what is actually true, verified against source.

### There are no real compiler file ids on the provider path

`lower_owned_body_input` (`revisioned_query_database.rs:2056`) is the only production construction of a `BodyRirBundle` (`:2128`). It takes `OwnedBodyInput` — two text terminals, deliberately no module source, declaration slice, or live interner — concatenates them, and re-parses via `SourceSnapshot::single("<rue-body-input>", …)`, which stamps `FileId::DEFAULT` (`source_snapshot.rs:76-82`).

So **every span the engine sees on the provider path carries `FileId(0)`**, and that file is synthetic — it does not exist in the merged program. Real file ids are allocated `1..=snapshot.len()`; real file 0 exists only via `SourceSnapshot::single`.

This kills the fix proposed in the previous revision of this section ("hand the pool the request's real file ids"). It would have set `StructDef.file_id` to a module's real id while every `span.file_id` stayed 0 — strictly worse than the current consult-order numbering, which at least gives the body's own module a consistent handle.

### What `file_id` actually is

Not one thing. Reachable from the provider path it serves five jobs, and two of them need *different* strings:

| Job | Needs |
| --- | --- |
| symbol mangling (`intern_pool.rs:1569` → `struct_symbol_name`) | the **logical** module path |
| pool-internal nominal dedup (`struct_by_file_name`) | any injective numbering |
| engine-facing by-file lookup (`generate.rs:2462, 2468`) | the **engine's** numbering |
| visibility domain (`is_accessible`, `aggregate_resolution.rs:328`) | the **physical** file path |
| escaping dependency coordinate (`typeck.rs:2796` → `DeclarationTypeDependencyEvent.target_file`) | a coordinate the compiler can rejoin |

The epoch keeps these straight with *two* tables under one key — `symbol_paths` (logical) and `file_paths` (physical) — seeded as a validated pair by `SemaMetadata`, which enforces identical key sets.

`BodyIdentityPool::file_for_module` (`body_identity.rs:1187`) mints handles into a scope it does not own and publishes only the logical column (`set_symbol_paths`, `:1194`). That is precisely why symbol mangling is correct today and visibility is not.

### Visibility currently fails open

`is_accessible` resolves *both* file ids through `facts.file_path(..)`, and `SemanticVisibilityDomain::is_visible_from` returns `true` when either domain is `None` (`semantic_type_resolution.rs:23-26`). A pool-minted `def.file_id` misses `ProviderAggregateFacts::file_paths`, which is keyed on caller-registered ids — so the defining domain is `None` and **every private item becomes visible**. E0460 and E0706 stop firing.

Roughly ten call sites read `def.file_id` straight off the pool and hand it to this: `aggregates.rs:98, 380, 415, 950, 990, 1050, 1103, 1377`, `control_flow.rs:1049`, `visibility.rs:129`. It is accept-invalid, silent, and no existing test catches it — there is no negative visibility test against the provider receiver.

Not yet a live defect: nothing selects the receiver at runtime. It is a design gap that must close before it can be.

### The shape of the fix, and what is still open

The body-local numbering stays — there is nothing else available, since the receiver's own RIR is a synthetic re-parse. What changes is that the table becomes **two-column and total**: every id the receiver can produce (its own `FileId(0)`, plus one per consulted module) must resolve to both a logical module path and a physical file path, so `file_path()` answers for exactly the ids `struct_def().file_id` can return.

Three things follow, and one is unresolved:

1. `file_for_module` should stop republishing the whole map on every newly-seen module (`body_identity.rs:1194-1199`) — that is O(m²) per body today, a cost the fix removes rather than adds.
2. A durable nominal whose module is absent from the table is a stale input. It should fail closed, not receive a synthetic id: there is no correct synthetic id, because every consumer above either misreads it or fails open on it.
3. **Open:** `DurableNominal` carries only `module_path`, the logical path (`body_identity.rs:127`). Visibility needs the physical one, and it is not on the durable side. Either the durable nominal grows a physical path, or the visibility domain stops being derived from a filesystem directory. That is a semantics question, not a plumbing one, and it is the next thing to settle.

The receiver's `pool_module_path` and the identity seam are unaffected: they were already keyed on logical module paths, deliberately (see `ConsultedDefinition`, which drops the file number on purpose). Digests take only strings and do not observe any of this.

## What the reverted `InferenceFactSource` attempt got wrong

Recorded because the failure mode generalizes. Six of fourteen arms:

| Arm | Defect |
| --- | --- |
| four const arms + `inference_module_binding_type` | converted engine file ids through a pool-keyed table |
| `inference_struct_type_by_file`, `inference_enum_type_by_file` | same mismatch, no conversion at all |
| `uncached_function_sig`, `inference_function_by_file` | wired to `ProviderEndpointFacts`, which returns `None` for both **by design** — the answers live on `ProviderCallFacts`. Every function signature would have been absent |
| `inference_builtin_struct_type`, `inference_builtin_enum_type` | live pool read at `FileId::DEFAULT`, where `Str(N)`, anonymous structs and generated slices also live. The epoch reads a snapshot frozen before the body runs, so a mid-body mint cannot retroactively change an earlier lookup |

Two lessons carried forward:

1. **The tests could not have caught it.** Both installed symbol paths by hand with `set_symbol_paths` instead of letting `file_for_module` assign them, so neither exercised the mismatch — and one asserted the silent-absent behavior *as correct*. Same shape as `15088b8`, which survived a green CI run because its tests drove a stub that accepted whatever number it was handed.
2. **`Sema::builtin_structs` is never written anywhere in the tree.** Declared, initialized empty, read. The epoch's builtin-struct answer comes entirely from the frozen overlay, so any provider implementation reading a live table diverges by construction.

## The worklist, as executable data

`HOST_CAPABILITY_LEDGER` carries one row per method on `OrdinaryBodyAnalysisHost`, stating where its answer comes from. It parses the real trait source and asserts equality in both directions.

| Category | Count |
| --- | ---: |
| `BodyLocal` | 35 |
| `ProviderFact` | 14 |
| `ProviderState` | 8 |
| `Configuration` | 6 |
| `ProviderExport` | 5 |
| `InapplicableToBodies` | 3 |
| `BodyRir` | 1 |
| `NeedsProviderWiring` | 1 |

Known gap: it parses only methods declared directly in the trait body, so supertrait obligations are invisible to it.

Its end state: the row set should survive the flip — once `Sema` is gone it stops being a worklist and becomes the thing that fails the build if someone answers a contract method from a declaration-wide table again. The migration-only vocabulary (`NeedsProviderWiring`) should not survive.

## The supertrait narrowing

`BodyAnalysisHost` required `BodyAnalysisReadHost` — four epoch-shaped fact sources, 61 point queries. Three are never read off `Self`; hosts supply endpoint, call and aggregate facts through the three associated types. The supertrait is now `InferenceFactSource` alone.

Review could not break the safety of that and independently confirmed `HostInferenceFacts` touches only `InferenceFactSource`. It corrected two things:

- **The stated reason was wrong.** Inference also reaches the engine through an associated type; `Sema` merely chooses `HostInferenceFacts<'a, Self>`. By the rule the commit gave, `InferenceFactSource` had no more claim than the three dropped.
- **The 47-method obligation lives elsewhere.** `EpochFacts` in `call_resolution.rs:215/219/225` and `aggregate_resolution.rs:113/117/123` bind on all four sources while calling one — unlike `body_endpoint.rs:509/513/519`, which correctly binds on one. Fixing those six sites is what would let `BodyAnalysisReadHost` be deleted.

The guard has since been fixed: it was a prefix match that still passed for `: InferenceFactSource + BodyAnalysisReadHost + Sized`, which is how a bound actually gets widened. It now compares the whole supertrait list, and `MinimalHost` — implementing the narrow source and nothing else — is a compile-time witness, since `Sema` satisfying all four proved nothing.

## The receiver

`ProviderBodyHost` owns the body-local overlay, this body's `BodyRirBundle`, `ProviderBodyAnalysisState`, one fact boundary, and `ProviderBodyConfiguration`. None is sized by the declaration universe, so constructing a receiver is `O(1)` in the program rather than `O(declarations)`. That is where the `O(bodies × declarations)` term goes: the epoch's cost was in *building* the receiver.

One caveat review raised on the "single fact authority" claim: `S`, the durable nominal/const source, is a separate generic parameter from `P`, the fact boundary. The boundary answers *which* declaration; `S` answers *what it is*. Whether `S`'s answers carry edges is a property of whoever constructs it, enforced by nothing in the type system. Not yet a defect — the only implementations are test-only — but the invariant as written is stronger than the types guarantee.

## Configuration is mostly derived, not supplied

Only the target and preview feature set are genuine query keys. `known_symbols` is interned from the receiver's own interner; the three builtin enum identities are read from its own type pool, which pre-registers `rue_builtins::BUILTIN_ENUMS` at `FileId::DEFAULT` as a fresh import epoch does. `require_preview` reproduces the epoch's gate verbatim.

## Failure stays closed

The host deliberately does **not** inherit the epoch's synthetic-test fallback in `stable_definition_token`, which mints a token from an FNV hash when its table is empty. Unmapped callable symbols and unresolvable module identities fail typed. The visibility gap above is the one place this discipline is currently not met, and closing it is part of the file-table work.

## Scope discipline

Nothing here is selected at runtime; `analyze_body_query` remains the single authority until the commit that deletes it. That is why the reverted defects shipped no wrong artifact — but a wrong *rationale* in a module whose discipline is that its stated fact sources can be trusted is worth reverting rather than patching.

## Known-red CI, and why it is not this branch

`cli.examples_mosaic::mosaic_builds_site_and_shared_artifacts` intermittently times out at the `heavyweight` 30 s budget on aarch64. It has failed on `linux-arm64-cli-shard-1` and on `macos-cli-shard-2`, and passed on both — on `bfa485e` the arm64 shard passed and macOS failed instead. A deterministic regression does not move between runners that way.

"Nothing here is reachable" stopped being sufficient once this branch contained a trait-bound narrowing and a helper de-duplication the epoch path calls, so it was measured. Same container, same build path, `rue examples/mosaic/main.rue`, three runs each:

| | run 1 | run 2 | run 3 | median | fastest |
| --- | ---: | ---: | ---: | ---: | ---: |
| `trunk` @ `3ef6186` | 38431 ms | 38466 ms | 38632 ms | 38466 ms | 38431 ms |
| head @ `39ed874` | 39284 ms | 37953 ms | 38695 ms | 38695 ms | **37953 ms** |

+229 ms of median (+0.6%), head's fastest run beating trunk's fastest, and head's own 1331 ms spread enveloping all three trunk samples. No signal at n=3. Only the base-vs-head comparison is meaningful; both sides were measured back to back on the same machine.

Raising the budget remains the one fix that will not be made: the x86-64/aarch64 gap on this example *is* the regression this PR exists to delete.

## Validation

- `scripts/rue unit air` — 676 passed, 0 failed
- `scripts/rue quick` — 30 pass, 0 fail
- `scripts/rue fmt` clean

### DIFF
--- a/crates/rue-air/src/intern_pool.rs
+++ b/crates/rue-air/src/intern_pool.rs
@@ -1729,6 +1729,21 @@ impl TypeInternPool {
         }
     }
 
+    /// The relocation-stable logical identity of one defining source file.
+    ///
+    /// Keyed and exact: a caller that has a file in hand asks for that file,
+    /// never for the table. The provider-backed body receiver renders stable
+    /// symbol components through this, and its table holds only the modules
+    /// that body consulted.
+    pub(crate) fn symbol_path(&self, file: FileId) -> Option<String> {
+        self.inner
+            .read()
+            .unwrap_or_else(PoisonError::into_inner)
+            .symbol_paths
+            .get(&file)
+            .cloned()
+    }
+
     /// Set relocation-stable source identities for type-derived symbols.
     pub(crate) fn set_symbol_paths(&self, symbol_paths: HashMap<FileId, String>) {
         self.inner
--- a/crates/rue-air/src/lib.rs
+++ b/crates/rue-air/src/lib.rs
@@ -105,6 +105,16 @@ pub use sema::{
     SemanticProducedAnonymousNominal, SemanticProducedAnonymousNominalShape, SourceParamAbi,
     SpecializedFreeFunctionDependencyEvent, SpecializedFreeFunctionOrigin,
 };
+
+/// The provider-backed body receiver the RUE-1092 flip installs.
+///
+/// Exported so the rFinal differential can build one over the real fact
+/// boundary and compare it against the epoch receiver. Nothing selects it at
+/// runtime yet; `analyze_body_query` remains the single body path.
+pub use crate::sema::provider_host::{
+    BodyHostFacts, ExportCallableFacts, ExportedMethodSite, ProviderBodyConfiguration,
+    ProviderBodyHost, StableIdentityFacts,
+};
 pub use semantic_body::{
     SEMANTIC_BODY_INST_KINDS, SemanticAnonymousBodyExport, SemanticBody, SemanticBodyAnchor,
     SemanticBodyCallArg, SemanticBodyCandidate, SemanticBodyCandidateInstallWork,
--- a/crates/rue-air/src/sema/body_identity.rs
+++ b/crates/rue-air/src/sema/body_identity.rs
@@ -2161,6 +2161,16 @@ impl BodyRirBundle {
             .collect()
     }
 
+    /// The body's own lowered RIR.
+    ///
+    /// Borrowed from the bundle rather than through [`Self::view`] so the
+    /// borrow outlives one expression: the ordinary-body engine holds a `&Rir`
+    /// across a whole body evaluation, and a view materialized per call cannot
+    /// lend one.
+    pub(in crate::sema) fn rir(&self) -> &Rir {
+        &self.rir
+    }
+
     /// Borrow the one local RIR/index/interner authority for provider fact
     /// facades. The view is cheap and carries no independent identity state.
     pub fn view(&self) -> BodyRirView<'_> {
--- a/crates/rue-air/src/sema/consistency_tests.rs
+++ b/crates/rue-air/src/sema/consistency_tests.rs
@@ -1413,12 +1413,28 @@ mod tests {
     /// consults the epoch-shaped read sources. Widening this back to
     /// `BodyAnalysisReadHost` would silently oblige every host — including the
     /// provider-backed receiver — to implement 47 point queries nothing calls.
+    ///
+    /// The assertion compares the WHOLE supertrait list, not a prefix. A prefix
+    /// match passes for `: InferenceFactSource + BodyAnalysisReadHost + Sized`,
+    /// which is how a bound actually gets widened again — by appending, not by
+    /// replacing. `the_narrow_host_contract_has_a_witness` below is the other
+    /// half: a string test cannot prove a 14-method type satisfies the contract,
+    /// only that the source says so.
     #[test]
     fn the_host_contract_requires_only_the_inference_source() {
-        assert!(
-            FACT_MODE_SOURCE.contains("pub(crate) trait BodyAnalysisHost: InferenceFactSource"),
-            "the host contract's supertrait must stay narrowed to the one source \
-             `HostInferenceFacts` actually borrows the host for"
+        let declaration = FACT_MODE_SOURCE
+            .lines()
+            .find(|line| line.starts_with("pub(crate) trait BodyAnalysisHost:"))
+            .expect("the host contract is declared on one line");
+        let bounds = declaration
+            .trim_start_matches("pub(crate) trait BodyAnalysisHost:")
+            .trim_end_matches('{')
+            .trim();
+        assert_eq!(
+            bounds, "InferenceFactSource + Sized",
+            "the host contract's supertrait list must stay exactly the one source \
+             `HostInferenceFacts` borrows the host for, plus `Sized`; appending \
+             another source silently re-imposes its point queries on every host"
         );
     }
 
--- a/crates/rue-air/src/sema/consistency_tests.rs
+++ b/crates/rue-air/src/sema/consistency_tests.rs
@@ -1406,6 +1406,22 @@ mod tests {
         assert!(!CONTROL_FLOW_SOURCE.contains("struct BodySema"));
     }
 
+    /// The host contract requires the inference source and nothing wider.
+    ///
+    /// Endpoint, call, and aggregate facts reach the engine through the three
+    /// associated types, so a host that answers them with its own facades never
+    /// consults the epoch-shaped read sources. Widening this back to
+    /// `BodyAnalysisReadHost` would silently oblige every host — including the
+    /// provider-backed receiver — to implement 47 point queries nothing calls.
+    #[test]
+    fn the_host_contract_requires_only_the_inference_source() {
+        assert!(
+            FACT_MODE_SOURCE.contains("pub(crate) trait BodyAnalysisHost: InferenceFactSource"),
+            "the host contract's supertrait must stay narrowed to the one source \
+             `HostInferenceFacts` actually borrows the host for"
+        );
+    }
+
     #[test]
     fn one_body_host_contract_is_representation_agnostic() {
         assert!(FACT_MODE_SOURCE.contains("trait BodyAnalysisHost"));
@@ -1481,8 +1497,8 @@ mod tests {
             ),
             (
                 INFERENCE_CONTEXT_SOURCE,
-                "pub(crate) struct HostInferenceFacts<'a, H: BodyAnalysisReadHost>",
-                "impl<H: BodyAnalysisReadHost> LazyInferenceFacts for HostInferenceFacts<'_, H>",
+                "pub(crate) struct HostInferenceFacts<'a, H: InferenceFactSource>",
+                "impl<H: InferenceFactSource> LazyInferenceFacts for HostInferenceFacts<'_, H>",
             ),
         ];
         for (source, struct_header, impl_header) in adapters {
--- a/crates/rue-air/src/sema/fact_mode.rs
+++ b/crates/rue-air/src/sema/fact_mode.rs
@@ -85,7 +85,17 @@ impl<T> BodyAnalysisReadHost for T where
 /// Associated fact facades return owned or short-lived read-only values; the
 /// mutable operations take a request object so the contract stays independent
 /// of the current analyzer representation.
-pub(crate) trait BodyAnalysisHost: BodyAnalysisReadHost + Sized {
+///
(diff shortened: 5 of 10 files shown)


steveklabnik:
`test (linux-arm64-cli-shard-1)` is red on `a508f25` with two `cli.examples_mosaic` timeouts:

```
TIMEOUT: compiler phase timed out (elapsed=30011 ms, budget=30000 ms)
TIMEOUT: compiler phase timed out (elapsed=30019 ms, budget=30000 ms)
```

Not caused by this branch, and not a rerun-and-it-passes flake. Two facts:

1. The branch's entire diff against its merge base is `crates/rue-air/src/sema/provider_host.rs` (new, `#[allow(dead_code)]`, nothing selects it at runtime) plus one `mod` line. There is no code path here that a compile can reach.
2. Mosaic compiles in **23.9s** on an x86-64 container against the `heavyweight` contract's 30s budget. The arm64 runner is slower, so the same compile lands just past the wall — which is why both cases stop at ~30.01s rather than failing on output.

This is the RUE-1092 regression sitting on the CI budget boundary: ~85% of large-program wall time is repeated per-body prepare/project/install, and Mosaic is a maintained large example on the pre-flip path. Raising the budget would hide the thing this PR exists to delete, so I'm not doing that. It clears when the flip lands and the per-body counters go flat — the same milestone that un-stubs `RUE_CALDERA_SUCCESS_STUB`.

Progress this round: `ProviderBodyHost` now exists — the receiver tying the body-local overlay, `ProviderBodyAnalysisState`, and one `BodyFactProvider` together — and implements `SemanticBodyExportHost`, so publication reaches the provider-backed export path through the same neutral contract it reaches the epoch through. That closes items 1 and 2 of the worklist in the PR body.

---

steveklabnik:
`test (macos-cli-shard-2)` failed on `bfa485e`. It is the known-red heavyweight-budget case, not a regression from this commit, and I am deliberately not "fixing" it.

```
---- cli.examples_mosaic::mosaic_builds_site_and_shared_artifacts ----
TIMEOUT: compiler phase timed out (elapsed=30016 ms, budget=30000 ms; process group killed)
```

16 ms over a 30 s budget. Two things worth recording:

- **It moved runners.** `test (linux-arm64-cli-shard-1)` — the arm64 job this PR has been carrying as known-red — **passed** on this head. The same Mosaic case failed on macOS instead. Both are aarch64. So this is not a linux-arm64 property; it is the general aarch64 boundary, and which job shows it depends on which runner is slowest that run. I have corrected the PR description accordingly.
- **It is not this commit.** `bfa485e` adds configuration state to `ProviderBodyHost`, which lives behind `#[allow(dead_code)]` and is selected by nothing at runtime — `analyze_body_query` is still the only body path. There is no mechanism by which it changes Mosaic's compile time.

Everything else on this head is green, including `clippy`, `fmt`, both `spec` jobs, `compiler reproducibility`, `scaling matrix`, `asan`, and `valgrind`. The other three cancelled jobs are fail-fast casualties of this one.

Raising the budget is the one fix I will not make. The gap between 23.9 s on x86-64 and >30 s on aarch64 *is* the `O(bodies × declarations)` regression this PR exists to delete; a budget bump would hide the signal that tells us whether the flip worked. It clears when the flip lands, and the flat-counter gate is what this PR merges on.

---

PR #2125, opened 2026-08-07
RUE-1243: define warm edit measurement contract

steveklabnik:
Fixes RUE-1243.

## Summary

- accept ADR-0068 as the retained-session edit-performance measurement contract
- keep warm measurements separate from ADR-0067's fresh-build history and headline
- require one canonical filesystem-backed host shared by the CLI, benchmark runner, and watch mode
- measure Mosaic and Lattice through no-op, unreachable, body, signature, layout, import, reachability, and error-introduction edits
- record diagnostics-ready, codegen-ready, objects-ready, and runnable-ready endpoints with structural work and retention gauges
- require both `-j1` and production automatic-worker rows on a manifest-pinned reference host
- publish warm/fresh divergences as retained failing artifacts
- advance incremental-linker design only when the link band is both material in share and at least 10 ms, exceeds 20 ms, or demonstrably blocks the agreed interactive objective

## Review changes

The accepted revision incorporates every substantive point from Fable's review: the linker gate now has an absolute floor, oracle mismatches fail loudly, diagnostic latency is first-class, worker modes are pinned, the reference host is manifest-owned, and the scaling criterion describes a bounded edit against growing maintained programs.

## Validation

- `scripts/validate-adrs.py`
- `scripts/rue quick` (30 targets, 0 failures; 808 compiler tests passed)

### DIFF
--- a/docs/designs/0068-incremental-edit-performance-measurement.md
+++ b/docs/designs/0068-incremental-edit-performance-measurement.md
@@ -1,0 +1,383 @@
+---
+id: 0068
+title: "Incremental edit-scenario performance measurement"
+status: accepted
+tags: [tooling, compiler, incremental, performance]
+feature-flag: null
+created: 2026-08-07
+accepted: 2026-08-07
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["ADR-0063", "ADR-0067", "RUE-1033", "RUE-1086", "RUE-1096", "RUE-1200", "RUE-1242", "RUE-1243"]
+---
+
+# ADR-0068: Incremental edit-scenario performance measurement
+
+## Status
+
+Accepted. ADR-0063 implemented the retained query graph through per-function
+object projection and a fresh link. ADR-0067 deliberately limits its historical
+performance system to fresh compiler processes and lists incremental and edit-
+scenario performance as future work requiring a separate ADR. This record
+defines that separate regime; it does not change either predecessor's results or
+terminology.
+
+## Summary
+
+Rue will measure interactive compiler performance as revision sequences through
+one retained `CompilerSession`, using the canonical filesystem/import-discovery
+host path and maintained Rue programs. Every scenario prepares an untimed
+baseline, applies one declared successor edit, and records the latency and exact
+structural work needed to reach three successful-compilation endpoints:
+collected `CodegenUnit`s, retained object projections, and a freshly linked
+executable. An expected erroneous successor instead records the time to its
+canonical diagnostics.
+
+Warm observations remain a separate lower-frequency report, never part of
+ADR-0067's fresh-build headline or baseline history. Correctness is established
+by comparing the warm result with a fresh session for the successor source.
+Structural work is the architectural gate; host-specific latency decides where
+to investigate, not whether compilation is correct.
+
+## Context
+
+ADR-0063's Phase 12 witness proves that a body-only edit can recompute one body,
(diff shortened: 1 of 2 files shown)


steveklabnik:
Review of the ADR-0068 proposal, as requested. (I'm Fable, reviewing at Steve's ask.)

Overall this is strong: the two-regime separation is clean, "structural work is the gate, latency is the pointer" is the right epistemology, the prohibition on benchmark-only compiler paths matches the one-canonical-path rule, and the baseline-overlay mechanism plus separate schema/fixture revisions are well thought out. Four substantive points, in priority order:

**1. The linker gate's 20% criterion is nearly auto-satisfied — it won't do the prioritization work §7 claims.** The Phase 12 witness already makes the case: 300µs to codegen-ready, 1,618µs to runnable — the fresh link is ~80% of the warm interval. That's the expected shape for any body-only edit: incremental locality shrinks the compute term toward zero while fresh-link work scales with whole-program size, so link *fraction* trends high on every decent fixture even when absolute link time is interactively imperceptible. Since the gate is an either/or, the 20%-of-interval rule will trigger on Lattice almost regardless of whether a stateful linker is worth the complexity — a 5ms total with a 4ms link crosses the gate despite being far below anything a user can feel. Suggest either conjoining the relative criterion with an absolute floor (≥20% *and* ≥N ms), or dropping it and gating purely on the 20ms/interactive-objective criteria.

**2. A failed fresh oracle deserves loud publication, not just an invalid run.** §5 lumps "a failed fresh oracle" in with sample shortfalls and malformed edits as "invalid and publishes no partial report." But a warm/fresh divergence is the single most valuable thing this harness can ever find — a retained-session correctness bug (the RUE-1193/RUE-1236 class) — and under the current text its only trace is a report that silently doesn't appear. Suggest splitting the invalid-run taxonomy: infrastructure invalidity suppresses the report; an oracle mismatch publishes a failing artifact with divergence details and expects an issue filed. Different events, opposite desired visibility.

**3. All seven scenario classes are valid→valid edits; interactive editing largely isn't.** The most user-felt warm latency is arguably edit→first-diagnostic on a broken revision, and the matrix never times it — errors appear only in §6 (as correctness assertions) and Phase 6 watch tests (product tests, untimed). Either add an error-introduction scenario class with a diagnostics-ready endpoint, or spend one sentence in §2's exclusions explicitly deferring diagnostic latency so the omission reads as a decision rather than a gap.

**4. The matrix doesn't pin worker counts.** The raw contract records query-worker count, but no row declares which `-j` values constitute the suite. Warm edit latency at high `-j` is exactly where the current lock-convoy work lives (RUE-1235/1238/1239, and RUE-1241's `QueryFamily::node` convoy is specifically a warm-path memo-index phenomenon) — a suite that happens to run single-threaded would be blind to regressions in precisely the dimension being actively optimized. Suggest declaring at least `-j1` and the default per row, mirroring RUE-1241's own acceptance methodology.

Two smaller notes:

- The 20ms threshold is tied to "the recorded reference host," but nothing says where the reference host is *declared*. Pinning it in the versioned manifest keeps the gate's meaning stable when hardware changes.
- The third gate criterion's phrasing "fresh-link work grows materially with unchanged program size" reads backwards — presumably it means link work scales with program size even when the *edit* is small. Worth rewording; this is the criterion someone will argue over later.

The §5 cost consequence (5 sessions × 7 scenarios × 2 workloads, each paying an untimed Lattice baseline) is honestly acknowledged and I think correctly resolved in favor of sample independence — no change suggested.

---

PR #2137, opened 2026-08-07
Delete write-only dependency projections; document the parse shim's edge contract

steveklabnik:
Fixes RUE-1204. Fixes RUE-1145.

Two commits, one theme: the dependency graph should record only what production actually reads.

## Commit 1 — RUE-1145: document the parse shim's key-identified dependency contract

`execute_parse_query` runs the per-module parses *outside* the outer `compiler.parse` body and records one synthetic whole-source leaf, so the outer node's edge set never reflects the modules it consumed. The issue offered two resolutions; this takes the documentation option, and the code itself argues for it:

- the node is **key-identified by construction** — `ExactSourceInput` embeds every file's exact content identity, so a source change selects a different node and recorded edges never drive its invalidation;
- the **successor path already records real edges** (predecessor-terminal adoption plus appended-module input leaves), so the fiction is confined to the ordinary path;
- recording real module edges would require growing an out-of-band observation API in rue-query (`observe_adopted_terminal` is deliberately content-addressed-only) for a shim node that ADR-0063 Phase 12 (RUE-1033) deletes.

Both sites now carry a deletion-gate comment stating the contract: no consumer may read this node's edges.

## Commit 2 — RUE-1204: delete the write-only function dependency projections

The manifest has derived free-function and method edges solely from the query-owned body-references channel since RUE-1027/RUE-1209, but semantic analysis kept producing the observer projections. A full-workspace audit found **zero readers** — the only reference outside their own production was a `Send + Sync` assertion in a test. Removed:

- `SemaOutput` fields: `ordinary_free_function_dependencies` (+`_complete`), `specialized_free_function_origins`, `specialized_free_function_dependencies` (+`_complete`), `named_method_dependencies` (+ both method completeness flags), `named_destructor_dependencies` (+`_complete`);
- types: `OrdinaryFreeFunctionDependencyEvent`, `SpecializedFreeFunctionOrigin`, `SpecializedFreeFunctionDependencyEvent`, `NamedMethodDependencyTargetEvent`, `NamedMethodDependencyEvent`, `NamedDestructorDependencyEvent`;
- the two dependency-event builder helpers and their four call sites in `analyze_function_bodies_lazy`, the push loops in body analysis and specialization, and the baseline snapshot/restore entries;
- five write-only `BodyAnalysisWork` counters and their unstable-projection serializations (no fixture or test asserts those JSON keys).

Deliberately untouched (verified live): `declaration_type_dependencies` / call-head channels (read by `binding_manifest`), `named_const_dependencies` (event type constructed in `canonical_semantic`), `body_named_dependencies`, `analyzed_body_owners`, and `ImplicitNamedDestructorDependencyEvent`.

Net −526/+34.

## Testing

- `scripts/rue quick` — 31/31 targets
- `scripts/rue premerge` — full tier passed (79 targets, clippy gate included)
- `scripts/rue fmt`

---
_Generated by [Claude Code](https://claude.ai/code/session_01DrKqpQgkHzDHGqpEMjxYdN)_


steveklabnik:
The merge-queue ejection was `rue-compiler-differential-oracle-test` failing in the queue run — the RUE-1244 intermittent (false layout cycles in the differential oracle), not this diff: the PR-head run passed the same suite, this change deletes only reader-free projection code, and the suite passes 7/7 locally on the exact merge result the queue tested (this branch merged with trunk `b75856ab`). Pushed the trunk merge and re-enabled auto-merge to re-queue.

---

PR #2140, opened 2026-08-08
RUE-1258: fail CI when the performance series stalls

DorianListens:
Refusing an observation protects comparability, but a refusal nobody sees stops the series while every job reports success. Measurement that has quietly ceased is worse than no measurement, because it is still believed — the last stall ran ten days and hid RUE-1257, a 3.7x–7.6x compile-time regression that the flagging rule had detected correctly on 50 of 102 points.

Closes RUE-1258.

## Two gates

Both live in `linux-premerge`, which already has a built compiler and runner; a dedicated job would pay for a second compiler build to answer two cheap questions.

**"Would this change stall the series?"** — `rue-bench check-pins`, decided from the pull request's tree alone. Every pin that can still refuse a run after RUE-1256 is a content hash of a file in that tree, so no measurement and no data branch are needed. Resolution goes through `pins`, the same module the collector uses, so the gate and collection cannot drift apart on what a pin means.

**"Is the series already stalled?"** — `scripts/validate-performance-stall.py`, defined as more than five merged trunk commits with no new plotted point on some platform. This fails *every* pull request, not only one that touches performance, because a stall is a repository-wide condition rather than a property of whichever change caused it. Series state is derived from the raw records rather than read from a stored flag, since ADR-0067 keeps everything derived off the data branch and a stored staleness flag is exactly the kind of value that drifts from its records and gets believed anyway.

## No escape hatch

Deliberate, and the reasoning is recorded on the issue and in ADR-0067. The remedy for either gate is declaring the next epoch, which needs no baseline (`PlatformEpoch.baseline` is `Option`, documented as "an epoch may exist and collect observations before it can express them as an index") and is therefore a small manifest change. Both failure messages print that remedy rather than only reporting that a hash moved.

A genuine emergency is served by bypassing branch protection, which is already maintainer-only, visible on the pull request, and uncomfortable enough not to become routine. A bypass label would have been a *cheaper* path than the emergency it existed for.

## Supporting changes

- **An epoch declares whether collection measures it** (`collection = true`). The collector drops its hardcoded `--epoch 2` and follows the manifest, so declaring the next epoch under time pressure is one file rather than two that can disagree. Parsing rejects two collection epochs on one platform.
- **A run that cannot enter its series fails the collection workflow** instead of warning — after its artifact uploads, so the evidence survives and the raw record still reaches the data branch. The old `::warning::` also claimed the run "will not be published" when it was published and then dropped at derivation.
- **The dashboard states its newest measurement's date**, and reports rejected runs in the health line rather than only behind a collapsed disclosure. A frozen page was previously indistinguishable from a current one.
- The CI contract (`validate-ci-gate.py`) now proves both gates and the `fetch-depth: 0` they need remain in `linux-premerge`, so they cannot be quietly dropped.

## Verification

Both gates were exercised against real data, in both directions.

`check-pins` on the current tree, and with `rust-toolchain.toml` edited:

```
rue-bench: 3 collection epoch(s) still match the tree            exit 0

rue-bench check-pins: 3 pins have drifted from the tree:
  x86_64-linux epoch 2 — toolchain_hash
    declared  30f50af0ac71...
    resolved  1b35b7bda8fd...
    from      rust-toolchain.toml                                exit 3
```

The staleness gate against the live `performance-data-v1`, then against the same data rewound to the point the series was actually frozen at:

```
aarch64-linux: current at 80c1944c388d (2026-08-08T00:13:26Z)    exit 0

The published performance series has stopped advancing.
  aarch64-linux: newest plotted point is 5859c3b82d15,
                 95 trunk commits behind (threshold 5)           exit 1
```

That second case is the real July stall replayed through the gate.

Also run: 104 `rue-perf-schema` tests, 70 `rue-bench` tests (7 new for `check-pins`), 7 new staleness-rule tests, the CI-gate validator and its 14 tool tests, the clippy gate across all 65 first-party targets, `scripts/rue fmt`, `scripts/validate-adrs.py`, and `scripts/validate-shell-pipefail-pipelines.py`.

## For review

The threshold of five commits is a judgement call — low enough to catch a real stall within hours at the current merge rate, high enough that one failed collection or collection merely lagging a merge never blocks the repository. It is a one-line change in `scripts/validate-performance-stall.py` if it proves wrong in either direction.

The design accepts that it has no graceful degradation: if a stall happens and the remedy is somehow not the small change argued above, the fallback is an admin merging pull requests by hand. That is the accepted cost of making a stall impossible to absorb.

Refs RUE-1256, RUE-1257.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MgbZ4wpBwZKj6GhsT52YAw)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,6 +164,12 @@ jobs:
     name: premerge (linux-x64)
     steps:
       - uses: actions/checkout@v6
+        with:
+          # The performance-staleness gate counts trunk commits merged since the
+          # newest plotted measurement, so it needs the measured commit in local
+          # history. The default depth-1 checkout has none of them, and a gate
+          # that cannot see the history fails rather than passing. (RUE-1258)
+          fetch-depth: 0
 
       - name: Install dotslash
         uses: facebook/install-dotslash@v2  # pinned: the moving `latest` branch broke macOS runners (sha256sum flags) on 2026-06-18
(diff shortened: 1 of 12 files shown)


DorianListens:
@steveklabnik this is fairly aggressive, so I'm definitely open to taking a different approach here if you'd prefer. That said, breaking the perf dashboard caused us to miss a 7x perf regression for like two weeks, which is sub-optimal, and without something like this, we're liable to do the same thing again in the future without realizing it.

PR #2143, opened 2026-08-09
Complete the performance dashboard's §11 tooltip contract

DorianListens:
Fixes RUE-1194.

ADR-0067 §11 requires the tooltip to carry the hovered band and its value, and commits-since-last-measurement when runs were skipped. Neither was implemented, and the comment introducing `drawTooltip` paraphrased §11 with both items quietly dropped rather than deferred. Along the way the charts turned out not to work on touch at all, the tooltip could not be pinned or followed to a commit, and review surfaced several more problems.

§11 is now implemented in full. Screenshots are rendered from a synthetic fixture, not from collected data — the series has not started publishing yet.

## The two §11 items RUE-1194 named

Phase chart tooltip

**Hovered band and value.** This needs the cursor's height, not just its position, so the stack is retained per column and hit-tested. The composition bar outlines the same band and bolds its legend entry, so the chart, the bar and the tooltip always agree on what is being pointed at.

RUE-1194 asked whether the composition bar already satisfies this criterion. It does not: the bar answers what a commit consists of, and only the tooltip can answer what the cursor is on. Both now say it.

**Commits-since-last-measurement.** Resolved at site build time next to the commit subject, as a position on trunk's **first-parent** line. That detail is load-bearing — counting ancestors would report a merged three-commit topic branch as three skipped trunk measurements. A commit that cannot be placed on that line gets no ordinal and the tooltip omits the line rather than guessing, the same degradation the subject lookup already had.

Headline chart tooltip

## Pinning and the commit link

The same §11 sentence ends *"Clicking pins the tooltip and links to the commit."* Both halves belong together, because neither works alone: with a mouse, travelling to the tooltip to read it — or to its link to follow it — drags the cursor back across the chart and rewrites the tooltip on the way.

Clicking either chart pins it; clicking the pinned point again, or Escape, releases it. A pinned chart ignores hover but still follows a deliberate drag, because on a touchscreen every tap pins — there is no hover for a pin to suppress — and without that exception pinning would break scrubbing on exactly the devices this PR taught the charts to serve. The cursor rule goes solid and the index mark fills when held, so the state is visible on the chart and not only in the tooltip's text.

Pinned tooltip with a commit link

The commit URL comes from the repository's own remote, resolved at build time beside the subject and the ordinal. It is **rebuilt** from the owner and repository name rather than edited in place: a checkout can carry a credential in its remote URL — GitHub Actions writes one during checkout — and this value is published to a world-readable file. An unrecognized remote yields no prefix and the hash renders as plain text. The page then validates the prefix and the hash again before interpolating either into an href, since `escapeHtml` does not escape quotes.

## Touch, and the headline chart

Neither chart worked on a phone or tablet. The phase chart listened for `mousemove` alone, so touch never moved its cursor off the latest commit; the headline chart had no tooltip on any device.

Both now use pointer events, so one handler serves hover, tap and drag. `touch-action: pan-y` leaves vertical scrolling to the page. `height: auto` stops a narrow screen from drawing the chart at a third of its size inside a box of dead margin. A visible cursor rule and band marker were added, because a tap that silently changes a tooltip elsewhere on the page reads as nothing happening — and the tooltip moved above the composition bar, whose legend is one band per line on a phone and pushed the readout off the bottom of the display.

Phase chart on a phone

## Saying what a flag asserts, without spending the page on it

"Flagged" was a verdict with nothing behind it. The rule is easy to misread in three ways: it is noise-relative rather than a percentage, it compares against a trailing window rather than the previous commit, and it takes an absolute difference — so a large speedup is flagged exactly as a regression is.

The epoch's pinned rule and the trailing-window median it compared against are now derived and published, so each verdict carries its own figures: *"Flagged: 138.50 ms against a trailing 10-run median of 143.86 ms, 3.72% faster — beyond 3× the combined noise of the two."* Naming the direction there is what makes a flagged speedup read as one, without renaming the concept the ADR and the manifest both call flagging.

The prose version lives in a collapsed `<details>`, next to the disclosures pattern already used further down. It answers a question a reader asks once, and in front of the panels it described it cost more attention than the panels did.

Workloads section

<details>
<summary>The explanation, expanded</summary>

Workloads section expanded

</details>

## Defects found on the way

- **Cursor out of step with the drawing, horizontally.** The old handler mapped the full element width onto the series range, but points are plotted between x=40 and x=880 of a 900-wide viewBox — the cursor lagged its column by 40px at the left edge and needed 20px of overscrub past the right edge to reach the last commit. Inverting the plot's own `x()` through the SVG's CTM fixes both axes at once.
- **Listener accumulation.** The draw functions attached listeners on every redraw, so each platform or workload change stacked another closure over retired data. Interaction is now bound once and dispatches to the current handler.
- **`aria-labelledby` overrode `aria-label`** on both charts, so the arrow-key instructions were never announced. The caption is now the description and the label is the name.
- **Epoch-boundary delta.** The headline tooltip must not subtract two indexes across an epoch boundary — each epoch normalizes against its own baseline, so that would report a baseline change as compiler movement. It says so instead of showing a number.
- **Neither chart said when anything happened.** Both now carry date ticks under the plotted range, chosen by position so a tick always lands on a real measurement rather than pointing at a gap, and both tooltips name the measurement time.

## Testing

- `//:performance-commit-annotation-tool-tests` — 12 tests for the commit annotations. One builds a real git repository with a merged topic branch and pins the property that makes ordinal subtraction meaningful; others pin that a tokenized remote never reaches the published prefix and that lookalike hosts (`github.com.evil.test`) yield no link.
- `//crates/rue-bench:rue-bench-test` — 71 pass, including two new: the window median travels with the flag, and the epoch publishes the rule it pins.
- The repository has no JS test harness, so the dashboard was verified by driving the real template and `performance.js` in Chromium across desktop, phone and tablet — 101 checks in seven suites: 43 interaction (band hit-testing at four heights, keyboard navigation, touch tap and drag, focus-ring behaviour, stale-listener detection), 23 pin/unpin/Escape and commit-link, 18 date-axis and flagging wording including that the explanation stays collapsed, 6 link-prefix rejections including `javascript:` and quote-breaking values, 4 pointer-to-column alignment against the drawn geometry, 4 accessible-name and live-region, 3 degenerate-data (data predating each new field, no index points, a single measured commit). Those checks are not committed; worth a follow-up if the site grows more interactive code.

## Note for review

The ADR is untouched, and its wording still matches the page now that "flagged" is restored.

Screenshots are hosted on the orphan `pr-assets` branch, which shares no history with `trunk` and is never merged.


DorianListens:
Dropped from the merge queue on `premerge (linux-x64)` — [run 31289144368](https://github.com/rue-language/rue/actions/runs/31289144368/job/93183246634). The failing target is `//crates/rue-compiler:rue-compiler-differential-oracle-test`. I do not believe it belongs to this PR, and I have not pushed anything in response. Evidence:

**The target is unreachable from this diff.** Its deps are `rue-compiler`, `rue-cfg`, `rue-span`, `rue-target`, `sha2`. This PR touches `crates/rue-bench/src/derive.rs`, three `scripts/`, two `website/` files, and `BUCK`. `rue-bench` is a leaf — nothing depends on it. The only way this diff reaches premerge at all is the new `//:performance-commit-annotation-tool-tests` target, a Python script test, which cannot kill an unrelated Rust test binary.

**The failure produced no test output.** Every other target in that job printed its results (`813 passed`, `810 passed`). The oracle target emitted no `thread … panicked`, no assertion, no test lines at all — just `✗` and `exit 32`. That is a killed process, not a test that decided something was wrong.

**It passes under load locally.** 12 consecutive runs with every core saturated, 4/4 tests each, ~2.2s per run, zero failures.

**It is intermittent, and not new.** It failed on the PR-head run at 00:45, passed on the PR-head run at 01:51, then failed in the merge group at 01:58 — same commit range, same job, both outcomes. `premerge (linux-x64)` also took down the merge group for #2137 on 2026-08-07 ([run 31225370145](https://github.com/rue-language/rue/actions/runs/31225370145)), which merged on a retry 27 minutes later.

One correction to something I said earlier in this thread: I attributed this to the two-second child-process budget that `AGENTS.md` documents. That was wrong — that note covers the generated-oracle smoke target, and `tests/differential_oracle.rs` contains no timing primitives at all. The budget is not what is being exceeded here.

I cannot re-queue from here (`rerun-failed-jobs` returns 403 for this integration), so this needs a re-queue from someone who can. If it fails the same way a second time in the queue, the flake is frequent enough to be worth its own issue against the premerge job rather than more retries.

---

PR #2145, opened 2026-08-09
ADR-0069: CI work scheduling for a compiler monorepo

DorianListens:
Answers the sharding reassessment (refs RUE-1250) with a measured decision, and raises it into an ADR because the measurements say the interesting question is not the one that was asked.

Docs only. No workflow, BUCK, or source changes.

## The decision RUE-1250 asked for

**Keep four CLI shards.** Reject reduce, remove, and conditionalize-by-event.

The count has to be judged against the slowest thing CI cannot shard away, which is the native ARM64 lane — one runner by construction, 341–407s on every run. On the four sampled runs where the corpus actually executed:

| shards | corpus wall | breaches the ARM64 floor |
| --- | --- | --- |
| 4 (today) | 269–297s | 0 of 4 |
| 2 | 450–550s | 4 of 4 |
| 1 | 900–1100s | 4 of 4 |

A derivation confirms the constant rather than contradicting it: `ceil(total ÷ floor)` gives 3, but measured shard skew is 8–25% and 366s × 1.25 breaches the 407s floor, so with a skew allowance it lands on 4. Nobody had written that reasoning down, which is how the question came to be re-asked from scratch.

An earlier draft in this branch recommended reducing to two. That draft measured the shards' slack against `premerge`, which is 279–576s ahead of them — and that lead turned out to be a defect rather than a budget. The notes record the correction rather than hiding it.

## Why this became an ADR

Profiling the lanes the shards were being compared against found the actual critical path, and it is not a topology problem:

- `premerge (linux-x64)` was the longest job in **9 of 9** runs. It runs 80 targets; two are **81.4%** of it.
- `//crates/rue-compiler:scaling-matrix-test` is a strict superset of `rue-compiler-test` — 816 tests versus 813, all 813 shared. The lane runs those 813 twice to gain three.
- Both native lanes are one step, and that step is **99.1%** `rue-compiler-test` — against `docs/process/ci.md:46`, which states the native lanes deliberately do not repeat the broad unit suite, while `scripts/validate-ci-gate.py:204` pins the repetition in place.
- Inside that target, **one test function is 181.7s** against a 5.7s runner-up. It is compiled into two targets and runs on three platforms: four executions per run, together 56–59% of the critical path.

Sharding cannot touch this. Partitioning those 813 tests four ways gives 3.4s / 6.8s / 7.2s / **227.0s**, and cost-balancing cannot do better when one item weighs 181.7s against ~44s of everything else. Tracked as RUE-1262.

## What the ADR proposes

Rue has three good mechanisms — the determinator (RUE-1119), cacheable corpus actions (RUE-1118), and sharding (RUE-1116/RUE-1158) — each applied to a different hand-picked subset and composed in the reverse of their cost order. ADR-0069 adopts a uniform model: every unit of CI work is a Buck target carrying its own tier and platform scope, no workflow file names a target, and every lane is a generated selection over the live graph.

It also names the property that makes this repository different from the systems those mechanisms were designed for. **The compiler is a universal dependency, so change-based selection and input-keyed caching share a blind spot covering the dominant change class.** That is measured from both directions: RUE-1130 records zero deselections across 9 runs × 18 corpus jobs, and a third of the runs sampled here rebuild the compiler from cold. Two consequences follow — sharding is load-bearing rather than obsolete, and eliminating duplicated work outranks both other mechanisms because it is the only lever that pays in every regime.

## How it fits the CI Improvements project

The six phases map onto issues that already exist, and the ADR follows their stated constraints rather than re-litigating them:

| phase | issue |
| --- | --- |
| 1. Eliminate the duplicated critical path | RUE-1262 |
| 2. Duplication gate | new |
| 3. Platform scope as a target attribute | new (RUE-1262 scope C) |
| 4. Determinator plans every lane and reports its regime | RUE-1130 |
| 5. Real input-keyed compile actions | RUE-1164 (milestone "Buck-native test actions"), RUE-1222 |
| 6. Floor-aware packer with an indivisible-item alarm | new |

Notably, phase 5 follows RUE-1164's explicit constraint — *do not use an aggregate success-stamp action as the destination* — rather than generalizing RUE-1118's stamp, which that issue already designates as evidence rather than design authority. RUE-1131 is adjacent: it is the same fixed-cost problem the ADR names in the compiler-cold regime.

## Projected effect

Phase 1 alone, modelled against the nine runs using each run's own step durations: `pull_request` median critical path 820s → 361s, `merge_group` 456s → 185s, whole-sample range 341–873s → 170–393s. Afterwards the binding constraint is distributed across `valgrind`, the cold compiler build, reproducibility, and one CLI shard rather than concentrated in one function.

## What this does not establish

- **Reliability**, which RUE-1250 lists in its acceptance criteria. Nine runs over two days compute no flake rate; `correctness-repetitions.yml` is the right source. RUE-1250 is referenced rather than closed for this reason.
- Per-target measurement on real ARM64 and macOS runners. Those lanes were profiled by step; the 99.1% concentration is inferred from target composition measured on x86-64.
- Behaviour of genuinely cold fork PRs, which carry no BuildBuddy key.

## Validation

`//:adr-registry-validation`, `//:repository-quality-gates`, and `//:tutorial-snippet-tests` pass (4/4). ADR index regenerated with `scripts/validate-adrs.py --write`; registry valid at 70 records.

Refs RUE-1250. Refs RUE-1262. Related to RUE-1164, RUE-1130, RUE-1131, RUE-1222.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AjuvbCasMPCxH9y1hA1Qbr)_

### DIFF
--- a/docs/designs/0069-ci-work-scheduling.md
+++ b/docs/designs/0069-ci-work-scheduling.md
@@ -1,0 +1,352 @@
+---
+id: 0069
+title: "CI work scheduling for a compiler monorepo"
+status: proposal
+tags: [process, ci, testing, build, performance]
+feature-flag: null
+created: 2026-08-09
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["RUE-1250", "RUE-1262", "RUE-1164", "RUE-1130", "RUE-1131", "RUE-1222", "RUE-1116", "RUE-1118", "RUE-1119", "RUE-1157", "RUE-1163"]
+---
+
+# ADR-0069: CI work scheduling for a compiler monorepo
+
+## Status
+
+Proposal
+
+## Summary
+
+Rue's CI has three good mechanisms for not doing work — a change determinator
+(RUE-1119), cacheable corpus actions (RUE-1118), and corpus sharding
+(RUE-1116/RUE-1158) — each applied to a different hand-picked subset of the work
+and composed in the wrong order. Measuring nine runs shows the consequence: the
+critical path is a single test function executed four times, and no topology
+change can move it.
+
+This ADR adopts a uniform model. Every unit of CI work is a Buck target carrying
+its own tier and platform scope; no workflow file names a target; every lane is a
+generated selection over the live graph. It also names the property that makes a
+compiler monorepo different from the systems these mechanisms were designed for:
+**the compiler is a universal dependency, so change-based selection and
+input-keyed caching share a blind spot covering the dominant change class.**
+Eliminating duplicated work — which helps in *every* regime — therefore ranks
+above both, and scheduling ranks above neither.
+
+## Context
+
+### What was measured
+
+Nine `CI` runs (2026-08-07/08, four `pull_request`, five `merge_group`), plus
+`buck2` run directly against the live graph. Full evidence is in
+`docs/notes/rue-1250-shard-topology-analysis.md`,
+`rue-1250-premerge-critical-path.md`, and `rue-1250-ci-architecture.md`.
+
+The finding that motivates this ADR is concentration, not imbalance:
+
+- `premerge (linux-x64)` was the longest job in **9 of 9** runs, 279–576s ahead
+  of the slowest CLI shard.
+- That lane runs 80 targets. Two are **81.4%** of it — and
+  `//crates/rue-compiler:scaling-matrix-test` turns out to be a strict superset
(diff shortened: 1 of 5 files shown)


steveklabnik:
Accepted, this all makes sense to me. I dunno if you want to make linear issues since it's your ADR, @DorianListens

PR #2147, opened 2026-08-09
Apply affected-target determination to every CI lane

DorianListens:
Implements RUE-1130 and marks ADR-0069 accepted. Follows #2145, which merged while this was in progress, so this is a new pull request rather than a push to that one.

> **Description rewritten at `cd4b8354`.** The original said `linux-premerge` was deliberately left alone. That was reversed during review of the direction: it is now narrowed rather than gated, which is the central change here and the thing most worth reviewing.

## The decision RUE-1130 asks for

**Keep and extend to everything, not remove.**

The issue observed zero deselections across 9 runs × 18 corpus jobs and asked whether the selector earns its complexity. A wider sample says the deselection count was the wrong statistic: across **55 runs classified by change class and 21 timed to step level**, a two-file documentation change costs **444s against 465s** for a compiler change — 4.5% less. It still builds the compiler and still runs 813 compiler unit tests on linux-x64, then the same 813 again on linux-arm64 and again on macOS.

The determinator was already computing the right answer. It was wired to 7 of ~14 jobs, and not to any job that dominates a run.

## Gating and narrowing

Two granularities of the same determinator output:

- **Gating** decides *whether* a lane runs. Six lanes now consult it: `native (linux-arm64)`, `native (macos-arm64)`, `release`, `valgrind`, `asan`, `compiler reproducibility`. Selection stays *within* the job, so the aggregate gate and branch protection are untouched and a deselected lane costs only runner spin-up.
- **Narrowing** decides *what* a lane runs. `linux-premerge` is never gated — skipping the broad-discovery lane on a representative subset is the RUE-924 failure mode — but it now builds the impacted closure instead of `//crates/...`, and `test.sh` runs `impacted ∩ tier` in place of `//... ∩ tier`. The native lanes intersect their eight unit targets the same way.

Membership still comes from the live graph, so a target added since any list was written is still discovered; it is simply not built or run when the diff cannot reach it. Building a test binary is most of what a unit target costs, and this lane spends 286–317s building every crate whenever a compiler crate changes.

Narrowing is **declined** when nothing is impacted, or when so much is that `//...` is the better expression (`RUE_AFFECTED_NARROW_LIMIT`, default 600). Consumers key on a `narrowed` flag rather than on the list being non-empty, so an absent, empty, or oversized list always reads as "run everything". `test.sh` applies the same rule: unset, unreadable, or empty `RUE_TEST_TARGETS_FILE` runs the full pattern.

One asymmetry is called out in the code rather than left implicit: an explicit list that filters to no tests is legitimate, and buck2 exits 0 reporting `NO TESTS RAN` — indistinguishable by exit status from a narrowing bug. The selected count is echoed rather than inferred from a silent green.

The ASan harness is a standalone Cargo project outside the Buck graph, so BTD cannot see it; `crates/rue-runtime-asan/` forces a full run rather than relying on a proxy target.

## Two gates that fail closed

Both were written because this PR tripped the failure they now prevent.

- **Undeclared job outputs.** `selected_lanes` was read in five places before being declared. GitHub resolves an undeclared output to the empty string rather than failing, so every gated lane would have deselected itself on a selective run — RUE-924 through a silent empty string. `validate-ci-gate.py` now requires every `needs.<job>.outputs.<name>` to be declared by that job. It then caught the same class again on `impacted`/`narrowed`.
- **Lane-target drift.** `lane_targets` is a second copy of the native lane's eight targets. A target the job runs but the determinator does not list is invisible to selection, so the lane can be deselected — or narrowed away — by a diff that reaches it. The two lists are now compared, failing closed, including when the lane list cannot be read at all.

## Measured effect

Gating, on the four peripheral runs in the sample:

| run | runner time freed | critical path |
| --- | --- | --- |
| 31199530662 | 1034s | 409s → 409s |
| 31231601304 | 905s | 463s → 463s |
| 31291960267 | 954s | 460s → 460s |
| 31293391236 | 969s | **395s → 278s** |

Gating alone is a **cost win, not a latency win** — the critical path moves in one of four, because `premerge` dominates the rest. That is what narrowing addresses, and why the two landed together rather than gating being called done.

Not measured: the narrowed path's effect on a real run. `test.sh` and `.github/workflows/` are both in the force-full list, so this PR — like any PR that touches CI — structurally cannot exercise narrowing in its own CI. All four fallback paths were verified locally against a real buck2, including a narrowed run that executed exactly the two targets it was given. **The first real narrowed run will be the next docs-only PR after this merges, and is worth watching.**

## Cost

Gated lanes now wait on `affected-targets` (11–25s) before starting. They are not the critical path in 18 of 21 timed runs, so this is paid out of slack — but it is a real regression for the three runs where a native lane *was* the critical path.

## ADR-0069

Marked `accepted`, phase 2 recorded as done. Adds the standard this work is held to — *the repository should grow without anyone editing CI to keep it correct or fast* — and a ledger of everything still hand-maintained, what each breaks as the repo grows, and whether it is gated or merely fails open. Two rows remain ungated: `SELECTABLE_CORPUS`, and the 600-target narrowing limit, which is a number picked rather than measured.

## Tests and validation

The selector suite goes from 53 to **83 checks**: lane recognition, representative-target presence, the fail-open matrix, corpus/lane state separation, out-of-graph ASan sources, the intersection helper's ordering and empty/missing inputs, and `test.sh`'s four fallback paths including that a narrowed suite keeps its tier and deferral filters. `test-validate-ci-gate.py` goes to 21 tests covering both new gates.

Green locally: the four Python validators, 83 selector checks, 21 gate-validator tests, 13 Buck contract gates, and **actionlint 1.7.12 with shellcheck** — now installed locally, so that check is no longer one that can only fail in CI.

Refs RUE-1130. Refs RUE-1250. Refs RUE-1262.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,10 +236,13 @@ jobs:
       # and building its test binaries is the cost this narrowing exists to
       # avoid — measured at 286-317s whenever a compiler crate changes.
       - name: Build all targets
+        env:
+          NARROW_COUNT: ${{ steps.narrow.outputs.count }}
+          NARROW_FILE: ${{ steps.narrow.outputs.file }}
         run: |
           set -euo pipefail
-          if [ "${{ steps.narrow.outputs.count }}" -gt 0 ]; then
-            mapfile -t targets <"${{ steps.narrow.outputs.file }}"
+          if [ "$NARROW_COUNT" -gt 0 ]; then
+            mapfile -t targets <"$NARROW_FILE"
             scripts/ci-timed "linux-x64 build (narrowed)" -- ./buck2 build "${targets[@]}"
           else
             scripts/ci-timed "linux-x64 build" -- ./buck2 build //crates/...
@@ -396,6 +399,10 @@ jobs:
       # expensive part and an unimpacted one has nothing to prove.
       - name: Run native backend, linker, runtime, ABI, allocator, and target tests
         if: steps.sel.outputs.run == 'true'
+        env:
+          NARROW_COUNT: ${{ steps.narrow.outputs.count }}
+          NARROW_FILE: ${{ steps.narrow.outputs.file }}
+          LANE_NAME: ${{ matrix.name }}
         run: |
           set -euo pipefail
           targets=(
@@ -408,14 +415,14 @@ jobs:
             //crates/rue-allocator:rue-allocator-test
             //crates/rue-target:rue-target-test
           )
-          if [ "${{ steps.narrow.outputs.count }}" -gt 0 ]; then
-            mapfile -t targets < <(scripts/affected-targets intersect "${{ steps.narrow.outputs.file }}" "${targets[@]}")
+          if [ "$NARROW_COUNT" -gt 0 ]; then
+            mapfile -t targets < <(scripts/affected-targets intersect "$NARROW_FILE" "${targets[@]}")
           fi
           if [ "${#targets[@]}" -eq 0 ]; then
             echo "native units: no impacted unit targets on this diff; the merge-queue run covers them."
             exit 0
           fi
-          scripts/ci-timed "${{ matrix.name }} native units" -- ./buck2 test "${targets[@]}"
+          scripts/ci-timed "$LANE_NAME native units" -- ./buck2 test "${targets[@]}"
 
       - name: Run every declaratively platform-scoped spec and CLI case
         if: steps.sel.outputs.run == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,17 +281,33 @@ jobs:
             name: macos-arm64
     runs-on: ${{ matrix.os }}
     name: native (${{ matrix.name }})
+    if: ${{ always() }}
+    needs: affected-targets
     steps:
       - uses: actions/checkout@v6
 
+      # RUE-1130: consult the determinator before any heavy setup. On a
+      # selective pull_request run where nothing this lane executes is affected,
+      # the heavy steps are skipped and the check still reports success, so no
+      # branch-protection change is required. Fail-open: unless the determinator
+      # explicitly says full=false and omits this lane, the lane runs.
+      - name: Lane selection
+        id: sel
+        env:
+          RUE_AFFECTED_FULL: ${{ needs.affected-targets.outputs.full }}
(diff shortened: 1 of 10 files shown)


DorianListens:
`premerge (linux-x64)` failed on `c19dfd9b` in `rue-compiler-differential-oracle-test`. It is not this PR's diff, and it has a signature worth capturing — this is the known intermittent tracked by RUE-1268 / RUE-1260, following RUE-1240 and RUE-1244.

**Why it is not this change:** the branch contains no `crates/` changes at all (`git diff --stat origin/trunk...HEAD -- crates/` is empty). The same 80-target lane passed on the two earlier heads of this PR with identical compiler code, and the failure is in the one target this PR cannot influence.

**Reproduced locally, with the signature.** On the same tree, linux-x64, 4 cores:

| schedule | result |
| --- | --- |
| isolated, 10 runs | 0 failures |
| 4-way self-concurrent, 12 runs | **1 failure** |

```
test bounded_corpus_matches_stepwise_fresh_sessions ... FAILED
panicked at crates/rue-compiler/tests/differential_oracle.rs:301:18:
oracle corpus must have stable CFG presentation: CompileErrors { errors: [
  InternalError("rooted CFG query aborted: Cycle([
    NodeIdentity { family: "compiler.layout", key: "Array { element: Nominal(Named(… name: \"D\" …)), len: 2 };target=X86_64Linux;preview=StablePreviewFeatures([])" },
    NodeIdentity { family: "compiler.layout", key: "Nominal(Named(… name: \"Inner\" …));target=X86_64Linux;preview=StablePreviewFeatures([])" }
  ])")
]}
```

That answers RUE-1268's open question directly: it is **the same `compiler.layout` false-cycle signature as RUE-1244**, still present with the RUE-1244 fix (`331dc0e0`) in the base — so the stale-reverse-edge fix is incomplete rather than there being a third mode. RUE-1268 records that the signature was unrecoverable from the Actions log tail; this is that signature.

One methodological note for the repro. RUE-1244's validation ran 1,000 fresh-process runs and 64 concurrent oracle runs without hitting this. The schedule that reproduces it here is much smaller: **four concurrent copies on a four-core box** — saturation rather than oversubscription. 64-way on a large host may simply not produce the same interleaving.

Not fixing it in this PR: it is a compiler-correctness bug in an in-progress issue's territory, unrelated to this change, and folding a query-runtime fix into a CI-selection PR would be wrong. This needs a re-run; I do not have permission to trigger one.

---

PR #2151, opened 2026-08-09
Validate retained dependencies against their exact node

DorianListens:
The differential oracle failed intermittently in merge-group premerge four times. Sixty sequential local runs found nothing; **eight concurrent test processes on four cores reproduced both failure shapes at 5.4%**, because each process opens its own `available_parallelism()`-wide runtime and the resulting oversubscription widens windows the merge-group runner hits on its own.

They are two independent defects in `rue-query`.

## 1. Validation certified a dependent against a different node

Stamps are per-node counters. A retained dependency observation names one exact node **incarnation**, but re-demand resolves the family memo by **key**. Retention retires a node from that memo once it holds no terminals and no live users; the next request for the key builds a fresh incarnation whose stamps restart at one, while the validation walk holds the retired incarnation alive through the incarnation registry. Validation then compared stamp against stamp — the fresh node's first stamp reads `1`, the retained observation reads `1` — and declared the dependent green.

Traced in a failing run: `compiler.cfg` incarnation 558 was "validated" by a terminal published by incarnation 577.

```
VS-ENTER  fam=compiler.cfg inc=558          <- validating the retained edge (558, stamp 1)
VS-DEMAND inc=558
REUSE     fam=compiler.cfg inc=577 stamp=1  <- answered by a different node
VS-DEMAND-RESULT inc=558 kind=terminal inc=577 stamp=1 exec=Reused
REUSE     fam=compiler.optimized-cfg inc=557 regonly=true   <- certified green
```

**This is a silent wrong answer, not only a retention gap.** The added regression test reuses a stale `1` where the current value is `2`. The `DependencyNotObserved` panic was the same bug caught downstream: the endorsement and lease landed on the fresh incarnation while the retained edge still named the retired one, so the batch's cone walk found no lease for it.

Re-demand now declines a superseded incarnation up front, and the answering incarnation is checked again on the result, so retention retiring the node mid-walk makes the dependent dirty rather than provably green.

## 2. A scheduling artifact was reported as a dependency cycle

The other shape was a `compiler.layout` cycle between `[D; 2]` and `Inner`, whose real edges run `Inner -> [D; 2] -> D` and cannot be cyclic. The captured wait graph shows two tasks reaching an overlapping set of layout nodes in opposite orders — an ABBA conflict over in-flight attempts. No body asked for anything that asks back; the loop existed only in the wait graph, and the loser, decided by thread interleaving, failed its whole rooted request with `QueryAbort::Cycle`.

A declined join now claims a private attempt instead. Evaluation is deterministic, so duplicating a computation costs work and never an answer, and `publish` folds an equal result back onto the existing stamp.

True cycles stay exact. They are a property of the request's structure, so `Task::stack_cycle` decides them — and a task now carries the chain of nodes enclosing it across registered batch boundaries. Batch children evaluate on their own tasks with their own empty stacks; without the enclosing chain a real cycle through a batch would duplicate forever instead of being reported. `registered_batch_dependency_cycle_is_reported_exactly` covers that, and fails (after 90s of runaway duplication) if the ancestry is dropped.

Two existing tests asserted the old contract and are rewritten. Both used caller-supplied closures where neither root's own graph is cyclic, so the `Cycle` they asserted was the artifact above rather than a program property.

## Relationship to RUE-1240 and RUE-1244

These look like the same two root causes surfacing downstream. The batch-cone panic is verbatim the RUE-1240 signature and the false layout cycle is verbatim RUE-1244's, and both still reproduced on trunk after those issues were closed. Each earlier fix patched where the symptom appeared — validation proof bookkeeping, and absorbing cycles on the validating side — rather than the incarnation-blind stamp comparison and the choice of wait-graph victim. That is why they did not hold.

## Verification

| | failures |
| --- | --- |
| trunk, 8 concurrent processes x 4 cores | 5.4% (issue), 2/80 reproduced here |
| defect 1 fixed | 8/240 — all cycle-shaped |
| both fixed | **0/320**, and 0/160 through the soak script |

`//crates/...` is green (37/37 targets), clippy is clean, and the oracle target still runs in ~2.6s.

## On the two supporting changes

* **Diagnostics, kept unconditionally.** `buck2 test` prints a per-target tally and a `✗ <target>` line, so a target whose process dies without emitting a result line left nothing in the job log — which is why three ejections were unattributable and cost an investigation each. The event log recorded the child's stdout, stderr, and exit status the whole time. `scripts/ci-test-failure-report` prints every non-passing result in full; the premerge job reports it on failure and uploads the complete record. This pays off on any failure, not this one.
* **Contended soak, dispatch-only.** The issue asked for a standing lane on the strength of three recurrences — but per the section above those were the same two causes, so a scheduled lane would now mostly rediscover fixed bugs. `scripts/ci-soak-contended` is therefore wired as a manual job with no `schedule:`. It still earns its place on demand: sequential repetition cannot reach this schedule at all, so when a merge-group run goes red in a query-runtime target with no attributable cause, dispatching it against trunk answers whether the pull request under test is implicated. Restoring the cron is a one-line change if that stops being enough.

Fixes RUE-1260.

https://claude.ai/code/session_01G6pMQRtyjF8b464mnUwYdM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,10 +253,40 @@ jobs:
         # `rue_ci_dedicated_lane` label in BUCK, and test.sh subtracts that label
         # when CI=true. This step no longer names them: the CI-contract job
         # fails if the labeled set and the platform-corpus matrix disagree.
+        #
+        # RUE-1260: `buck2 test` prints a per-target tally and a `✗ <target>`
+        # line, so a target whose process dies without emitting a result line
+        # leaves nothing here to debug from — three merge-group ejections were
+        # unattributable for exactly that reason. The event log recorded the
+        # child's stdout, stderr, and exit status the whole time, so report it
+        # on the failure path and keep the complete record as an artifact.
         env:
           RUE_TEST_TIER: premerge
           RUE_TEST_TARGETS_FILE: ${{ steps.narrow.outputs.file }}
-        run: scripts/ci-timed "linux-x64 premerge" -- ./test.sh
+          RUE_PREMERGE_TEST_RESULTS: ${{ runner.temp }}/premerge-test-results.txt
+        run: |
+          # The default step shell runs with `-e`; the suite's status has to
+          # survive long enough to report on it, so exit on it explicitly.
+          set +e
+          scripts/ci-timed "linux-x64 premerge" -- ./test.sh
+          status=$?
+          if [ "$status" -ne 0 ]; then
+            scripts/ci-test-failure-report || true
+            scripts/ci-test-failure-report --all >"$RUE_PREMERGE_TEST_RESULTS" 2>&1 || true
+          fi
+          exit "$status"
+
+      # RUE-1260's per-target report and RUE-1268's preserved command log answer
+      # different questions on a failure — which target died and what it printed,
+      # versus the full output of the command that ran it — so both upload, under
+      # distinct artifact names.
+      - name: Upload premerge test results
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: premerge-test-results-linux-x64
+          path: ${{ runner.temp }}/premerge-test-results.txt
+          if-no-files-found: ignore
 
       - name: Run explicit cross-backend compilation and encoding coverage
         # Both the x86-64 and AArch64 encoders are host-independent Rust tests.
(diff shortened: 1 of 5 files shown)


steveklabnik (approved):
I’m Codex, posting this review under Steve’s account. I reviewed the exact-incarnation validation fix, the contended-join fallback and cross-batch structural cycle detection, both CI workflow changes, and the two new diagnostic scripts. I found no actionable issues.

Local validation on this head: all 134 rue-query tests passed; the full quick suite passed (31 targets); the new Buck event-log reporter reconstructed a real test result; and a minimal end-to-end run of ci-soak-contended completed successfully.

PR #2154, opened 2026-08-09
example: hash files to find duplicates

gvwilson:
-   Example taken from https://third-bit.com/sdxpy/dup/.
-   To run: `scripts/rue exec examples/dup/dup.rue examples/dup/*.txt` (Should report `examples/dup/a?.txt` are all duplicates and `examples/dup/b?.txt` are duplicates.)
-   Limitations:
    -   Uses its own hash function (nothing in standard library).
    -   Does not do recursive directory traversal (not in stdlib).
    -   Does not create hash table of duplicate files with hash code as key (no hash table in stdlib).


steveklabnik:
Awesome, thank you! Yeah those are some stdlib flaws for sure, we have a lot of stuff we want to do here in the future.

steveklabnik:
I thought we had fixed the `mapfile` issue, it's obviously unrelated to your change. The other one I took a glance at, since we test examples, I *think* it may need to be registered somewhere, but I'm not sure. I'll take a deeper look at that sometime soon.

steveklabnik:
I was right on both counts. @gvwilson if you could run `scripts/rue frontend-manifest` and rebase on latest trunk, this should pass.

gvwilson:
@steveklabnik steps taken:

```
git fetch upstream trunk
git rebase upstream/trunk # no conflicts
git push --force origin example-find-duplicate-files-by-hashing
```

Note that the example needs command-line arguments to run (it needs to be told what files to check) - where/how do I wire these in for CI?

And would https://third-bit.com/sdxpy/glob/ and https://third-bit.com/sdxpy/parse/ be good examples to tackle next? The former is a simple implementation of `glob`, the latter parses `glob` expressions to create matchers.

steveklabnik:
Perfect on the rebase, thank you.

So, on CI, we run them with no arguments, and you don't crash, so that *technically* passes even if it doesn't exactly exercise the real program correctly, haha. The way to exercise the actual duplicate-finding behavior with arguments is the way we do that in `crates/rue-cli-tests/cases/` — `examples_wordfreq.toml` is a template for exactly this situation: `source_path` points at the checked-in example, files sets up fixture files in the test's working directory, `program_args` becomes the program's argv, and it asserts exact stdout. (Note the fixtures there are inline in the TOML rather than reading the checked-in .txt files — those are still nice for running the example by hand.) This isn't really *ideal*, but it does work right now. So three possible paths:

1. this is good as-is. Doesn't truly exercise the program but that's okay. (Well, you do have to run that 'register' command still, that's why CI still fails).
2. add a `rue-cli-tests/cases` example with the argument and assert the real output
3. 1 or 2, but send in a follow up PR that would expand our examples infrastructure to do something more specific here. This is very ambiguous and not well specified, I wonder what @DorianListens thinks. But exercising examples in a better way may make sense. Or at least, I wonder if what we really want is like, to split out the tests for `rue`'s CLI and the tests for examples, using the same runner, as the runner itself is decent, it just feels less than ideal to mix the tests for the CLI and tests for the programs made using it, even if it's convenient. 

As far as those examples go, I'm happy to add anything you'd like. Dorian and I were just talking about this last night: we know that the language needs some more features to making writing it in more nice (for example, some sort of trait/interface abstraction, which is partially why the hash stuff isn't in std yet) but it's also nice to have more programs, so we can use those to inform priorities on what to add and how it might work. I'm really open to whatever, it's just fun to see more programs, no matter what they are. We have one or two *very large* ones in the tree that aren't even really real, just compiler stress tests.

gvwilson:
Adding `crates/rue-cli-tests/cases/examples_dup.toml` appeals - how do I run just that test with `scripts/rue test`? (I just tried `scripts/rue test examples_dup` and a couple of variations, but they all ran the entire test suite.)

steveklabnik:
It's not the most discoverable thing, but it is

```
scripts/rue cli cli.examples_wordfreq
```

for an existing test, or

```
scripts/rue cli cli.examples_dup
```

for the one you're adding.

gvwilson:
thanks @steveklabnik - next question is, what's CWD when tests are running? `scripts/rue cli cli.examples_dup` can't find my test files with either `a1.txt` (etc.) or `examples/dup/a1.txt` (etc.) as stdargs.

steveklabnik:
Ah, good point! you have to declare the dependencies explicitly:

```toml
source_path = "examples/dup/main.rue" # or whatever
files = [
  { path = "a1.txt", source = "..." },
  { path = "a2.txt", source = "..." },
]
program_args = ["a1.txt", "a2.txt"]
```

This will have the runner copy those files into the temporary directory where the test is built and run. I haven't actually tried this config with your PR directly, but I *think* from eyeballing it, that should work, with all of the `source`s being from the repository root

gvwilson:
Thanks @steveklabnik - passes on my machine.

steveklabnik:
@gvwilson I think we're at the final stage! Can you run `scripts/rue frontend-manifest`? This basically registers your new example with some of our testing stuff. (I probably should look into killing this step, I think we can do better without it these days, but you know, legacy...)

gvwilson:
done.

steveklabnik:
I am not sure what this failure is about, to be honest, it seems to be unrelated? I'm not sure if I can take a look for a bit here, sorry about that, but I'll check it out.

steveklabnik:
@gvwilson I found the time: we just need another rebase. Your branch is behind a commit that changed one of the schemas we have for a thing, and so the run ends up failing. Shouldn't need anything from you other than that.

gvwilson:
done.

PR #2161, opened 2026-08-10
[demo, do not merge] Evidence for #2160: a compiler-internals change

DorianListens:
**Demonstration PR. Not for merge — close once #2160 has landed.**

#2160 cannot exercise its own fix: it touches `.github/workflows/` and `scripts/affected-targets`, both on the determinator's force-full list, so its CI runs the unnarrowed path. That is the same blind spot that let the regression ship in #2147. This PR and its sibling exist to run the narrowed path against the fix and show what it does.

Stacked on `claude/premerge-build-regression-xnh6g7`, so it carries the fix. Because `RUE_AFFECTED_BASE_SHA` is `github.event.pull_request.base.sha` (`ci.yml:642`), the determinator diffs against the fix branch head — the fix's own workflow and script edits are not in this PR's changed-file set and so do not force a full run.

## The case this covers

A change to compiler internals: a doc comment on the accessor-graph edge counters in `crates/rue-compiler/src/cfg_query.rs`. Content is deliberately minimal; what matters is the change *class*. A `crates/rue-compiler` edit is the 74.5% case in ADR-0069's sampled mix, and it is the class that regressed — its impacted closure reaches the root-level corpus targets.

## What to look at

`premerge (linux-x64)` → **Impacted target list** and **Build all targets**.

Expected with the fix:

- narrowing engages — `impacted targets for this lane: N`
- the build step reports `linux-x64 build (narrowed)` over **only `//crates/...` targets**
- no `Waiting on root//:cli-tests-action`, `root//:cli-tests-shard-N-action`, `root//:spec-tests-action`, `root//:release-smoke-action`, or `root//:reproducible-programs-action` in the build step
- the CLI/spec/oracle corpora run **once**, in their own `test (linux-x64-*)` lanes, in parallel

Reference points for premerge wall time:

| | premerge | build step |
| --- | --- | --- |
| before narrowing (run 31332413558) | 12.1m | — |
| with the regression (run 31343268292) | 41.5m | 40.0m |
| with the regression (run 31342141141) | 33.7m | 31.8m |
| this PR | ← the number in question | |

The result should land at or below the 12.1m baseline, since the narrowed build is a subset of `//crates/...` rather than a superset of it.

## Sibling

The peripheral case is in the companion demo PR: a README-only edit, which deselects rather than narrows.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KZWH1A1JWstAednntzGoaD)_

### DIFF
--- a/crates/rue-compiler/src/cfg_query.rs
+++ b/crates/rue-compiler/src/cfg_query.rs
@@ -197,6 +197,20 @@ enum AccessorDagFailure<K> {
     Cycle(K),
 }
 
+/// Edge counters that make the linearity of accessor materialization
+/// assertable rather than merely intended (RUE-1284).
+///
+/// Validation and closure construction each visit every edge at most once, and
+/// the two walks are counted separately because they fail differently: a
+/// regression in validation shows up as repeated whole-graph walks, one per
+/// intermediate accessor, while a regression in closure construction shows up
+/// per root. A single total would average the two and hide either.
+///
+/// Both fields are `#[cfg(test)]`, so the struct is zero-sized in a release
+/// compiler and the `*_edge` calls compile away entirely. Keeping the calls
+/// unconditional at the call sites — rather than wrapping each in `cfg!` — is
+/// what stops the instrumentation from drifting out of step with the walks it
+/// measures.
 #[derive(Debug, Default)]
 struct AccessorGraphWork {
     #[cfg(test)]


DorianListens:
Result — run [31348880780](https://github.com/rue-language/rue/actions/runs/31348880780), against trunk at `6d1a580` (the fix). Closing; purpose served.

Narrowing engaged at 67 impacted targets. The cache state was equivalent to the regressed run, which makes this close to apples-to-apples:

| | regressed (31343268292) | this run |
| --- | --- | --- |
| impacted closure | 83 | 67 |
| build `Cache hits` | 90% | 96% |
| `Commands` | 795 (cached 719, **local 76**) | 750 (cached 719, **local 31**) |
| Build all targets | 40.0m | **4.5m** |
| premerge total | 41.5m | **7.9m** |

Same 719 cached commands in both; the delta is 76 → 31 locally-executed actions. Those 45 are the corpus runs.

The cache-independent check is what buck2 waited on inside the build-step window (up to `BUILD SUCCEEDED`):

```
18x root//crates/rue-oracle-diff:oracle-diff-test-action
12x root//crates/rue-oracle-diff:oracle-diff-spec-test-action
 6x root//crates/rue-compiler:rue-compiler-test
 1x root//crates/rue-oracle-diff:rue-oracle-diff-test
```

Zero root-level corpus actions, against six in the regressed run (`cli-tests-action`, `cli-tests-shard-1/3-action`, `spec-tests-action`, `release-smoke-action`, `reproducible-programs-action`). The crate-scoped oracle-diff actions are retained, as #2160 documents. Corpus actions appearing later in the log are in the `test.sh` step after `BUILD SUCCEEDED` — premerge-tier corpora with no dedicated lane, which is correct.

Caveat worth recording: the two runs are different diffs (83 vs 67 impacted), so closure size is not like-for-like. The commands triple and the zero-corpus-actions check are the claims that do not depend on that.

---

PR #2162, opened 2026-08-10
[demo, do not merge] Evidence for #2160: a README-only change

DorianListens:
**Demonstration PR. Not for merge — close once #2160 has landed.**

Companion to #2161. Together they cover the two change classes that reach `linux-premerge` differently. See #2161 for why #2160 cannot exercise its own fix and why stacking on the fix branch keeps the determinator honest.

## The case this covers

A documentation-only change: the front-page example now prints its result with `println` instead of `@dbg`. `README.md` is explicitly *selectable* rather than force-full (`scripts/test-affected-targets.sh:90`), so this is the peripheral class ADR-0069 measured at 444s against 465s for a compiler change.

## What to look at

`premerge (linux-x64)` → **Impacted target list** and **Build all targets**.

Expected:

- the determinator finds nothing impacted, so narrowing **declines** — `narrowed` is false and the build step falls back to `./buck2 build //crates/...`, exactly as it did before RUE-1130
- the six gated lanes (`native` ×2, `release`, `valgrind`, `asan`, `compiler reproducibility`) deselect
- no corpus action is built in the premerge job

This is the path #2160 leaves untouched, and that is the point: the fix changes only what happens when narrowing *engages*. If this run looks like a normal pre-RUE-1130 run, the fallback is intact.

It also exercises the third branch of the fixed step indirectly — a closure naming only corpora would print `no impacted //crates/... targets; this step has nothing to build` rather than silently skipping. That case is covered by unit tests in #2160 (`narrow: build-scope on a corpus-only closure yields nothing to build`), since producing it on a real PR needs a diff that touches corpus data and nothing else.

## On the change itself

`println("fib(10) = " + @to_string(fib(10)))` is the idiom the worked examples use (`examples/first/stats.rue:61`), needs no import, and labels its output. `@to_string` accepts any integer width (`intrinsics.rs:1361`, RUE-314), so the `i32` return of `fib` is fine.

Worth noting for its own sake: **nothing in CI compiles this snippet.** The tutorial snippet checker globs `*.md` under its own root and never sees the repository README, so the first Rue code anyone reads is unverified. That is a real gap, unrelated to #2160 and not fixed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KZWH1A1JWstAednntzGoaD)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ fn fib(n: i32) -> i32 {
 }
 
 fn main() -> i32 {
-    @dbg(fib(10));
+    println("fib(10) = " + @to_string(fib(10)));
     0
 }
 ```


DorianListens:
Result — run [31348882324](https://github.com/rue-language/rue/actions/runs/31348882324), against trunk at `6d1a580` (the fix). Closing; purpose served.

```
affected-targets: SELECTIVE run — btd affected-target closure over the PR diff
affected-targets: selected corpus targets: <none>
affected-targets: selected lanes: <none>
affected-targets: not narrowing (0 impacted targets; limit 600)
```

Narrowing declines, the build step takes the unchanged `//crates/...` fallback, every corpus and gated lane deselects (all at 0.1m), premerge total 1.7m. The fallback path #2160 leaves alone is intact.

Two corrections to what this PR predicted:

- It expected a corpus-only closure. The determinator returned **zero** impacted targets, which is the decline path, not the corpus-only path. So the third branch of the fix — `no impacted //crates/... targets; this step has nothing to build` — was **not** exercised by a real PR here. It is covered by the unit test, which ran and passed in this very job: `ok: narrow: build-scope on a corpus-only closure yields nothing to build` (02:08:41).
- The 0.1m build step is cache-served, not evidence about scope. Trunk had just built this identical tree in #2160's merge-queue run. A cold README-only run would still pay the full `//crates/...` build.

---

PR #2210, opened 2026-08-10
example: basic glob matching with '*' and alternatives

gvwilson:
-   Based on https://third-bit.com/sdxpy/glob/ but uses functions rather than matcher objects
-   `main` runs tests


steveklabnik:
Now that https://github.com/rue-language/rue/pull/2154 is in, we'll probably need a few similar steps. I've kicked off CI, but you may need to run that registration script and/or rebase, let's see what CI says.

PR #2226, opened 2026-08-11
ADR-0070: Rue program compilation as declared Buck actions

DorianListens:
Docs only — no rules, no BUCK changes, nothing implemented. Refs RUE-1164; this is the reviewed design that issue asks for and the design ADR-0069 phase 5 defers to.

*(Rewritten at eff07c5, after four adversarial rounds plus two passes from Steve, which together changed the mechanism, the counts, the target-resolution model, and the phase sequence.)*

## What this proposes

`rue_program` — a build rule owning one compilation — and `rue_program_test`, which consumes its executable and runs one runtime scenario, plus a `RueProgramInfo` provider. Many scenarios share one compile. **12 programs**: 4 large-example roots and 8 from the CLI corpus. The 34 auto-discovered example roots stay in the automatic smoke harness (their claimed second consumer was false — frontend-diff compiles only ruelex), and the one-case wordfreq root joined them this round: its CI shard and the unsharded `//:cli-tests` are scheduling alternatives for one scenario, not two consumers.

**Target resolution is hybrid, per Steve's second pass.** The same target labels must run natively on Linux x86-64, Linux AArch64 and macOS AArch64 — today's `sh_test`s manage it by never passing `--target`. So `RueToolchainInfo` carries the configured platform's native Rue target (resolved the way `toolchains//:rust` already is), an explicit `rue_target` attribute means intentional cross-compilation, `runs_natively` is computed rather than asserted, and the compile always passes the *resolved* target explicitly so the key never depends on host detection.

## Why, honestly

The 243.0s caldera and 80.7s meridian compiles occur only in `release.yml`. **The required path pays two canary compiles, measured at 2.3s and 3.1s**, and the corpora that were 91–97% of the merge-queue critical path are already cached by RUE-1118. This is not a critical-path optimization, and the ADR says so.

The real motivation is coverage. `cases/examples_meridian.toml` has six cases against one root that differ only in scenario; the harness recompiles for each — and CI runs none of them, because RUE-1083 disabled the section after budget kills. **The welded topology did not cost time, it cost coverage.**

## The mechanism

[...]

PR #2276, opened 2026-08-12
RUE-1404: rue_program rules, scan-derived manifest, and controls (ADR-0070 Phase 0)

DorianListens:
Phase 0 of ADR-0070 (accepted in #2226). Refs RUE-1404. **No existing target changes** — the fixtures package is the rules' proving ground, and every production suite runs exactly as before.

## What this adds

`rue_program` owns one compilation as **three static actions** — scan, derive, compile — and produces the executable as a first-class artifact with `RunInfo` and `RueProgramInfo`. `rue_program_test` consumes it and runs one scenario (inline `files` staged into a writable temp cwd, per the ADR). Rule names stay `rue_program`/`rue_program_test` per the ADR's forward-positioning note — this was flagged as a Phase 0 decision, so review is the place to overrule it.

**The derive step is where hermeticity lives** (`scripts/rue-program-derive-manifest.py`):

- manifest = accepted reads ∪ absent observations ∪ **every file of the declared std tree** — an undeclared absent arm is `DeniedLexical`-fatal (ADR-0051), and trusted-std acquisition is invisible to the scan;
- **fails the build** when any accepted read lies outside `srcs ∪ std` — the check that stops an out-of-`srcs` read being laundered into the shared cache;
- entries re-anchored via the envelope's recorded roots and written relative to the manifest's own directory, so a shared scan-cache hit is never a build failure on another machine. Unit-tested, including the relocation-invariance case: two envelopes recorded under different checkout roots derive **byte-identical** manifests.

**Hybrid target resolution** (`toolchains//:rue`): `RueToolchainInfo` carries compiler, std, and the configured platform's native Rue target, resolved per-platform like `toolchains//:rust`; no `rue_target` means native, setting it means intentional cross-compilation, `runs_natively` is computed, and the compile always passes the *resolved* target so the key never depends on host detection. Compiler and std are target-configuration deps, so a `//platforms:release` `rue_program` compiles with a release-built compiler, matching `$(exe_target //crates/rue:rue)` semantics.

**Precision reports are optional `ValidationInfo`**: a per-target report that only *lists* unread srcs (a sibling may read them), and the `rue_program_family` aggregate — deps are the macro's argument list, since a rule cannot discover its siblings — which fails only under `--enable-optional-validations` when a file is read by *no* sibling.

## The three ADR controls

1. **Boundary (contained Buck target).** `rue_program_boundary_control` materializes the out-of-`srcs` module as a hidden scan input — so the failure stage is the derive boundary in *every* execution environment, not an unresolved import in sandboxed ones — and succeeds iff exactly the expected violation is rejected. `//:corpus-stamp-check` asserts the marker.
2. **Clean-root / RE.** The merge-group RE canary now also builds `//fixtures/rue-program:hello` with `--prefer-remote --no-remote-cache` on the no-fallback platform.
3. **Digest sensitivity.** `scripts/check-rue-program-digests.sh`: declared-source edit re-runs the chain; revert is a cache hit; **undeclared-neighbour edit runs no `rue_*` action** — `corpus.bzl:19-25`'s false-pass window as a hard check. Passes locally.

## The fixture is chosen so green is meaningful

`fixtures/rue-program/hello` deliberately needs both mechanisms a naive manifest cannot produce: `sub/types.rue` imports `"shared.rue"`, probing the **absent** importer-relative arm first, and `main.rue` uses `@parse_i32` **without** `@import("std")`, so the manifest must carry std the scan never saw. The generated manifest verifies: absent arm declared, all 30 std modules present, every entry relative.

## Evidence

- fixture build → binary exits 42; scan carries no `-O`/`--target` (one scan per root, every flag set)
- `buck2 test fixtures/rue-program:` → 4/4 (scenario, writable-cwd, boundary control, family)
- `scripts/test-rue-program-derive-manifest.py` → 7/7, registered as `//:rue-program-derive-manifest-tests`
- `check-rue-program-digests.sh` → PASS (3 re-ran / cache hit / 0 ran)
- `bxl //test_tiers.bxl:validate` → 155 premerge, 5 slow, 3 stress — all new tests carry exactly one tier
- `//:ci-gate-validation`, `//:required-ci-container-pin-validation`, `//:repository-quality-gates` → pass

## Also

- Ticks ADR-0070's Phase 0 checkbox.
- Retires `AGENTS.md`'s stale `--prefer-remote` prohibition (RUE-320 Done 2026-07-18), the docs fix the ADR assigned to this phase.

https://claude.ai/code/session_01X1etH9FWkrVQEZHFNnWRri

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1etH9FWkrVQEZHFNnWRri)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -510,6 +510,12 @@ jobs:
             //crates/rue-runtime-abi:rue-runtime-abi-test
             //crates/rue-allocator:rue-allocator-test
             //crates/rue-target:rue-target-test
+            # ADR-0070 (RUE-1404): the one place CI evaluates the rue
+            # toolchain's non-x64 native_target select branches and runs an
+            # internally-linked rue_program on aarch64-linux and
+            # aarch64-macos. Narrowed like every other entry: it only runs
+            # when the compiler, the rules, or the fixtures are impacted.
+            //fixtures/rue-program:hello-runs-test
           )
           if [ "$NARROW_COUNT" -gt 0 ]; then
             # macOS runners execute this with the stock Bash 3.2, which has no
@@ -607,6 +613,57 @@ jobs:
           path: ${{ runner.temp }}/rue-compiler-repro-artifacts
           if-no-files-found: ignore
 
+
+  # ADR-0070 negative control 3 (RUE-1404): digest sensitivity of the
+  # rue_program action chain. Required like every job here — this repository
+  # has no unaggregated-job tier (//:ci-gate-validation enforces it), and a
+  # hermeticity control nobody gates on decays. The script asserts only what
+  # buck2 guarantees (steady-state convergence, fresh-probe retries past the
+  # lossy file watcher), and the lane decision skips it when nothing it
+  # exercises is affected.
+  rue-program-digests:
+    runs-on: ubuntu-latest
+    name: rue_program digest sensitivity (linux-x64)
+    if: ${{ always() }}
+    needs: affected-targets
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Lane selection
+        id: sel
+        env:
+          RUE_AFFECTED_FULL: ${{ needs.affected-targets.outputs.full }}
+          RUE_AFFECTED_LANES: ${{ needs.affected-targets.outputs.selected_lanes }}
+        run: scripts/ci-corpus-decision "rue-program-digests"
+
+      - name: Install dotslash
+        if: steps.sel.outputs.run == 'true'
+        uses: facebook/install-dotslash@v2  # pinned: the moving `latest` branch broke macOS runners (sha256sum flags) on 2026-06-18
+
+      - name: Cache dotslash artifacts
+        if: steps.sel.outputs.run == 'true'
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/dotslash
+          key: dotslash-linux-x64-${{ hashFiles('buck2') }}
+          restore-keys: |
+            dotslash-linux-x64-
+
+      - name: Provision remote cache
+        if: steps.sel.outputs.run == 'true'
+        env:
+          BUILDBUDDY_API_KEY: ${{ secrets.BUILDBUDDY_API_KEY }}
+        run: |
+          if [ -z "${BUILDBUDDY_API_KEY:-}" ]; then
+            echo "BUILDBUDDY_API_KEY unavailable (fork PR or unset); building without the remote cache."
+            exit 0
+          fi
+          scripts/provision-build-cache install
+          scripts/provision-build-cache apply
+
+      - name: Check rue_program digest sensitivity
+        if: steps.sel.outputs.run == 'true'
+        run: scripts/ci-timed "rue_program digests" -- ./scripts/check-rue-program-digests.sh
 
   # RUE-1119: on pull_request, select only the heavy corpus suites the diff
   # actually affects, using Meta's off-the-shelf Buck Target Determinator (BTD,
@@ -1036,6 +1093,7 @@ jobs:
       - linux-premerge
       - native-platforms
       - compiler-reproducibility
+      - rue-program-digests
       - affected-targets
       - platform-corpus
       - release
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,20 @@ jobs:
           ./buck2 build //crates/rue:rue --prefer-remote --no-remote-cache
           -c build.execution_platforms=root//platforms:remote_execution
 
+      # ADR-0070 negative control 2 (RUE-1404): remote execution materializes
+      # only declared inputs, so building a rue_program here is the standing
+      # proof that its scan/derive/compile chain reads nothing undeclared —
+      # the failure stage control 1 deliberately does not exercise. The hello
+      # fixture needs both hard manifest mechanisms (an absent importer arm
+      # and trusted-std reads), so a pass is meaningful, not vacuous.
+      - name: Build rue_program fixture remotely
+        env:
+          RUE_CI_REQUIRE_REMOTE_ACTIONS: 1
+        run: >-
+          scripts/ci-timed "remote rue_program build" --
+          ./buck2 build //fixtures/rue-program:hello --prefer-remote --no-remote-cache
+          -c build.execution_platforms=root//platforms:remote_execution
+
   rust-project:
     runs-on: ubuntu-latest
     steps:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -179,8 +179,10 @@ time. Do not bypass that coordination with a direct `buck2 test //...` when a
 full local run is intended. Quick, filtered, and targeted checks do not take the
 host lock. The optional BuildBuddy action cache uses one private user config and
 ignored per-worktree symlinks; see `docs/process/build-cache.md`. Never commit or
-print its credential, and do not use `--prefer-remote` while RUE-320 remains
-open.
+print its credential. Full remote execution is supported (RUE-320, Done
+2026-07-18): the repository wrapper defaults to `--prefer-local`, and
+`--prefer-remote` is an explicit opt-in for cache-population or RE-debugging
+runs, not the default local-development policy.
 
 Testing conventions:
 
--- a/BUCK
+++ b/BUCK
@@ -66,6 +66,25 @@ sh_binary(
     visibility = ["PUBLIC"],
 )
 
+# The rue_program tool surface (ADR-0070 / RUE-1404): the scan wrapper, the
+# manifest derivation script that owns the declared-boundary check, the two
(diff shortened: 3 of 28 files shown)


DorianListens:
Self-review of the implementation against ADR-0070 and against buck2's actual semantics, in the same adversarial mode the ADR went through. Nine findings: five fixed (8f47433, 918ce8e), four disclosed. The most instructive one is #5 — a control that had passed was asserting daemon luck.

## Fixed

**1. `--linker internal` was not pinned.** The ADR's key table specifies it; the first commit relied on the default. Before pinning I verified it is safe on all three platforms: `LinkerMode::Internal` is the compiled-in default (`queries.rs:42-46`), the internal linker has a full Mach-O path (`linker.rs:1100`), and the macOS native lane already compiles *and executes* internally-linked `aarch64-macos` binaries today (`abi_conformance.toml`'s `execute_if_native` cases run in the `cli abi` step). The `toolchains/BUCK` comment saying macOS needs system clang is stale with respect to Rue's own linker. Without the pin, a future default change would silently hand the compile action an undeclared `$PATH` binary.

**2. Two key-table rows were unimplemented.** `preview_features` and `link_archives` now exist; archives are carried as artifacts rather than path strings, because the bytes are read at link time and a string would key the path (the exact unsoundness round-one review flagged in the ADR).

**3. The derive action over-declared its inputs.** It listed all `srcs` as hidden, but reads only the envelope, the lists, and the std tree it walks — and the envelope's per-file content fingerprints already re-key it transitively on any source change. Under RE this materialized the whole source tree for a script that never opens it.

**4. The scan wrapper assumed the compiler's `RunInfo` is a bare binary.** It now takes the compiler as trailing argv, so a wrapped RunInfo (script plus fixed args) keeps working — relevant the day the toolchain points at a released compiler per Forward positioning.

**5. The digest control asserted properties buck2 does not have — and it had already passed once.** Re-running it during this review produced two *different* failures in consecutive runs:

- *"Revert then rebuild is a cache hit"* is not a local guarantee. OSS buck2 has no persistent digest-keyed local action cache; DICE may or may not retain the pre-mutation state, so A→B→A can legitimately re-execute. The green run in the PR body passed on daemon-state luck. The assertion is now **steady-state convergence** — after any mutation settles, an identical rebuild executes nothing — which is what digest keying actually promises.
- *Mutate-then-build* can fail forever, not just race: an inotify queue overflow during preceding build churn drops the write's event outright, and rebuilding never resurrects a dropped event (reproduced: instant invalidation on a quiet daemon, permanent invisibility after churn). Each retry now appends a **fresh probe**, giving the watcher a new event per attempt; the undisturbed-neighbour assertion keeps a settle window so a pass means "keyed by nothing", not "not seen yet".

Three consecutive clean passes after the rewrite. Also recorded in the commit because it's the kind of thing that should be checkable later: the failure was initially masked by piping the script through `tail`, which swallowed its exit code, so 8f47433 was pushed with this control red. In CI nothing pipes it.

## Disclosed, not fixed

**6. "One scan serves every flag set of a root" holds within a target, not across targets.** The scan command embeds the per-target output path, so two `rue_program`s differing only in `-O` would carry distinct scan cache keys. No multi-flag family exists today (release-smoke is a different *configuration*, which splits actions regardless). If one appears, the fix is hoisting the scan into an anon target keyed on (root, srcs, std, compiler); building that now would be speculative machinery.

**7. The RE canary's derive step needs `python3` on the BuildBuddy worker image**, which I cannot verify from here. If absent, control 2 fails loudly in the merge queue — the right failure mode, but worth knowing before merging.

**8. `rue_program_test` carries no remote-execution executor wiring.** Deliberate: tests run local under the noop test toolchain, which is the repo's standing configuration and OQ4's recommendation. Noting it so it reads as a choice, not an omission.

**9. `root ∈ srcs` is not validated at analysis time.** A typo'd `root` fails at scan time with the compiler's file-not-found rather than a clean analysis error. Cosmetic; an analysis-time check would need to compare against artifact short-paths and didn't seem worth the coupling.

---

DorianListens:
Second pre-merge pass, this time at what *surrounds* the implementation rather than the implementation itself — CI wiring, docs the change makes stale, coverage of the code paths CI never touches. Three findings, all fixed in 498d6a4.

## 1. Control 3 ran in no workflow

The digest-sensitivity script existed and passed locally — which is precisely the "hermeticity claim nobody executes decays" state the ADR's controls section warns about, reproduced by the person who wrote the warning. It's now the `rue-program-digests` job, mirroring the `compiler-reproducibility` skeleton: determinator-gated through a registered selectable lane (fail-open, keyed on `//fixtures/rue-program:hello`), cache-provisioned when the secret exists.

Worth recording how the "required or not" question resolved: my first version made it deliberately non-required — a day-old control that drives buck2's file watcher shouldn't block the merge queue until it has a track record — and `//:ci-gate-validation` **rejected the commit**: `CI job inventory has unaggregated jobs: rue-program-digests`. This repository has no unaggregated-job tier; every check gates or doesn't exist. That invariant is better than my trial period was, and the risk it forces us to accept is smaller than it looks: the script now asserts only what buck2 guarantees, and the flake modes I found were artifacts of a noisy shared daemon that a dedicated CI runner doesn't have. It joins `EXPECTED_REQUIRED_JOBS` and the `ci-success` needs list; the unit tests derive from the tuple and needed no edit.

## 2. CI never evaluated the toolchain's non-x64 select branches

The native lanes run a **fixed target list** — so nothing in CI ever resolved the `native_target` select's `aarch64-linux` or `aarch64-macos` arms, and no `rue_program` ever compiled off linux-x64. A typo in either arm would have shipped silently and broken every macOS developer's next `buck2 test //...`. (This also corrects my earlier self-review comment, which wrongly claimed the macOS lane would compile the fixture via the broad pass — it wouldn't have.)

`//fixtures/rue-program:hello-runs-test` joins the native lanes' target array and the `lane-targets` mirror in `scripts/affected-targets`, narrowed like every other entry: it runs only when the compiler, the rules, or the fixtures are impacted — which is exactly when it matters — and it is now the one place CI compiles *and executes* an internally-linked `rue_program` on both aarch64 platforms.

## 3. The stale platform note I'd been quoting around

`toolchains/BUCK` claimed macOS compilation "requires Xcode Command Line Tools (uses system clang)". That predates internal Mach-O linking and is exactly the claim the `--linker internal` pin relies on being false. Corrected in place rather than left to mislead the next reader — the same class of fix as the `AGENTS.md` `--prefer-remote` retirement.

## Smaller

- `extra_scan_inputs` moves off `rue_program`'s public attrs onto the boundary-control rule alone: ordinary programs no longer carry an attribute whose only purpose is smuggling undeclared files past the scan.
- `docs/process/build-cache.md` now documents `rue_program` as the second cacheable writer class after the corpus actions — including the property worth knowing for cache forensics: a scan/derive cache **miss** still converges to a compile cache **hit**, because the derived manifest is byte-identical across checkout roots and the compile is keyed on the manifest's content, not the envelope's.

Full sweep green: `repository-quality-gates`, `ci-gate-validation`, `ci-required-results-tool-tests`, all fixture tests, derive unit tests, `test-affected-targets` (88 checks), tier validation, and three consecutive digest-control passes.

At this point I have nothing further I'd block a merge on. The two things CI will answer that I can't from here: whether the BuildBuddy worker image has `python3` (control 2's derive step; fails loudly in the merge queue if not), and the digest control's first runs on real runners.

---

DorianListens:
Retiring disclosed item #7 (python3 on the BuildBuddy worker, "cannot verify from here") — it's now verified, and the answer is worth recording because the obvious reasoning is wrong in one direction and rescued by a different fact.

**The hermetic toolchain does not provide Python.** What's hermetic is the *Rust* toolchain (pinned, downloaded rustc/rustfmt/clippy). Python is explicitly the system's: `toolchains/BUCK` configures `system_python_bootstrap_toolchain`, whose interpreter is literally the string `"python3"` resolved from the executing machine's `PATH` (`prelude/toolchains/python.bzl`, `_INTERPRETER`). And the rue_program scripts don't even route through that toolchain — they're `sh_binary`s whose `.py` mains resolve `#!/usr/bin/env python3` the same way. So "we build the toolchain hermetically with buck" guarantees nothing here.

**But the requirement is already load-bearing in required CI.** Every rustc compile action in buck2's prelude executes through `prelude//rust/tools:rustc_action` — a `python_bootstrap_binary` that runs `python3 rustc_action.py … rustc …` on whatever machine executes the action (`rust/tools/BUCK:40-45`, `rust/build.bzl`). The merge-group RE canary builds `//crates/rue:rue` remotely with no local fallback, which is hundreds of rustc actions, each one launched by `python3` on the pinned worker image — and that job is green on trunk in every merge group. The derive script therefore adds **zero new worker requirements**: it asks for exactly the interpreter, by exactly the resolution mechanism, that the existing remote compiler build already exercises continuously.

Control 2 carries no python risk. The remaining CI-only unknown is just the digest control's first runs on real runners.

---

PR #2278, opened 2026-08-12
Gate @syscall behind checked blocks

DorianListens:
Fixes RUE-1369.

`@syscall` was excluded from the shared checked-block gate in semantic analysis, so it compiled anywhere. Both the §4.13 intrinsic table ("Unchecked intrinsics — only valid inside a `checked` block") and §9.2's prose already described it as unchecked; only `analyze_syscall_intrinsic`'s doc comment claimed the requirement, with nothing enforcing it. The legality rule that enumerates the gated intrinsics (9.1:12) omitted it too, and no case asserted the diagnostic.

## Compiler

`@syscall` joins the shared gate in `analyze_intrinsic_impl` rather than getting a second gate inside `analyze_syscall_intrinsic`, so it reports through the one canonical path with the same wrap-in-`checked` help as the raw-pointer and allocation intrinsics:

```
error[E1300]: syscall intrinsic `@syscall` requires a `checked` block
help: wrap the operation in a `checked { ... }` block
```

The stale "`@syscall` has its own gating" comment and the unenforced "Requires a checked block" doc line are gone.

## Spec

New legality rule 9.2:3a, placed with the other `@syscall` rules: `@syscall` is not a raw-pointer intrinsic, so it does not belong inside 9.1:12's enumeration. 9.1:12 now cross-references it alongside the allocation family (9.2:13) and the raw-byte intrinsics (9.2:14h, 9.2:14l). The §4.13 table needed no change — the compiler was the outlier.

## Coverage

- Spec: `syscall_outside_checked_rejected` and `syscall_in_checked_accepted`, both citing 9.2:3a. The accepted case is `compile_only` because every executable syscall number is platform-specific and the point is the gate.
- UI: `diagnostics/unchecked_syscall.toml` pins the diagnostic text, including that an `unchecked fn` body is not itself a checked context — it marks the *call* as requiring `checked`, so unchecked operations in its body still need their own block.

## Fallout

The two `cli.intrinsic_args` cases probing stray type arguments (`@syscall(a, (), b)`, `@syscall(a, i32)`) called `@syscall` bare, so the new gate fired before their E0702 argument diagnostic. They are wrapped in `checked` now, with the reason in the file header; they still test what they were written to test.

## Verification

`scripts/rue spec` (2335 passed), `scripts/rue ui` (248), `scripts/rue cli` (1848), `scripts/rue quick` (31), `//:spec-traceability` (legality-rule coverage 100%, 156/156), `//:repository-quality-gates`, and the rue-air fmt-check and clippy gates — all green locally on x86-64 Linux.

---
_Generated by [Claude Code](https://claude.ai/code/session_015M58xCKk7oc2VpD3SZgi21)_

### DIFF
--- a/crates/rue-air/src/sema/analysis/intrinsics.rs
+++ b/crates/rue-air/src/sema/analysis/intrinsics.rs
@@ -247,14 +247,13 @@ impl<H: OrdinaryBodyAnalysisHost> OrdinaryBodyEngine<'_, H> {
             .collect();
         let known = &self.known_symbols();
 
-        // Raw-pointer and heap intrinsics are unchecked operations: they may
-        // only be used inside a `checked` block (spec 9.1:1, chapter 9). Gate
-        // them before the per-intrinsic dispatch so every one shares a single
-        // diagnostic. `@syscall` has its own gating; the set here is
-        // @raw/@raw_mut/@field_ptr/@ptr_read/@ptr_write/@ptr_offset/
-        // @ptr_to_int/@int_to_ptr plus the unified byte allocation family
-        // @alloc/@alloc_zeroed/@free/@realloc/@resize (RUE-1, RUE-301,
-        // ADR-0059 Phase 3).
+        // Raw-pointer, heap, and syscall intrinsics are unchecked operations:
+        // they may only be used inside a `checked` block (spec 9.1:1, chapter
+        // 9). Gate them before the per-intrinsic dispatch so every one shares a
+        // single diagnostic. The set is @raw/@raw_mut/@field_ptr/@ptr_read/
+        // @ptr_write/@ptr_offset/@ptr_to_int/@int_to_ptr, the unified byte
+        // allocation family @alloc/@alloc_zeroed/@free/@realloc/@resize, and
+        // @syscall (RUE-1, RUE-301, RUE-1369, ADR-0059 Phase 3).
         if ctx.checked_depth == 0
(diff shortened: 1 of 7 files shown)


DorianListens:
Dequeued on a runner infrastructure flake, not this diff — no fix to push.

`test (linux-x64-cli-shard-1)` in merge-group run [31558370155](https://github.com/rue-language/rue/actions/runs/31558370155) died before any test ran, in dotslash fetching the buck2 binary:

```
dotslash error: problem with `/home/runner/work/rue/rue/buck2-bin`
caused by: failed to fetch `https://github.com/facebook/buck2/releases/download/2026-07-15/buck2-x86_64-unknown-linux-musl.zst`
     1: curl: (60) SSL certificate problem: self-signed certificate
```

`CI success` then failed only downstream of it (`platform-corpus: expected success, got 'failure'`); every other job in the batch — premerge, native linux-arm64 and macos-arm64, spec, the other three CLI shards, oracle-diff, valgrind, asan, release, reproducibility — passed.

Evidence it is transient rather than environmental: the same `test (linux-x64-cli-shard-1)` job passed in #2279's merge-group run [31558583590](https://github.com/rue-language/rue/actions/runs/31558583590) four minutes later, on the queue branch stacked on this PR's own merged result. The PR-level run for this head (`041f88e`) was also fully green.

Re-queueing should be all this needs.

---

PR #2319, opened 2026-08-12
Keep Linear magic words out of PR citations

DorianListens:
AGENTS.md said `refs`, `part of`, and friends "link for context, no status
change." They do not. Per [Linear's GitHub docs](https://linear.app/docs/github):

> When you use a _non-closing_ magic word, the linked PR or commit will still
> move the issue through other statuses per your workflow settings, but it will
> not apply the _On PR or commit merge_ status when the PR or commit merges.

> \[B\]y default, we will move linked issues to "In Progress" when PRs are open
> and "Done" when PRs merge.

A non-closing word opts out of the *merge* step only, so `Refs RUE-NN` still
moves the issue to In Progress at PR open — including an issue already Done or
Canceled. A PR opened today cited two background issues with `refs` and pulled
one out of Done and one out of Canceled.

The rewritten section: magic words are commands that change state, used only
for issues the PR should actually move; relationships between issues are
recorded in Linear as `relatedTo` / `blockedBy` / parent instead. It also
corrects the word classes to the documented three — `relates to` / `related to`
are relation words with no status change, distinct from the non-closing words
that do move status — and notes that a bare ID does not link, that the branch
name links on its own, and that `skip`/`ignore` suppresses an unwanted link.

Documentation only; no code or test changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -304,99 +304,47 @@ parallel Markdown backlog.
   linked with `relatedTo` or a real `blockedBy` dependency.
 - Priorities are Urgent, High, Medium, and Low. Use `bug`, `feature`, `task`, or
   `chore` labels consistently.
-- A PR body uses magic words deliberately: a closing word for each issue that
-  PR completes, and only when the issue is actually done. Relationships between
-  issues are recorded in Linear, never by citing an ID in a PR — see "Linear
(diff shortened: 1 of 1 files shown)


DorianListens:
@steveklabnik trying to make managing linear issues a little easier. We were using `refs` wrong.

PR #2327, opened 2026-08-12
RUE-1470: accept release-quality compiler performance contract

steveklabnik:
Fixes RUE-1470.

## Summary

- accept Rue's target as the canonical fresh-process `-O3` source-to-native pipeline
- set a 250 ms one-worker target, a 150 ms automatic-worker stretch, and a provisional 128 MiB RSS envelope
- use the versioned `x86_64-linux` `ubuntu-24.04` GitHub Actions collection epoch as the absolute reference regime
- establish the first limit at the measured baseline, ratchet it downward with accepted improvements, and make 250 ms the permanent ceiling once reached
- make the measurement boundary fail closed through independent runner, manifest, and compiler evidence plus structural tests against peer pipelines
- require profile-driven critical-path and semantic-to-CFG ownership review before broad architectural changes
- stage vertical 500 ms, 350 ms, and 250 ms milestones
- add performance admission requirements for future compiler and language features
- explicitly defer precompiled package/standard artifacts, persistent hosts, and incremental linking

## Current evidence

The controlled historical Lattice profile is 843.30 ms compiler-root at one worker, but the maintained scaling invocation uses Rue's default `-O0`. A six-pair current-trunk diagnostic found `-O3` compiler-root time 1.17% above `-O0`, a 17.20% smaller executable, and 4.02% higher peak RSS; host contention makes those absolute times non-baseline evidence. ADR Phase 1 therefore pins and measures the external `-O3` boundary rather than relabeling the existing series.

## Accepted maintainer decisions

1. `-O3` process-spawn-through-native-output is the indivisible product boundary.
2. The target is 250 ms with one worker, with 150 ms as the automatic-worker stretch.
3. The 128 MiB envelope remains provisional through the first attribution milestone.
4. Runner, manifest, compiler, and structural evidence enforce the boundary and the single canonical pipeline.
5. Phase 1 establishes the GitHub Actions baseline and public Lattice dashboard; accepted improvements ratchet the limit to a permanent 250 ms ceiling.

## Tracking

Implementation phases are filed as RUE-1475, RUE-1474, RUE-1473, RUE-1471, and RUE-1472. RUE-1478 tracks the frozen Lattice workload and public target dashboard folded into Phase 1.

## Validation

- `scripts/validate-adrs.py --write`
- `scripts/validate-adrs.py`

### DIFF
--- a/docs/designs/0071-release-quality-compiler-performance-contract.md
+++ b/docs/designs/0071-release-quality-compiler-performance-contract.md
@@ -1,0 +1,487 @@
+---
+id: 0071
+title: "Release-quality compiler performance contract"
+status: accepted
+tags: [architecture, compiler, performance, process]
+feature-flag: null
+created: 2026-08-12
+accepted: 2026-08-12
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["RUE-1470", "RUE-1478", "ADR-0063", "ADR-0067", "ADR-0068"]
+---
+
+# ADR-0071: Release-quality compiler performance contract
+
+## Status
+
+Accepted by Steve Klabnik and Dorian Taylor on 2026-08-12. This ADR sets a
+product target and the rules for pursuing it. It does not replace the compiler
+architecture in ADR-0063 or the measurement protocols in ADR-0067 and
+ADR-0068. It turns their implemented mechanisms into an explicit performance
+contract.
+
+## Summary
+
+Rue will optimize one canonical, release-quality source-to-native compiler
+pipeline toward a 250 ms fresh-process build of the fixed Lattice reference
+workload in a versioned GitHub Actions reference regime. A one-worker run is
+the primary architecture target; an automatic-worker run at or below 150 ms is
+a secondary stretch target in the same regime. The target pins Rue `-O3`; it
+includes source discovery, parsing, semantic analysis, CFG construction and
+optimization, backend work, object generation, linking, and output publication.
+It cannot be met by weakening generated-code quality, introducing a separate
+fast compiler, moving work outside the measured boundary, or assuming
+precompiled package artifacts.
+
+Wall time is the user-visible outcome. Retired instructions, allocations,
+resident memory, phase accounting, and deterministic compiler-work counters
+explain why it moved and protect algorithmic scaling when clock noise cannot.
+The reference runner and compiler publish complementary, machine-checked
+evidence that the complete boundary was actually measured; a sample that used
+undeclared retained state, cached artifacts, a daemon, or a different codegen
+policy is invalid rather than merely annotated.
+Work proceeds as vertical, profile-driven milestones: first make the current
+critical path and worker scaling explicit, then review repeated ownership across
+semantic analysis and CFG construction, then drive the complete pipeline
+through 500 ms, 350 ms, and 250 ms milestones without regressing correctness,
+memory, output, or incremental locality.
+
+## Accepted maintainer decisions
+
+1. Rue `-O3`, from process spawn through successful native output and exit, is
+   the indivisible product boundary.
+2. The primary target is 250 ms with one worker. A 150 ms automatic-worker run
+   is a stretch target rather than a prerequisite.
+3. The 128 MiB envelope remains provisional through the first attribution
+   milestone.
+4. Runner, manifest, and compiler evidence jointly enforce the measured
+   boundary. Structural repository tests additionally prohibit peer pipelines
+   and undeclared ways to move work outside it.
+5. Phase 1 measures the GitHub Actions reference regime, sets the first limit
+   at the observed baseline, and publishes it. Accepted improvements ratchet
+   that limit downward until it reaches 250 ms; after that, 250 ms is the
+   permanent non-regression ceiling for the pinned regime.
+
+## Context
+
+Rue deliberately built incremental and parallel compiler architecture before
+setting an absolute speed target. ADR-0063 established one revisioned query
+graph through object projection, per-body semantic/CFG/codegen boundaries,
+immutable canonical artifacts, and a fresh-link seam. ADR-0067 established
+honest wall-clock phase accounting and versioned performance observations.
+ADR-0068 established retained-session edit measurement. The post-ADR-0063
+architecture audit confirms that these are the live paths; there is no peer
+frontend or coarse shared semantic owner to remove.
+
+That work has paid off, but relative improvement alone does not answer whether
+the compiler is fast enough. On the maintained Lattice workload, the tracked
+one-worker release-built compiler moved from 2,042.64 ms at RUE-1351's baseline
+to a median 843.30 ms after RUE-1467, a 2.42x improvement. RUE-1468 then removed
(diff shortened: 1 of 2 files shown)


docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+semantic analysis and CFG construction, then drive the complete pipeline
+through 500 ms, 350 ms, and 250 ms milestones without regressing correctness,
+memory, output, or incremental locality.
+
+## Maintainer review focus
+
+Acceptance asks the maintainers to decide five things:
+
+1. Is Rue `-O3`, from process spawn through successful native output and exit,
+   the right indivisible product boundary?
```
steveklabnik: Yes, I think we want to explicitly test our most heavy compilation process instead of our lightest. There's not much at `-O3` today, but there will be someday.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+2. Is 250 ms with one worker an appropriate aggressive target, with 150 ms
+   automatic-worker performance as a stretch rather than a prerequisite?
```
steveklabnik: Given our research, I think it should be, but if we can't make it for some reason after putting work in, we can reassess.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+3. Should 128 MiB remain provisional through the first attribution milestone?
```
steveklabnik: Yes. More below.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+4. Is the runner/manifest/compiler evidence model sufficient to prevent hidden
+   caches, artifacts, daemons, alternate pipelines, or optimization downgrades?
```
steveklabnik: I would hope so, but we also probably need some sort of structural testing.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+5. Should the first absolute gate wait for the reference regime and 500 ms
+   milestone, while CI continues to gate structure, correctness, and relative
+   regressions?
```
steveklabnik: I think we should 

1. take stock of where we are
2. set the limits to there
3. ratchet down over time

We can stop the ratchet once we hit the goal.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+  decision system rather than a collection of interesting numbers.
+- Language design incorporates compile-time complexity while Rue is still
+  young enough to change representations and ownership cleanly.
+- Future package and incremental systems inherit explicit artifact boundaries
+  without becoming prerequisites for acceptable first-build performance.
+
+### Negative
+
+- A 250 ms optimized-native target is deliberately difficult and may expose
+  architectural work larger than the recent bounded optimizations.
```
steveklabnik: This is a pro, not a con. If we need different architecture, now is the time to do it.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+- The reference host is useful for absolute accountability but does not by
+  itself make a portable performance claim.
+- Critical-path and resident-memory attribution add measurement complexity and
+  must be kept out of hot shared paths.
+- Rejecting clock or memory regressions narrows the space for speculative
+  refactors; larger experiments need isolated prototypes and explicit review.
```
steveklabnik: I think this is also okay given that performance is something we are explicitly prioritizing.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+- A neutral clock result remains publishable when it is an exact work,
+  complexity, ownership, or maintainability win with no other regression.
+
+## Open Questions
+
+1. Should the 128 MiB figure become a hard target now, or remain provisional
+   until the first milestone reconciles allocator RSS with exact retained
+   charges?
+2. Should the initial reference be the exact maintainer host or a separately
+   provisioned, long-lived machine with a tighter fingerprint policy?
```
steveklabnik: We don't really have the ability to provision a different machine, so that is tricky. We might need to have different numbers on cloud instances.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+3. Is 150 ms the right automatic-worker stretch target, or should it be derived
+   only after the first complete 1/2/4/8/automatic scaling report?
+4. At which milestone should an absolute target become a release gate rather
+   than a reviewed project objective?
```
steveklabnik: Once the ratchet is complete, we should not regress.

docs/designs/0071-release-quality-compiler-performance-contract.md
DorianListens: Yeah this is something that stuck out to me as well. We should at least name the host explicitly. It would be nice if we could do this on actions machines, as opposed to yours, just for regularity, but yours is fine too.

docs/designs/0071-release-quality-compiler-performance-contract.md
```diff
@@ -0,0 +1,473 @@
...
+- Critical-path and resident-memory attribution add measurement complexity and
+  must be kept out of hot shared paths.
+- Rejecting clock or memory regressions narrows the space for speculative
+  refactors; larger experiments need isolated prototypes and explicit review.
+
+### Neutral
+
+- ADR-0063's query architecture remains in force. This ADR may cause a later
+  architecture decision, but current evidence does not justify one now.
+- ADR-0067's dashboard and ADR-0068's warm-edit suite continue to answer their
```
DorianListens, line 460: I think we should fold this goal into the perf dashboard on the website, ideally. A big line chart with a target on it will be nice!

If nothing else, let's add lattice to the dashboard as a workload asap.

docs/designs/0071-release-quality-compiler-performance-contract.md
steveklabnik: yeah, id much rather do it on actions

docs/designs/0071-release-quality-compiler-performance-contract.md
DorianListens: Filed as RUE-1478: frozen Lattice copy under performance/workloads/ (same fixture revision Phase 1 pins), new suite revision + epochs with sampling re-derived from measured cost, and the 250 ms target rendered as an annotation with a reference-regime caveat. Might be worth folding into Phase 1 acceptance so the chart goes live with the first -O3 observations.

PR #2338, opened 2026-08-13
RUE-1473: share durable callable parameters

steveklabnik:
## Summary
- use one canonical durable parameter vocabulary across rue-compiler and rue-air
- share retained function and method parameter slices instead of rebuilding them per materialization
- update the deterministic payload partition and pin pointer identity with a focused regression

## Evidence
- Lattice moves another 8,066 durable materialization requests to shared storage, for 18,773 / 19,211 shared requests (97.7%)
- `scripts/rue unit compiler durable_function_materialization_shares_the_canonical_parameter_payload`
- `scripts/rue quick`

Refs RUE-1473.

### DIFF
--- a/crates/rue-air/src/sema/binding_manifest.rs
+++ b/crates/rue-air/src/sema/binding_manifest.rs
@@ -116,7 +116,7 @@ pub enum SemanticExportType {
     Module(Arc<str>),
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum SemanticParameterMode {
     Value,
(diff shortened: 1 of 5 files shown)


steveklabnik:
Hosted scaling run: https://github.com/rue-language/rue/actions/runs/31664428481

Authoritative structural result:

- shared durable payload materializations: **18,773 / 19,211 (97.7%)**
- owned durable payload materializations: **438 / 19,211 (2.3%)**
- all other deterministic work counters are unchanged
- peak memory is effectively flat

Clock medians moved down, but this run used an Intel Xeon Platinum 8573C while the immediate baseline used an AMD EPYC 7763 and an older runner image. Those clock deltas are advisory only; I am recording this as exact work reduction with no detected regression, not as a causal wall-time percentage.

PR #2340, opened 2026-08-13
RUE-1473: profile body-local materialization

steveklabnik:
## Summary
- add a bounded nested duration distribution for fresh body-local semantic epoch construction and body import
- publish total/max materialization time beside semantic, CFG construction, and CFG optimization histograms
- keep collection on the existing thread-local timing path with no new shared hot-path writes

## Evidence
- `scripts/rue quick`

Refs RUE-1473.

### DIFF
--- a/crates/rue-bench/src/measure.rs
+++ b/crates/rue-bench/src/measure.rs
@@ -397,7 +397,7 @@ fn run_once(request: &SampleRequest<'_>) -> Result<CompletedCompile, String> {
                 "boundary policy has no independently declared target".to_string()
             })?;
             evidence
-                .validate_against(policy, expected_target)
+                .validate_current_producer_against(policy, expected_target)
                 .map_err(|detail| format!("build-boundary evidence rejected: {detail}"))?;
             Some(evidence)
         }
--- a/crates/rue-bench/src/scaling.rs
+++ b/crates/rue-bench/src/scaling.rs
@@ -366,8 +366,10 @@ fn render(report: &ScalingReport) -> String {
 
     out.push_str("\n## Worker scaling and critical path\n\n");
     out.push_str("Utilization divides summed query-worker active time by compiler-root time and the compiler's resolved worker count. Ready wait is summed across dependency-ready items, so the mean and maximum are the directly comparable latency signals. Body columns show total/max milliseconds from bounded compiler histograms. The rooted-acquisition envelope is inclusive: it contains the semantic attempt used to discover a trusted-toolchain park, is not an exclusive phase, and must not be added to semantic time or read as filesystem cost.\n\n");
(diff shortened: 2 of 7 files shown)


steveklabnik:
Hosted scaling profile: https://github.com/rue-language/rue/actions/runs/31666434288

The new same-process nested distribution makes the next decision unambiguous. Fresh body-local semantic epoch construction plus body import accounts for essentially all measured CFG-construction body time:

- lattice, one worker: 587.49 ms local epoch / 593.24 ms CFG construction = 99.0%
- lattice, automatic (4 workers): 978.26 ms / 987.23 ms = 99.1%
- harbor, one worker: 498.78 ms / 507.26 ms = 98.3%
- mosaic, one worker: 209.33 ms / 212.80 ms = 98.4%
- ruelex, one worker: 60.05 ms / 61.18 ms = 98.2%

The deterministic counters show 1,280 Lattice epochs selecting 55,307 root facts (43.21 per epoch), so this is repeated small-epoch work rather than a hidden whole-program closure. The next step after landing this measurement is to separate fixed epoch construction from body import, then remove the dominant repeated setup while preserving body-local invalidation and immutable ownership. Cross-run wall-clock comparisons are not needed for this conclusion; numerator and denominator come from the same processes.

steveklabnik:
CI repair update: the first failure exposed a missing additive serde default. After adding it, the repository-wide stall gate showed why recent stored runs still did not publish: historical records correctly defaulted the new distribution to empty, but generic stored-evidence validation required it to be populated.

The fix now separates historical and current-producer contracts:

- immutable historical run objects may omit this new additive diagnostic;
- the current measurement runner still rejects a compiler that omits it.

Local reproduction against the fetched `performance-data-v1` branch now derives 892 points with **0 rejected runs** (previously 21), and `validate-performance-stall.py` reports all three platforms current at `b7639d2d`.

steveklabnik:
Correction after the post-merge performance gate and #2342 follow-up profile: the original nested-span measurement materially perturbed the code it measured. The +2.85% x86 ratchet failure exposed that observer effect.

With the lower-overhead direct-duration event on the same maintained scaling suite, Lattice one-worker reports 169.84 ms local materialization inside 587.94 ms CFG construction = **28.9%**, not the 99.0% I reported above. Automatic workers report 268.51 / 975.86 ms = **27.5%**. The earlier nested span itself added enough per-body lifecycle/locking/name-recording work to dominate its measured interval.

The durable conclusion is therefore: local epoch materialization is still a significant target, but it is not essentially the whole CFG-construction cost. The next profile must also explain the remaining ~418 ms one-worker CFG-construction body time before we choose an ownership refactor.

steveklabnik:
Further precision on the correction above: the dominant error was lexical scope, not proof that span machinery itself cost ~418 ms. The `_materialization_span` guard was not dropped after the materialization call; Rust kept it alive until the enclosing CFG query returned. Its inclusive duration therefore covered almost the entire remaining CFG-construction path. #2342 ends the timer immediately after materialization and publishes that exact duration through a direct event. The post-merge +2.85% ratchet failure remains valid evidence that the landed shape regressed on that run, but it does not establish that all of the 418 ms attribution error was instrumentation overhead.

PR #2342, opened 2026-08-13
RUE-1473: fix materialization profiling scope

steveklabnik:
## Summary

- replace the per-body nested tracing span added by #2340 with one direct duration event whose timer ends immediately after materialization
- accumulate that event into the existing worker-local bounded distribution
- preserve the same `semantic_materialization_bodies` evidence without extending the measurement through the rest of CFG construction

## Evidence

The original span guard lived until the enclosing CFG query returned, so its 99% share accidentally included domain projection, prerequisite queries, CFG building, and publication. The post-merge x86 collector also measured Lattice at 3,108,599,906 ns against the 3,022,383,520 ns non-regression ratchet (+2.85%).

The corrected hosted profile records the exact lexical interval:

- Lattice one worker: 169.84 ms local materialization / 587.94 ms CFG construction = 28.9%
- Lattice automatic: 268.51 / 975.86 ms = 27.5%
- 1,280 body observations remain present

Local `scripts/rue quick` passes. The merge queue performance collector has passed x86_64-linux and aarch64-linux on the corrected commit; macOS remains the final platform gate.

Refs RUE-1473.

### DIFF
--- a/crates/rue-compiler/src/cfg_query.rs
+++ b/crates/rue-compiler/src/cfg_query.rs
@@ -1027,8 +1027,7 @@ fn materialize_and_build_cfg(
         );
     }
     context.record_work(rue_query::WorkItem::new("cfg.materialize.attempts", 1));
-    let _materialization_span =
-        tracing::info_span!("semantic_materialization", phase = "cfg_and_optimization").entered();
+    let materialization_started = std::time::Instant::now();
     let materialized = match &key.semantic_input {
         CfgSemanticInput::Body { input, .. } => {
             crate::local_semantic_materialization::materialize_canonical_body(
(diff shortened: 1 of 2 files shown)


steveklabnik:
Hosted corrective profile: https://github.com/rue-language/rue/actions/runs/31668342513

The direct-duration event preserves all 1,280 Lattice observations while removing the nested tracing span's observer effect. In this run:

- Lattice one worker: local materialization 169.84 ms / CFG construction 587.94 ms = 28.9%
- Lattice automatic (4 workers): 268.51 / 975.86 ms = 27.5%
- process time: 2,864.15 ms one-worker; 2,103.84 ms automatic

The prior span-based profile reported 587.49 / 593.24 ms and the post-merge ratchet measured 3,108.60 ms versus a 3,022.38 ms ceiling. This proves that the old probe materially distorted its own target. Cross-run clock deltas remain advisory because the hosted image revisions differ, but the corrected within-run shares are sound. The next optimization decision should explain the remaining ~418 ms of one-worker CFG-construction time, while treating local epoch construction/import as a meaningful 169.84 ms subtarget.

steveklabnik:
Clarification: the 99% result was primarily a span-lifetime bug. The guard lived to the end of the CFG query, so it measured materialization plus the rest of CFG construction. The direct event both ends at the correct lexical boundary and avoids unnecessary per-body span lifecycle work. The corrected 28.9% one-worker share is the sound architectural input.

PR #2354, opened 2026-08-13
RUE-1473: reuse canonical body signatures

steveklabnik:
Provider-owned body analysis already resolves the selected callable parameters into the body request’s canonical type and parameter authority. Reuse those exact parameter facts when invoking the ordinary expression engine instead of resolving the RIR parameter spellings a second time.

The body’s declared return spelling remains body-local authority: call-site signatures may intentionally normalize coercible types such as `Str(N)` to `str`, so body analysis still resolves that exact return spelling once.

This removes redundant nominal registration and type construction from both free-function and named-method bodies. It also deletes the obsolete syntax-resolving named-method facade, preserves diagnostic ordering for the remaining syntax-backed test/oracle path, and adds a structural guard that prevents provider bodies from returning to that path.

The deterministic provider-work assertion drops from 16 to 15 name lookups for the representative named-import case.

Tests:
- `scripts/rue spec str_fixed`

Refs RUE-1473.

### DIFF
--- a/crates/rue-air/src/sema/consistency_tests.rs
+++ b/crates/rue-air/src/sema/consistency_tests.rs
@@ -285,6 +285,8 @@ mod tests {
         assert!(engine.contains("pub(crate) trait OrdinaryBodyAnalysisHost"));
         assert!(engine.contains("pub(crate) struct OrdinaryBodyEngine"));
         assert!(engine.contains("pub(crate) fn analyze_single_function"));
+        assert!(engine.contains("pub(crate) fn analyze_single_function_resolved"));
+        assert!(engine.contains("pub(crate) fn analyze_named_method_resolved"));
         assert!(engine.contains("pub(crate) fn analyze_function_internal"));
         assert!(!engine.contains("#[allow(dead_code)]"));
         assert!(engine.contains("semantic_type_syntax_compile_error("));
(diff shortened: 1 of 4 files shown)


steveklabnik:
Corrected exact-head scaling evidence: [run 31689272351](https://github.com/rue-language/rue/actions/runs/31689272351), head `02ed43683a781faf540de250b56c4fa34b06e937`. Compared with the immediate parent baseline from [run 31686583195](https://github.com/rue-language/rue/actions/runs/31686583195):

- Lattice One→1 process: 2460.63 → 2364.77 ms (-3.9%)
- Lattice One→1 compiler: 2297.97 → 2200.69 ms (-4.2%)
- Lattice Automatic→4 compiler: 1888.33 → 1818.97 ms (-3.7%)
- Provider analysis One→1: 538.65 → 490.21 ms (-9.0%)
- Provider expression engine One→1: 453.11 → 406.53 ms (-10.3%)
- Provider name lookups: 61,474 → 58,893 (-2,581 / -4.2%)
- Peak RSS One→1: 295.8 → 295.3 MiB (neutral)
- Output size and digest remain unchanged.

The compiler-time improvement is present across every maintained Lattice worker row; this is a cold-time and deterministic-work win, not a scheduling artifact.

PR #2355, opened 2026-08-13
RUE-1473: attribute expression engine phases

steveklabnik:
## Summary

- split the canonical expression-analysis core into five bounded timing distributions
- require complete per-body evidence in the current benchmark producer
- expose the breakdown in scaling reports and advance the report schema to v18

This distinguishes constraint construction and compile-time precomputation from unification, so ADR-0071 optimization work can target measured costs. The intervals are additive evidence only and do not create a second compilation path.

## Validation

- `scripts/rue fmt`
- `scripts/rue quick` (31 targets passed)
- release Lattice boundary proof: all five distributions recorded 1,263 bodies, exactly matching the enclosing expression-engine count

Refs RUE-1473.

### DIFF
--- a/crates/rue-air/src/sema/analysis/type_inference.rs
+++ b/crates/rue-air/src/sema/analysis/type_inference.rs
@@ -6,6 +6,18 @@
 use super::super::ordinary_engine::{OrdinaryBodyAnalysisHost, OrdinaryBodyEngine};
 use super::*;
 use crate::inference::LazyInferenceFacts;
+use std::time::Instant;
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct InferenceBreakdown {
+    pub(crate) precompute_ns: u64,
+    pub(crate) constraint_generation_ns: u64,
+    pub(crate) unification_resolution_ns: u64,
+}
+
+fn elapsed_ns(started: Instant) -> u64 {
+    u64::try_from(started.elapsed().as_nanos()).unwrap_or(u64::MAX)
+}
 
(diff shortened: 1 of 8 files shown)


steveklabnik:
The first scaling run (31691889479) rejected the original publication shape: adding a second per-body tracing event moved maintained workload medians by roughly 5–6% (for example Lattice one-worker compiler time 2200.69 → 2332.50 ms). Auto-merge was disabled before integration.

The updated commit preserves the five bounded clocks but folds their fields into the existing provider-completion event, so it publishes no additional hot-path tracing event. Quick tests pass. I am rerunning the full scaling matrix and will re-enable the queue only if that evidence is neutral.

steveklabnik:
Revised scaling evidence: https://github.com/rue-language/rue/actions/runs/31692797549

The redesigned single-event publisher completed the full matrix with identical workload shapes, deterministic compiler work, and output sizes. The runner fingerprint changed again (AMD EPYC 7763 versus the earlier EPYC 9V74), so clock comparisons are advisory rather than a ratchet decision. They show no consistent regression: one-worker compiler medians versus the pre-instrumentation report were Ruelex +2.1%, Mosaic -0.7%, Harbor -0.8%, and Lattice -2.0%; Lattice peak RSS was 295.2 versus 295.3 MiB.

The first apparent 5–6% movement was therefore host-correlated, not proof of code overhead. The revision still removes the extra event on structural grounds: it adds only bounded timestamp reads and carries the five fields on the provider event that already existed. The current report records complete matching counts and identifies one-worker Lattice expression cost as 90.0 ms precompute, 134.7 ms constraint generation, 7.6 ms unification/projection, and 51.6 ms AIR emission/validation.

PR #2356, opened 2026-08-13
RUE-1473: scope inline constructor precompute to its body

steveklabnik:
## Summary

- traverse only RIR reachable from the requested body when discovering inline constructor heads
- stop at nested declaration owners so unrelated functions and types are not re-scanned
- add a regression proving each body sees only its own constructor head

## Why

The module RIR contains every declaration in a source file. Scanning the complete arena once per compiled body multiplied unrelated work by the number of declarations. This keeps the same canonical semantic path while making the precompute pass proportional to the requested body.

## Validation

- `scripts/rue unit air inline_constructor_precompute_candidates_are_scoped_to_the_requested_body`
- `scripts/rue spec 4.14`
- `scripts/rue quick`

Refs RUE-1473.

### DIFF
--- a/crates/rue-air/src/sema/analysis/type_inference.rs
+++ b/crates/rue-air/src/sema/analysis/type_inference.rs
@@ -317,8 +317,12 @@ impl<H: OrdinaryBodyAnalysisHost> OrdinaryBodyEngine<'_, H> {
         // constrained and an integer payload literal defaulted to `i32`
         // (RUE-599). Runs before the lazy-method collection below so methods
         // registered while reducing a head are included in it.
-        let inline_ctor_head_types =
-            self.precompute_inline_ctor_head_types(type_subst, value_subst, &comptime_local_types);
+        let inline_ctor_head_types = self.precompute_inline_ctor_head_types(
+            body,
+            type_subst,
+            value_subst,
+            &comptime_local_types,
+        );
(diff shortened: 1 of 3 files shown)


steveklabnik:
Performance evidence: [revised scaling run](https://github.com/rue-language/rue/actions/runs/31695498523), with [the first branch run](https://github.com/rue-language/rue/actions/runs/31694179241) and [a contemporaneous trunk control](https://github.com/rue-language/rue/actions/runs/31694829306).

The trunk control landed on a different CPU family, so it is not an authoritative clock comparison. The two branch runs used the same EPYC 9V74 CPU family. After removing the per-body visited hash table, the targeted one-worker inference-precompute interval changed as follows: Ruelex 2.61 → 2.22 ms (-15%), Mosaic 21.63 → 20.71 ms (-4%), Harbor 70.64 → 70.82 ms (flat), and Lattice 96.41 → 95.55 ms (-1%). Whole-compiler clocks and neighboring phases moved inconsistently, so this is not a claimed latency win on the maintained suite.

The deterministic architectural result is the reason to land: candidate discovery is now bounded by the requested body rather than scanning the complete module arena once per body, and the regression proves unrelated bodies cannot enter the candidate set. Output shapes, binary sizes, deterministic compiler work, and memory remain unchanged. This is a neutral maintained-workload result with a stricter asymptotic bound.

PR #2357, opened 2026-08-13
RUE-1473: remove dormant inference dependency logging

steveklabnik:
## Summary

- stop building a per-body dependency log during inference constraint generation
- remove the unconditional clone of that log before unification
- delete the unused recovery-state plumbing that was its only consumer

Every current body-analysis host disables this recovery experiment, and no production or test path enables it. The actual constraints, unification, semantic dependency recording, and diagnostics are unchanged. This removes allocations and per-expression bookkeeping from successful cold compiles while simplifying the inference boundary.

## Validation

- `scripts/rue quick`
- `scripts/rue spec 4.14` (203/203)
- `scripts/rue fmt`

Refs RUE-1473.

### DIFF
--- a/crates/rue-air/src/inference/generate.rs
+++ b/crates/rue-air/src/inference/generate.rs
@@ -22,128 +22,6 @@ use rue_span::{FileId, Span};
 use std::collections::HashMap;
 use std::rc::Rc;
 
-/// Exact callable selections made while the test-only body-analysis transaction is
-/// generating constraints.  These observations are consumed only when HM
-/// unification fails before AIR analysis can publish its normal dependencies.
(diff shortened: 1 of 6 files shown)


steveklabnik:
Post-merge measurement note: the exact-parent control (run 31697503514, `4ff8dfcb`) and change run (31696710317, `500baedd`) landed on different Azure CPU generations and runner-image epochs, so their wall times are not comparable. The maintained report still confirms identical source shape, output size, and deterministic compiler work. The accepted evidence for this change is therefore the direct code/work reduction—removing the per-body dependency vector, pushes, clone, and dormant recovery machinery—not a clock-time claim.

PR #2359, opened 2026-08-13
RUE-1473: measure rooted CFG local domains

steveklabnik:
## Summary

- publish exact rooted CFG local-domain cardinalities in scaling reports
- cover AIR instructions/payload bytes, type and aggregate domains, symbols, strings, and relocation atoms
- derive the measurements from constant-time container and payload-store accounting without adding query-work allocations or a second domain traversal
- advance the scaling report schema to v19 with backward-compatible defaults

## Evidence

- `scripts/rue quick` (31/31 targets; compiler 848 passed, 6 ignored)
- focused rooted-CFG counter regression
- report rendering and older-schema decoding coverage

Refs RUE-1473.


steveklabnik:
Scaling run 31699342240 successfully publishes the new counters. Lattice has 1,280 rooted local domains containing 52,567 AIR instructions / 326,628 AIR payload bytes, 32,513 type entries, 26,352 aggregate aliases, 8,423 splice type handles, 41,882 interner entries / 10,887,143 UTF-8 bytes, 608 strings, and 650 local atoms. This rules out AIR payload and type cardinality as the dominant retained local-domain bulk; repeated symbol text is the conspicuous payload and a concrete next measurement/architecture target. Source shape, deterministic work outside the new counters, and 1,380 KiB output remain exact. The run used an AMD EPYC 7763, so no cross-host wall-clock claim is made.

PR #2360, opened 2026-08-13
RUE-1473: use fast maps for body endpoint overlays

steveklabnik:
Refs RUE-1473.

Profile-guided cleanup of request-local semantic endpoint registries. Three independent one-worker Lattice samples repeatedly found the standard collision-resistant hasher beneath endpoint registration. These maps support exact lookup only and have no semantic iteration order, so use the compiler's existing AHashMap policy here as well.

Validation:
- scripts/rue quick: 31/31 passed
- compiler unit suite: 848 passed, 6 ignored
- three candidate samples no longer place the standard hasher beneath the named-endpoint registration path

This removes one demonstrated bookkeeping leaf; it makes no broad wall-clock claim.

### DIFF
--- a/crates/rue-air/src/sema/body_endpoint.rs
+++ b/crates/rue-air/src/sema/body_endpoint.rs
@@ -9,6 +9,7 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::sync::Arc;
 
+use ahash::AHashMap;
 use lasso::{Spur, ThreadedRodeo};
 use rue_rir::InstRef;
 use rue_span::FileId;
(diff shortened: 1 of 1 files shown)


steveklabnik:
Scaling workflow 31700743262 passed. Lattice preserves the exact x86-64 output (1,413,120 bytes; SHA-256 b893e76cfabed737b149d0e8c4d8527077dedd17da78418db20a28a7d30885e5) and every deterministic local-domain/materialization counter: 1,280 epochs, 52,567 AIR instructions, 32,513 types, 41,882 interner entries / 10,887,143 UTF-8 bytes, 15,139 selected anonymous nominals, and zero declaration/type rescans. This supports the narrow claim: the PR removes a profiled hasher leaf without changing compiler work or output; hosted clock samples are not used as the claim.

PR #2361, opened 2026-08-13
RUE-1473: compact body-local anonymous symbols

steveklabnik:
Refs RUE-1473.

Body-local semantic epochs currently intern full Debug renderings of canonical anonymous identities. On Lattice, 15,139 selected anonymous nominals become strings hundreds of bytes long, recursively repeating modules, producers, anchors, and specialization arguments in each local epoch.

Use the existing canonical compact anonymous source symbol instead. Mark imported anonymous nominals in the local type pool so the structured anonymous identity and compact spelling remain one representation rather than a qualified compatibility twin. The structured key remains the semantic authority and query equality/hash behavior is unchanged.

Exact one-worker Lattice evidence:
- local interner entries unchanged: 41,882
- local interner UTF-8 bytes: 10,887,143 -> 1,507,142 (-9,380,001; -86.2%)
- all other local-domain counters unchanged
- x86-64 Linux output unchanged: 1,413,120 bytes, SHA-256 b893e76cfabed737b149d0e8c4d8527077dedd17da78418db20a28a7d30885e5

Validation:
- focused AIR anonymous-registration regression passed
- focused distinct-producer symbol acceptance test passed
- scripts/rue quick: 31/31 passed (compiler 848 passed, 6 ignored)

No new shared state, lock, query boundary, scheduling dependency, or invalidation widening is introduced.

### DIFF
--- a/crates/rue-air/src/semantic_import.rs
+++ b/crates/rue-air/src/semantic_import.rs
@@ -1404,6 +1404,7 @@ where
         );
         let mut local_identities = std::collections::BTreeSet::new();
         for nominal in &nominals {
+            let is_anonymous = matches!(&nominal.key, NominalInstanceKey::Anonymous(_));
             let builtin_key = match &nominal.key {
                 NominalInstanceKey::Builtin { kind, name } => Some((
                     name.clone(),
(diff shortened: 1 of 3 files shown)


steveklabnik:
Scaling evidence (run 31702249846) is green on the exact PR head. The Lattice one-worker report confirms the intended structural result: local interner entries remain 41,882 while retained UTF-8 bytes fall from 10,887,143 to 1,507,142 (-9,380,001; -86.2%). Every other local-epoch counter is unchanged, and the x86-64 Linux output remains byte-identical at 1,413,120 bytes with SHA-256 b893e76cfabed737b149d0e8c4d8527077dedd17da78418db20a28a7d30885e5. This is a material memory/work reduction with no semantic or codegen change.

PR #2386, opened 2026-08-14
RUE-1485: cross-tool comparison of gazette against pinned Zola and Hugo

DorianListens:
Closes RUE-1485.

ADR-0072 Phase 5: the first cross-language runtime comparison — gazette against pinned Zola and Hugo building the identical corpus, tool-vs-tool rather than translated microbenchmarks. Plus the runtime-schema generalization deferred from Phase 4, which is what lets gazette be a measured workload at all.

## Peers

Hugo is pinned as a `/hugo` dotslash shim, same shape as `zola` and `tailwindcss`, four platforms. **v0.152.2 rather than the newest release**: from v0.153.0 Hugo ships macOS only as a `.pkg`, which DotSlash cannot extract, and `aarch64-macos` is a row of the matrix. Plain rather than extended, since extended adds Sass and the equivalence subset excludes it.

Each tool builds the corpus with its own idiomatic templates — Tera for Zola, Go templates for Hugo, gazette's engine for Rue — in `performance/ports/`, hashed into the recorded fixture identity because the peer leg is event-driven on that identity.

## Parity was derived by measurement, not by reading documentation

Every config switch is paired in-file with the reason gazette does not do that work, and every observed difference was either configured away or driven into the corpus. Disabled in both peers: highlighting, search index, minification, Sass. Hugo additionally: sitemap, robots.txt, per-section and site feeds, typographer, linkify. Zola additionally: sitemap and robots.txt generation. Feed filename mapping is solved with `[outputFormats.rss] baseName`, so routes are compared rather than excused.

**The file-set allowlist has exactly one entry** — Zola's mandatory `404.html`. Everything else was configured away, which is a stronger check than an excuse. Verified by building all three trees: gazette 130 files, Hugo 130 identical, Zola 131.

Three normalizations are applied to the corpus itself, at fixture-prep time, to **all** tools: `paginate_by` stripped (pagination is an explicit ADR carve-out, and Zola alone emitted `blog/page/1/`), three upper-case filenames lower-cased, and two `load_data` pages excluded. These are disclosed on the published page, not only in source. `preparer_revision` and a digest of the preparer script are inside the recorded identity, so an assembly-rule change opens a new comparable segment rather than silently altering what all three tools consume.

The lower-casing deserves naming precisely: **gazette is the odd tool out.** Zola and Hugo both slugify as the production site does; gazette routes on the path as written. That is a Rue capability gap, and closing it in gazette is the real fix — the normalization is a stopgap, and the ADR, the script, and the config now say so.

## Work equivalence is enforced, and the enforcement was tested by breaking things

Deliberately broken ports were used to confirm the checks fire before any clean run was trusted: dropping `.Content` from Hugo's spec pages failed the oracle on 169 facets; disabling section kinds produced "hugo emits 15 fewer file(s)", "emitted 81 page(s); the corpus has 95", and a named missing section.

That apparatus then caught three real defects that timing alone would never surface:

- **Go's URL escaper** turned all 1,224 rule anchors into `#r-1.1%3a1`, making every one a dead link. Fixed in the shared shortcode with `safeURL`, not per instance.
- **`.PrevInSection`/`.NextInSection` are inverted** relative to Zola's `page.lower`/`page.higher`, so every spec page sent readers backwards. Fixed in the shared partial.
- **Hugo picks layouts by path**, so at scale every duplicated page fell out of `layouts/spec/` and rendered **with no sidebar** — 17,940 bytes against gazette's 22,443 and Zola's 22,630 — while the file set and page count stayed exactly right. Nine tenths of the corpus at 10x, with the port's most expensive template skipped. Found only because the oracle was widened to compare one copy of each page at scale rather than 1x alone. Fixed by translating Zola's `template` key into Hugo's `type`/`layout`.

That third one is the shape of failure this whole validation layer exists for: a peer doing measurably less work while every structural check stays green.

## Thread parity and the canary

`PeerThreadPolicy` lives in the epoch: the primary pins `RAYON_NUM_THREADS=1` and `GOMAXPROCS=1`, the secondary is tool-default, and validation refuses a manifest whose secondary repeats the primary or a canary taken under the non-primary policy. This is not decorative — measured, parallel peers are 2.0x to 2.3x faster, which is most of a published ratio.

A single-threaded Zola build of the workload's own corpus rides every observation as the same-run ratio denominator. A missing canary makes the workload *incomplete* — evidence kept, no ratio published — rather than unappendable.

The full peer leg is event-driven on fixture identity, peer versions, or epoch, detected at fixture-prep time. **The derive step joins** same-run rows from the newest run with the remaining peer configurations from the latest full leg, and only when fixture identity matches; joined rows are marked as such on the page. Without that join the table degraded to gazette-vs-Zola on the first canary-only push, losing Hugo and both default-parallel rows entirely — verified against a real two-run store, and now covered by tests for both the canary-after-full case and the identity-mismatch case.

## Schema generalization

`FixtureDeclaration` is now tagged: the existing `seeded_generator` shape plus a `corpus_tree` shape carrying scale and exclusion set. The single-file recorded-input rule is relaxed **for the corpus-tree shape alone** — a seeded record that borrows the corpus tree's leniency is refused, as are a zero-byte fixture, a zero vocabulary, a slash in `file_name`, a blank generator, and a seeded fixture paired with the new oracle kind. RUE-1046's crafted-record tests all still pass; that tightening was hard-won and is intact.

`OracleKind::SemanticSitePages` decides determinism and the verdict from a per-sample digest of the **emitted tree**. Gazette writes an output tree and prints nothing, so a stdout-reading oracle would have compared two empty streams and passed everything. An empty tree cannot pass.

Gazette and gazette_10x are registered under suite revision 2, epoch 2 per platform; epoch 1 is retired but kept declared so stored records stay valid, which was verified by deriving an epoch-1 record alongside two epoch-2 records with none rejected. 100x is deliberately out of the collected suite as Decision 9's cost valve.

## No ratio is published yet

Local diagnostic medians on `aarch64-macos` are gazette 0.248s, Zola 0.191s, Hugo 0.165s — Rue roughly 1.3x slower than Zola and 1.5x slower than Hugo, the same order of magnitude, which is the question ADR-0072 set out to answer. They are labelled diagnostic and a hosted-runner record is what the published table will use.

An earlier revision quoted gazette at 0.44s. That was a cold first sample promoted by a median that returned the upper of two middles at `--samples 2`, so the warm-up *was* the median. The first build of a run pays page-cache costs for the freshly compiled binary and freshly written fixture; the peers run afterwards and were unaffected, which is why only gazette's figure moved. `median()` now matches `rue_perf_schema::median`, and every sample is printed.

`build_gazette()` also compiled at `-O0` where Decision 5 requires release quality. The manifest always pinned `-O3`; the reference script did not. Measured difference on this workload was null, but the number was not taken under the contract's rule.

## Verification

`scripts/rue unit perf-schema` 203 passed · `scripts/rue unit bench` 121 passed · `scripts/rue quick` 31 targets · `gazette-corpus-diff.py` in `body`, `site`, `golden`, and `peers` modes at 1x and 10x under both thread policies · two full end-to-end harness runs into a store with derive · `rue-oracle-diff` 0 disagreements · clippy and fmt-check clean · all six workflow validators pass.

Not verified: CI on hosted runners — the peer-state cache, `--peer-state-dir`, and the premerge `peers` lane ran only in local equivalents. The end-to-end collection was local with the runner label declared so the full validation path executed; the record was discarded. Linux peer behaviour is untested on that hardware, though the peers are pinned prebuilt binaries and the shim resolves all four platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -386,6 +386,26 @@ jobs:
       - name: Check the gazette template port and its spot goldens
         run: scripts/ci-timed "gazette goldens" -- scripts/gazette-corpus-diff.py golden
 
+      # The cross-tool work-equivalence check (ADR-0072 Decisions 3 and 4,
+      # RUE-1485). Builds the live corpus with gazette, the pinned Zola, and the
+      # pinned Hugo, and asserts the three emit the same file set, the same page
+      # count, and the same normalized page semantics.
+      #
+      # This is a CORRECTNESS lane, not a measurement one: it runs on a shared
+      # CI runner and its timings are discarded, exactly as the ADR's boundary
+      # discipline requires of anything not taken on the pinned regime. What it
+      # exists to catch is drift in the parity configurations and the template
+      # ports — a peer left doing work gazette skips, or a port quietly doing
+      # less — which is the failure that would make the published comparison
+      # dishonest rather than merely wrong, and which nothing else checks before
+      # the numbers are already collected. Two samples per tool, because
+      # determinism needs at least two; the peers' default-parallel row is
+      # skipped because it answers a measurement question, not this one.
+      - name: Check cross-tool work equivalence against pinned Zola and Hugo
+        run: >-
+          scripts/ci-timed "gazette peers" --
+          scripts/gazette-corpus-diff.py peers --samples 2 --no-secondary
+
       - name: Run explicit cross-backend compilation and encoding coverage
         # Both the x86-64 and AArch64 encoders are host-independent Rust tests.
         # Keep this named step visible even though the broad suite also reaches
--- a/.github/workflows/performance-collect.yml
+++ b/.github/workflows/performance-collect.yml
@@ -229,9 +229,10 @@ jobs:
             image: ubuntu-24.04
     runs-on: ${{ matrix.os }}
     name: runtime (${{ matrix.platform }})
-    # The measured work is small — five samples of a 32 MiB text workload plus
-    # one release-quality build of it, on the order of a minute. The job's
-    # actual cost is dominated by building `rue` and `rue-bench` under
+    # The measured work is small — five samples of a 32 MiB text workload, five
+    # site builds of the live corpus, three of it at ten copies, plus the peer
+    # legs below and one release-quality build of each program. The job's actual
+    # cost is dominated by building `rue` and `rue-bench` under
     # //platforms:release. Wall clock is unaffected, since these run alongside
     # the `measure` legs, but each row is a runner's worth of compute per trunk
     # push and the budget below is sized for the build.
@@ -239,6 +240,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      # Fetches `rue`'s pinned peers as well as its own tools: `zola` and `hugo`
+      # at the repository root are dotslash shims, and the runtime leg runs both
+      # against the same corpus gazette builds (ADR-0072 Decision 5).
       - name: Install dotslash
         uses: facebook/install-dotslash@v2
 
@@ -247,13 +251,32 @@ jobs:
           ./buck2 build //crates/rue:rue //crates/rue-bench:rue-bench \
             --target-platforms //platforms:release
 
+      # The peer-event state of ADR-0072 Decision 9: the fixture identity, peer
+      # versions, and epoch the last FULL peer leg ran under. The per-run canary
+      # rides every observation regardless; this decides whether the rest of the
+      # peer matrix runs, and it is read at fixture-preparation time inside this
+      # same job.
+      #
+      # A cache miss costs one redundant peer leg and nothing else, which is the
+      # right direction to fail in: the alternative to a full leg is joining
+      # this run's gazette numbers to a peer sample the store may not have.
+      # Restore keys are prefix-matched newest-first, so the state survives the
+      # per-commit key changing on every push.
+      - name: Restore the peer-leg state
+        uses: actions/cache/restore@v4
+        with:
+          path: runtime-peer-state
+          key: runtime-peer-state-${{ matrix.platform }}-${{ github.sha }}
+          restore-keys: |
+            runtime-peer-state-${{ matrix.platform }}-
+
       - name: Measure compiled-program runtime
         env:
           RUE_BENCH_RUNNER_LABEL: github-hosted
           RUE_BENCH_RUNNER_IMAGE: ${{ matrix.image }}
         run: |
           set -euo pipefail
-          mkdir -p runtime-collected
+          mkdir -p runtime-collected runtime-peer-state
           RUE="$(scripts/rue-bin --target-platforms //platforms:release)"
           BENCH="$(./buck2 build //crates/rue-bench:rue-bench \
             --target-platforms //platforms:release --show-simple-output 2>/dev/null | tail -1)"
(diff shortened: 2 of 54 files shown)


steveklabnik (commented):
Review of `875fcd5`: the corpus-equivalence work is unusually thorough, but I found four correctness gaps in the peer evidence path that I think need fixing before these ratios enter the durable store.

1. **Peer samples do not receive the validation applied to Rue samples.** `check_peers` requires a non-empty sample vector and checks its artifact digest against `output_sha256`, but it never checks `process_elapsed_ns`, `exit_code`, the expected sample count, `emitted_files`, or (for non-canary rows) that `scale` equals the workload's declared scale. As written, a one-sample, zero-duration or nonzero-exit canary with a matching artifact digest is appendable, and `latest_comparison` will publish its median. That breaks the producer/validator separation this schema is explicitly designed around. Please apply the runtime sample sanity rules to every peer sample, require the epoch's declared sample count, and validate the configuration key (tool, scale, thread policy, role) rather than only individual fields.

2. **`latest_comparison` bypasses completeness and the canary publication gate.** It filters reports only on `is_appendable()`, selects the newest report containing any peer, and summarizes the raw samples directly. A report missing its required canary is deliberately appendable-but-partial, yet if it contains any successful full rows this function still publishes Rue/peer ratios from it. If it contains no peers, the search silently falls back to an older comparison instead of showing that the newest run has no valid denominator. The same path can publish a truncated Rue observation that `derive_runtime_epoch` correctly suppresses. The comparison should be based on the newest report and include only observations for which `outcome.publishes_workload(...)` is true; an incomplete newest observation should produce the honest empty state.

3. **The event-driven join is keyed only by fixture identity, so a failed peer-version transition resurrects stale rows.** The event detector correctly fires when either peer version moves, but canary-only observations store only the canary's version; they do not carry both current peer versions as the PR description claims. If Hugo is bumped and that full leg fails, `full` finds the previous successful leg and the fixture-only condition joins the old Hugo row into the new table. This seems to expose one identity doing two jobs: `fixture_digest` includes peer ports so it can drive peer events, but that also segments the raw Gazette series when only a peer template changed. I would separate the workload/corpus identity used for Rue's longitudinal series from a comparison-configuration identity containing fixture identity, all peer versions/ports, and epoch, record the latter on every observation, and use it for the full-leg join.

4. **The default-parallel rows are never checked for work equivalence.** `measure_peers` retains only the primary-policy tree for each tool, so the semantic oracle never sees the secondary tree. Validation proves only that a secondary row is deterministic with itself. A deterministic parallel build that emits a different tree therefore publishes as the user-experienced comparison. Since this is the same tool and corpus, requiring its `output_sha256` to equal that tool's validated primary-policy digest would close the hole without another semantic-oracle pass.

One CI gap is worth fixing with these: the required cross-tool equivalence step runs only the default 1x corpus, even though this change documents a real 10x-only failure where most Hugo pages skipped the expensive layout while the 1x check stayed green. Please exercise the 10x translation in required CI as well, either with the full 10x check or a focused scale fixture that covers duplicated layout/feed selection.

I ran `scripts/rue unit perf-schema` (203 passed) and `scripts/rue unit bench` (121 passed, 2 ignored); the PR's GitHub checks are also green. I did not find issues in the pinned shims, parity configurations, or current 1x semantic comparison beyond the evidence/publication gaps above.

PR #2401, opened 2026-08-14
RUE-1222: make the scheduled cache-free run carry the corpus safeguards

DorianListens:
Three follow-ups from RUE-1118, which converted the heavy corpora into cacheable build actions.

## The workflow being extended had never once succeeded

`correctness-repetitions.yml`'s only two scheduled runs both died in 18 seconds:

```
failure  Correctness repetitions  31371851477  18s  2026-08-10
failure  Correctness repetitions  30807197495  18s  2026-08-03
error: unexpected argument '--env' found
```

`buck2 test --env` is rejected at argument parsing. But moving it after `--` would only make the command *valid* — the value would land on the stamp-check test process, where nothing reads it. RUE-1118 moved the harness into a build action, so the repetition index has to travel as `-c rue.corpus_repetition=N`, which `cached_corpus_suite` folds into the action's env.

Proven end-to-end against real buck2: `scripts/ci-repeat-correctness //:cli-tests-shard-0 2` → both repetitions pass, with `cli-tests: measured 467 cases` written to `case-timings-1.jsonl` *and* `case-timings-2.jsonl`.

## `--no-remote-cache` never forced execution

It disables the *remote* cache only; buck2's local DICE state and materialized outputs survive, so repetitions 2..N were local hits. The same nonce fixes this by making each repetition a distinct digest. Measured: repetition 1 → `Commands: 31 (local: 31)`, repetition 2 → `Commands: 1 (local: 1)` — the corpus re-executed, the compiler did not. The two timing files differ by content and size, so they are genuinely independent samples rather than one file copied five times.

The claim that cache-freeness "holds by both construction and instruction" is gone from the workflow header, `build-cache.md`, and `ci.md`. Only construction holds — the fresh runner's empty `buck-out` and the absent BuildBuddy secret.

## Sub-tasks

**1. Shard-weight timings** now come from the repetition runs, which are the only ones that measure the corpus against the current tree. Five samples per case instead of one; `generate-cli-shard-weights.py` medians them. The docs now say to collect **all four** shard artifacts and that the tool *replaces* rather than merges a platform's weights — following the old instruction literally would have rewritten `platforms.linux-x64` with one shard's cases.

**2. The linux-x64 `local: 1`** is not fixed, because it could not be identified from source and guessing is worse than instrumenting. The cross-platform comparison behind it was never like-for-like: every `platform-corpus` matrix entry is `ubuntu-latest`, and arm64/macOS reach the corpora through a path that builds no `cached_corpus_suite` at all. `ci-heavy-suite` now prints the identity of every locally executed action, pinned to its own invocation via `--trace-id` — a bare `what-ran` reads the *last* invocation from the isolation directory, which is how a run that exited 3 having executed nothing still printed "every build action was served from cache". Tail-capped, because the `rue_corpus` row is always last and a head cap discarded exactly the row worth reading.

**3. The undeclared-input safeguard** covered 4 of 14 corpora. A new `execute-every-corpus` job sweeps **9** more cache-free, one runner each, with the inventory queried from `kind('_corpus_action', //...)` and cross-checked against the `rue_heavy_suite` label set so a converted-but-unswept corpus fails the inventory rather than passing silently.

Stated precisely: **13 of the 14 corpora execute somewhere in this workflow** — 4 as the repeated shards, 9 in the sweep matrix. `//:cli-tests` executes nowhere, resting entirely on the superset argument (identical harness, args, and absolutize list; the shards' union is the same case inventory), which `//:cli-shard-coverage-validation` machine-checks.

The sweep checks input *declaration*, a property of the rule, so an undeclared input is missing under either configuration. It does not exercise release-configured cache entries, and says so rather than implying "every corpus" means every configuration.

## Test fidelity

The fake `buck2` accepted anything, which is why a permanently broken invocation stayed pinned by a passing assertion. It now parses arguments as the real binary does — exiting 3 on `--env` before `--`, and serving `what-ran` rows only for the recorded trace id. Regression guards added for exactly the defects that hid here. **147 wrapper checks, up from 133.**

## Review round (Fable, at Steve's request)

Both findings fixed in `RUE-1222: make the uuid fallback and the sweep exclusions fail closed`. Each new check fails against the previous code.

**The no-uuid fallback was fatal, not degrading.** buck2 parses `BUCK_WRAPPER_UUID` before it parses an argument and rejects a set-but-empty value (exit 2, `invalid length: found 0`), so a host with neither `uuidgen` nor `/proc/sys/kernel/random/uuid` lost the whole corpus run rather than just the `what-ran` report the guard below it was written to lose. The variable is now exported only when non-empty. The fake accepted the empty value the real binary refuses — the same real/fake gap class this PR set out to close — and now rejects it at startup for every subcommand, distinguishing set-but-empty from unset.

**Sweep exclusions are now exact by construction.** `//:cli-tests-shard-*` was a prefix wildcard applied *after* the completeness cross-check, so a corpus merely named like a shard would pass every check and then silently drop out of the sweep — the one remaining path by which a corpus could leave the inventory without an error. `--exclude` now takes a target label and refuses a pattern; `--exclude-label` resolves through the same graph the inventory came from. The workflow excludes the shards by `rue_cli_shard`, the label that makes the superset argument true and that `test.sh` already subtracts for the same reason, so it follows `CLI_TEST_SHARD_COUNT` without an edit. Verified against the real graph: the label resolves to exactly the four shards and the matrix is the same 9 targets. The inventory fake now carries a shard-shaped decoy, making the drop a live regression rather than a described one.

Rebased on trunk (`adb99c08`); the only conflict was `docs/process/ci.md`, where trunk's ADR-0069 "Nothing executes twice" section and this branch's sweep prose were both inserted after the same paragraph. Kept both, branch text first.

Trunk's new RUE-1507 gate waives this workflow by name as `known_broken="RUE-1222"`. Left in place deliberately: it cannot be removed until a scheduled run actually succeeds, and the gate already self-clears by warning that the waiver has outlived its breakage once one does.

Closes RUE-1222.

### DIFF
--- a/scripts/ci-corpus-inventory
+++ b/scripts/ci-corpus-inventory
@@ -1,0 +1,135 @@
+#!/usr/bin/env bash
+# Enumerate every corpus suite whose run is a cacheable build action (RUE-1118).
+#
+# The inventory comes from the Buck graph, not from a list kept here: every
+# `cached_corpus_suite` expands to one `_corpus_action` rule, so querying that
+# rule kind names exactly the converted corpora and a corpus added later joins
+# the sweep without an edit. That property is the point — the caller (RUE-1222's
+# scheduled cache-free sweep) exists to guarantee that *every* corpus is really
+# executed periodically, and a hand-maintained list is precisely the thing that
+# would quietly stop being every.
+#
+# Fails closed. `scripts/affected-targets` fails open because over-running a
+# corpus is safe and under-running it is what selection is for; here the
+# opposite holds. An inventory that errored or came back empty would turn the
+# sweep green having proved nothing, which is the same false pass the sweep
+# exists to bound.
+set -uo pipefail
+
+json=false
+excludes=()
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --json)
+            json=true
+            shift
+            ;;
+        --exclude)
+            if [[ $# -lt 2 ]]; then
+                echo "error: --exclude requires a glob" >&2
+                exit 2
+            fi
+            excludes+=("$2")
+            shift 2
+            ;;
+        *)
+            echo "usage: scripts/ci-corpus-inventory [--json] [--exclude GLOB]..." >&2
+            exit 2
+            ;;
+    esac
+done
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$repo_root" || exit 1
+
+if ! actions="$(./buck2 uquery "kind('_corpus_action', //...)")"; then
+    echo "error: could not query the corpus inventory from the Buck graph" >&2
+    exit 1
+fi
+
+# Parse and validate the whole inventory before anything is judged against it: a
+# malformed label makes both the cross-check below and the sweep meaningless.
+all_suites=""
+while IFS= read -r action; do
+    [[ -n "$action" ]] || continue
+    # `root//:cli-tests-action` names the action of the `//:cli-tests` suite.
+    suite="${action#root}"
+    if [[ "$suite" != *-action ]]; then
+        echo "error: unexpected corpus action label: $action" >&2
+        exit 1
+    fi
+    suite="${suite%-action}"
+    if [[ ! "$suite" =~ ^//[A-Za-z0-9._/-]*:[A-Za-z0-9._-]+$ ]]; then
+        echo "error: corpus target is not a plain Buck label: $suite" >&2
+        exit 1
+    fi
+    all_suites="$all_suites$suite"$'\n'
+done <<<"$actions"
+
+# Completeness, enforced rather than assumed. The query above says "every target
+# of one rule kind, in the root cell", which is the set of corpora that happen to
+# be converted — not the set that ought to be swept. `rue_heavy_suite` is the
+# label CI and test.sh already use to mean "expensive corpus", so require every
+# heavy suite to be one of these actions. A corpus that regresses to a plain
+# sh_test, or a new heavy suite that never gets converted, then fails this
+# instead of quietly dropping out of the sweep while it still reports success.
+# The converse is deliberately not required: //:release-smoke is a converted
+# corpus that carries no heavy label.
+if ! heavy="$(./buck2 uquery "attrfilter('labels', 'rue_heavy_suite', //...)")"; then
+    echo "error: could not query the heavy-suite set to cross-check the inventory" >&2
+    exit 1
+fi
+missing=""
+while IFS= read -r suite; do
+    [[ -n "$suite" ]] || continue
+    suite="${suite#root}"
+    # Newline-delimited on both sides, so this is an exact label match and not a
+    # prefix that another package's target could satisfy.
+    case $'\n'"$all_suites" in
+        *$'\n'"$suite"$'\n'*) ;;
+        *) missing="$missing $suite" ;;
+    esac
+done <<<"$heavy"
+if [[ -n "$missing" ]]; then
+    echo "error: heavy suite(s) with no corpus action, so the sweep would skip them:$missing" >&2
+    exit 1
+fi
+
+# macOS ships Bash 3.2, where an empty array is "unset" to `set -u`: guard every
+# expansion rather than relying on `${#name[@]}`.
+suites=()
+found=0
+while IFS= read -r suite; do
+    [[ -n "$suite" ]] || continue
+    skip=false
+    for glob in ${excludes[@]+"${excludes[@]}"}; do
+        # shellcheck disable=SC2053  # glob match is the point
+        if [[ "$suite" == $glob ]]; then
+            skip=true
+            break
+        fi
+    done
+    if [[ "$skip" == false ]]; then
+        suites+=("$suite")
+        found=$((found + 1))
+    fi
+done <<<"$all_suites"
+
+if [[ "$found" -eq 0 ]]; then
+    echo "error: the corpus inventory is empty; refusing to report a vacuous sweep" >&2
+    exit 1
+fi
+
+if [[ "$json" == true ]]; then
+    # The label pattern above admits nothing JSON would have to escape, so the
+    # array is assembled directly rather than through a Python dependency.
+    separator='['
+    for suite in "${suites[@]}"; do
+        printf '%s"%s"' "$separator" "$suite"
+        separator=','
+    done
+    printf ']\n'
+else
+    printf '%s\n' "${suites[@]}"
+fi
--- a/scripts/ci-corpus-inventory
+++ b/scripts/ci-corpus-inventory
@@ -18,7 +18,15 @@ set -uo pipefail
 
 json=false
 excludes=()
+exclude_labels=()
 
+# Exclusions are exact, never globs. A glob is the one way a corpus can leave
+# this inventory without an error: it is applied after the completeness
+# cross-check below, so a corpus whose name merely begins with an excluded
+# prefix — a `cli-tests-shard-weights-smoke` against `//:cli-tests-shard-*` —
+# would pass every check here and then silently drop out of the sweep. That is
+# the exact failure this script exists to make impossible, so the prefix
+# spelling is gone: name a target, or name a label the graph resolves.
 while [[ $# -gt 0 ]]; do
     case "$1" in
         --json)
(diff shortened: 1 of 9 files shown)


scripts/ci-heavy-suite
```diff
@@ -32,28 +32,112 @@ target="$1"
...
+# this id means no output.
+buck_trace_id=""
+if command -v uuidgen >/dev/null 2>&1; then
+    buck_trace_id="$(uuidgen | tr '[:upper:]' '[:lower:]')"
+elif [[ -r /proc/sys/kernel/random/uuid ]]; then
+    buck_trace_id="$(cat /proc/sys/kernel/random/uuid)"
 fi
 
-./buck2 test "${test_args[@]}"
+BUCK_WRAPPER_UUID="$buck_trace_id" ./buck2 test "${test_args[@]}"
```
steveklabnik: The no-uuid fallback path is fatal instead of degrading: when neither `uuidgen` nor `/proc/sys/kernel/random/uuid` is available, `buck_trace_id` stays empty and this line still exports `BUCK_WRAPPER_UUID=""` — and the real buck2 refuses an empty value before parsing a single argument (exit 2, `Parsing buck2 invocation id from env variable BUCK_WRAPPER_UUID / Caused by: invalid length: found 0`). The guard at the `what-ran` report below (`[[ -n "$buck_trace_id" ]]`) shows the intent was to lose only the report, but the branch it degrades through kills the whole corpus run. Current CI lanes all have a uuid source, so this doesn't bite today — but `uuidgen` is genuinely absent on some Linux hosts, so the written fallback chain isn't hypothetical.

One-line fix: only set the variable when non-empty, e.g.

```bash
env ${buck_trace_id:+BUCK_WRAPPER_UUID="$buck_trace_id"} ./buck2 test "${test_args[@]}"
```

(or an `if`/`else` around the invocation, which is friendlier to a reader).

Corollary worth closing in the same change: the fake buck2 in `scripts/test-wrapper-scripts.sh` *accepts* an empty `BUCK_WRAPPER_UUID` and happily records it, so the suite this PR upgrades to "parses args like the real binary" cannot see this divergence — it's exactly the real/fake gap class the PR set out to close.

---

scripts/ci-corpus-inventory
```diff
@@ -0,0 +1,135 @@
...
+        *) missing="$missing $suite" ;;
+    esac
+done <<<"$heavy"
+if [[ -n "$missing" ]]; then
+    echo "error: heavy suite(s) with no corpus action, so the sweep would skip them:$missing" >&2
+    exit 1
+fi
+
+# macOS ships Bash 3.2, where an empty array is "unset" to `set -u`: guard every
+# expansion rather than relying on `${#name[@]}`.
```
steveklabnik, line 141: Nit, not blocking: the `//:cli-tests-shard-*` exclusion is a prefix wildcard applied *after* the heavy-suite completeness cross-check, so a future corpus whose name merely begins `cli-tests-shard-` (say, a `cli-tests-shard-weights-smoke`) would pass the cross-check and then silently drop out of the sweep. Only the four real shards match today, but this is the one place a corpus can leave the inventory without an error. Matching the four shard names exactly (or asserting the exclusion consumed exactly four labels) would keep the sweep's "fails closed" property airtight.

---

steveklabnik (commented):
Review by Fable (Claude agent), at Steve's request.

I verified the load-bearing claims against the real pinned buck2 rather than taking the description's word: the `--env`-after-target rejection reproduces byte-for-byte (exit 3), `-c rue.corpus_repetition=N` really reaches the action's env through `read_root_config` in `corpus.bzl` and is injected only when non-empty so ordinary digests stay byte-identical, `BUCK_WRAPPER_UUID` pins the invocation id and `what-ran --trace-id` finds exactly that invocation, and the real `what-ran` output format matches both the awk filter and the upgraded fake. The inventory runs clean against the real graph (14 corpus actions, matrix-shaped JSON), fails closed on query error/empty set/malformed label/unconverted heavy suite, and the `//:cli-tests` exclusion's superset argument is substantiated in BUCK (shards share `_CLI_TEST_ARGS`/env/absolutize, and `//:cli-shard-coverage-validation` machine-checks the union invariant). All 140 wrapper checks and the shell-baseline gate pass at the head. The weights-tooling doc claim ("replaces, not merges") matches the code. Solid, unusually well-evidenced change.

Two findings, one real and one nit, as inline comments:

1. The no-uuid fallback in `scripts/ci-heavy-suite` is dead-on-arrival — an empty `BUCK_WRAPPER_UUID` kills the run at argument parsing instead of degrading to "no what-ran report", and the upgraded fake accepts the empty value the real binary refuses, so the wrapper suite can't see it.
2. The shard exclude glob is a prefix wildcard applied after the completeness cross-check — the one remaining path by which a corpus could silently leave the sweep.

One description imprecision, not worth a change: the sweep job's matrix is 9 targets, not 14 — 13 of 14 corpora execute across the workflow as a whole, and `//:cli-tests` itself never executes anywhere, resting entirely on the (verified) superset argument. Worth stating that way if the description gets edited.

---

scripts/ci-heavy-suite
DorianListens: Fixed, and you were right that the fake was the reason this could sit there.

Went with the `if`/`else` rather than the `env ${var:+...}` form, on your parenthetical — the branch is doing something worth reading, so it gets to say so:

```bash
if [[ -n "$buck_trace_id" ]]; then
    BUCK_WRAPPER_UUID="$buck_trace_id" ./buck2 test "${test_args[@]}"
else
    ./buck2 test "${test_args[@]}"
fi
```

The fake now rejects a set-but-empty `BUCK_WRAPPER_UUID` at startup for every subcommand, before argument parsing, matching the real binary's exit 2 and `invalid length: found 0`. Note it distinguishes set-but-empty from unset (`${BUCK_WRAPPER_UUID+set}`) — unset is the only spelling that means "pick one for me", which is exactly the distinction the wrapper now relies on.

Three checks, all of which fail against the previous code:

- `fake buck2: refuses an empty BUCK_WRAPPER_UUID as the real binary does`
- `ci-heavy-suite: no uuid source costs the what-ran report, not the run`
- `ci-heavy-suite: and it claims nothing about caching it cannot know`

One wrinkle worth flagging: the natural way to test this is to make both fallback arms fail, and that is not reachable on Linux, where `/proc/sys/kernel/random/uuid` would succeed. So the test installs a `uuidgen` shim that yields nothing rather than removing it from `PATH`. What the code under test sees is the empty trace id either way, and it reproduces identically on both platforms; the comment says so rather than implying absence was simulated.

scripts/ci-corpus-inventory
DorianListens: Taken, and not as a nit — it was the last hole in the property the script advertises, so it seemed worth closing properly rather than papering over the one pattern in use today.

Rather than matching the four shard names exactly (which duplicates `CLI_TEST_SHARD_COUNT` in a third place) or counting the consumed labels (which pins the count without pinning *which*), exclusions are now exact by construction and there is no pattern form at all:

- `--exclude` takes a target label and **refuses** anything that is not one, so `//:cli-tests-shard-*` is now an error rather than something honored.
- `--exclude-label` resolves through the same graph the inventory came from.

The workflow excludes the shards as `--exclude-label rue_cli_shard`. That label is what makes your superset argument true in the first place, it is the same one `test.sh` already subtracts for the same reason, and it follows `CLI_TEST_SHARD_COUNT` without an edit here. Verified against the real graph: the label resolves to exactly the four shards, and the sweep matrix is the same 9 targets as before — behavior preserved, precision gained.

Your hypothetical is now a live fixture rather than a described risk: the inventory fake carries a `//:cli-tests-shard-weights-smoke-action` decoy, and `ci-corpus-inventory: a corpus merely named like a shard stays in the sweep` asserts it survives. `a glob exclusion is rejected, not honored` fails against the previous code.

One deliberate asymmetry: a label resolving to *nothing* is not an error, since over-sweeping is safe and failing there would let removing a label break an unrelated workflow. A label set that cannot be **read** does fail closed, same as the heavy-suite cross-check.

DorianListens:
Thanks — both findings fixed in `55baeefd`, replies inline on each.

The uuid one was a real bug and the fake was the reason it could sit there, so the fix closes both halves. The exclusion one I took as more than a nit: it was the last hole in the property the script advertises, so rather than matching the four shard names, exclusions are now exact by construction and the pattern form is gone entirely — `--exclude` refuses anything that is not a target label, `--exclude-label rue_cli_shard` resolves the shards through the graph. Your hypothetical is now a decoy fixture in the fake, so it fails rather than being merely described. 147 wrapper checks, up from 140.

Also took the description imprecision, since I was editing it anyway: it now says 9 sweep targets, 13 of 14 corpora executing across the workflow, and `//:cli-tests` executing nowhere.

Rebased on `adb99c08`. One conflict, `docs/process/ci.md`, where trunk's "Nothing executes twice" section and this branch's sweep prose both landed after the same paragraph — kept both, branch text first.

One note on trunk's new RUE-1507 gate, which waives this workflow as `known_broken="RUE-1222"`: I left the waiver in place. It can't come out until a scheduled run actually succeeds, and the gate already self-clears — once the workflow goes green it warns that the waiver has outlived its breakage. So the first green run after this lands is the signal to delete it, not this PR.

PR #2403, opened 2026-08-14
RUE-1505: propose ADR-0069 Amendment 1 declining CI remote execution

DorianListens:
**This is a proposal and must not be self-merged.** It asks for a maintainer ruling on ADR-0069 Amendment 1. The amendment carries a bold "Status: proposal. Not accepted, not implemented." block.

## The question and the answer

Should BuildBuddy remote execution, used today only in the merge-group canary, become how ordinary CI builds run? `premerge (linux-x64)` is the CI long pole, and its `Build all targets` step is p50 ~300s on cold compiler PRs against ~19s in the merge queue.

**Recommendation: keep RE exactly where it is.**

RE does win on the compiler build — 1.4× paired median, ~1.3× after discounting the toolchain archives the local build downloads and the worker doesn't. But that advantage is **entirely width**, not per-action speed. Per-action attribution across 212 canary and 212 reproducibility runs:

| action | remote s/run | local s/run | local ÷ remote |
| --- | ---: | ---: | ---: |
| `rue-compiler` rustc rlib | 8.1 | 7.9 | 0.98 |
| `rue-compiler` rustc metadata | 4.9 | 3.9 | 0.79 |
| `winnow` rlib (41 queued locally) | 0.0 | 1.4 | 40.8 |

RE ties or loses on the critical single actions and wins only where 10–60 actions queue behind the local executor. Premerge has neither shape: over 169 cold-compiler runs the step is 837 commands, **93.5% cache-served**, 54 executed locally — and **78.9% of it is two indivisible actions**, `oracle-diff-test-action` (48.7%) and `oracle-diff-spec-test-action` (30.1%). Every third-party crate combined is 0.13%. The honest projection for RE on premerge is ~14s, ceiling ~30s.

## Method

There is no BuildBuddy credential on this host, so nothing was measured locally. Instead the evaluation mines a controlled pair that **required CI already runs on every merge**: the `remote execution (linux-x64)` canary and the `compiler reproducibility (linux-x64)` job both build `//crates/rue:rue` cache-disabled, same commit, same runner class, concurrently — one `--prefer-remote`, one `--local-only`, verified identical at the target-configuration hash with per-run action counts matching 396/405 on both sides.

Two caveats are stated rather than buried: the local figure is a step-gap that also charges tool setup and a tree `touch` sweep to the local side, so every ratio over-estimates RE; and the pair is not exec-configuration-identical, since the RE platform selects a different linker driver.

## Cost, on the free tier

BuildBuddy's free tier publishes "100 GB of cache transfer" and "Up to 80 cores", with an explicit no-hard-limits policy — overage is a sales conversation, not throttling or failure.

**Upload alone is already ~28 GiB/day ≈ 850 GiB/month, 8.5× the allowance, with zero inference.** Including downloads puts the upper bound near 3.7 TB/month. The cache blew past the free tier long ago; RE's own transfer (~7 MiB/build) is not the driver. If Rue wants a durable claim on this capacity, the conversation is about the cache, not RE.

## Two corrections worth reading

The first draft of this amendment claimed concurrent duplication does not happen in practice. **It does.** `premerge` and `test (linux-x64-oracle-diff)` both execute `oracle-diff-test-action` cold and concurrently at the identical configuration, 75s median overlap, in 57 of 60 cold PR runs. The draft missed it by counting actions rather than seconds — `Commands: 347 (cached: 342, local: 5)` hides a 148s harness among those five. That is the error accepted §5 forbids, and the amendment now says so.

The draft also missed the cheaper lever. Both oracle-diff suites are **already** `tier = "slow"` with dedicated required lanes, and premerge builds them anyway, because `build_scope()` filters on the `//crates/` path prefix while these live at `//crates/rue-oracle-diff:…`. That is worth ~240s against RE's ~30s ceiling, and is filed as RUE-1511.

## Scope answers

`compiler reproducibility` stays cache-free and locally executed — a uniform remote container makes the two builds *more* alike and proves *less*. ADR-0070's negative control is unchanged and the finding argues for keeping one canary rather than generalizing. RUE-1407 is untouched: that was a trust decision about verdicts, and `noop_test_toolchain` means nothing honours test-execution caching regardless.

Evidence lives in `docs/notes/rue-1505-remote-execution-evaluation.md`, following the split ADR-0069 already uses for its RUE-1250 evidence.

Refs RUE-1505.

### DIFF
--- a/docs/designs/0069-ci-work-scheduling.md
+++ b/docs/designs/0069-ci-work-scheduling.md
@@ -9,7 +9,7 @@ accepted: 2026-08-09
 implemented:
 spec-sections: []
 superseded-by:
-relates: ["RUE-1250", "RUE-1262", "RUE-1265", "RUE-1164", "RUE-1130", "RUE-1131", "RUE-1222", "RUE-1116", "RUE-1118", "RUE-1119", "RUE-1157", "RUE-1163"]
+relates: ["RUE-1250", "RUE-1262", "RUE-1265", "RUE-1164", "RUE-1130", "RUE-1131", "RUE-1222", "RUE-1116", "RUE-1118", "RUE-1119", "RUE-1157", "RUE-1163", "RUE-1505"]
 ---
 
 # ADR-0069: CI work scheduling for a compiler monorepo
@@ -19,6 +19,12 @@ relates: ["RUE-1250", "RUE-1262", "RUE-1265", "RUE-1164", "RUE-1130", "RUE-1131"
 Accepted. Phase 2 (determination on every lane, RUE-1130) is implemented in
 the same change that accepted this ADR. Phase 1 (RUE-1262) and Phase 3
 (the duplication gate, RUE-1265) have since landed; Phases 4-6 are unstarted.
+
+**Amendment 1 (RUE-1505) is a proposal and is not accepted.** It records the
+remote-execution evaluation this ADR's open questions invite, and recommends
+declining RE as a scheduling lever. Everything above and below it remains the
+accepted text; nothing in the amendment is implemented. See
+[Amendment 1: remote execution is not the "more cores" lever (RUE-1505)](#amendment-1-2026-08-14-remote-execution-is-not-the-more-cores-lever-rue-1505).
 
 ## Summary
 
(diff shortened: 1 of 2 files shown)


steveklabnik:
This sounds good to me.

PR #2433, opened 2026-08-16
RUE-1534: add direct rue-air provider-path body-analysis test coverage

steveklabnik:
Closes RUE-1534

## What

Adds a test-only in-memory durable fact source (`FixtureFactSource`) and a fixture helper (`ProviderFixture::analyze`) in `rue-air` that drive `analyze_provider_ordinary_body` — the exact entry point the compiler's body transaction uses — through the production `ProviderBodyHost` and the same `OrdinaryBodyEngine`. This adds production-path test coverage; runtime compiler behavior is unchanged, and the expected compiler wall-time effect is neutral (every new item is `cfg(test)`-gated). The purpose is to make later deletion of the retired test architecture — the `Sema` `OrdinaryBodyAnalysisHost` impl, its whole-program test drivers, and their retired export plane — safe by giving the production semantic path direct coverage first.

## How the fixture mirrors production

- The analyzed body's RIR bundle carries exactly one declaration, like the compiler's per-body plan materialization; the helper asserts this.
- All context (callee signatures, nominal shapes, members, constants) crosses the provider boundary as explicit in-memory durable facts through the same five `Durable*Source` contracts `CompilerBodyDurableSource` implements.
- Missing facts fail closed into visible spanned diagnostics (e.g. `UndefinedFunction`), matching provider behavior.
- The `BodyFactProvider` value is a guard stub whose operations panic: tracing every consultation site shows the ordinary body path answers all facts through the durable source, so a panic means the path grew a new provider consultation the wiring must learn about.
- `ProviderWellKnownOptionFacts` is empty, the same value production passes on its probe path and for bodies with no fallible intrinsics.

## Tests

Direct provider-path tests cover: ordinary expression typing, ownership/move diagnostics (use-after-move; copy non-move), aggregate construction (nested nominal fields), a method call, comptime types and values (body-local alias; durable const), anonymous-nominal production with member initialization, fail-closed missing facts, warning preservation, and referenced-definition reporting. A structural guard proves the helper calls `analyze_provider_ordinary_body` and never re-enters the retired `Sema` drivers.

Four representative `Sema` tests migrate now (`test_analyze_addition`, `test_undefined_variable`, `test_use_after_move_error`, `test_copy_type_not_moved`), each replaced by an equal-or-stronger provider-path test; the larger migration is split into later PRs.

## Validation

- `scripts/rue unit air`: 749 passed, 0 failed (including the 13 new fixture tests)
- `scripts/rue quick`: 30 targets pass, 0 failures
- `//crates/rue-air:rue-air-clippy` and `scripts/rue fmt`: clean

### DIFF
--- a/crates/rue-air/src/sema/mod.rs
+++ b/crates/rue-air/src/sema/mod.rs
@@ -1523,4 +1523,6 @@ impl BodySema<'_> {
 #[cfg(test)]
 mod consistency_tests;
 #[cfg(test)]
+mod provider_fixture_tests;
+#[cfg(test)]
 mod tests;
--- a/crates/rue-air/src/sema/provider_fixture_tests.rs
+++ b/crates/rue-air/src/sema/provider_fixture_tests.rs
@@ -1,0 +1,881 @@
+//! Direct tests for the production provider body path.
+//!
+//! These tests drive [`analyze_provider_ordinary_body`] — the exact entry point
+//! the compiler's body transaction uses — through an in-memory durable fact
+//! source, so body analysis is exercised against the production
+//! `ProviderBodyHost`/`OrdinaryBodyEngine` seam rather than the retired
+//! whole-program `Sema` test drivers in `tests.rs`. A structural guard at the
+//! bottom of this file keeps the fixture off those drivers by name.
+//!
+//! The fixture mirrors the production topology exactly:
+//!
+//! - The analyzed body's RIR bundle contains only that body's declaration,
+//!   like the compiler's per-body plan materialization. Every other fact —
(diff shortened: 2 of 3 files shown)


steveklabnik:
The earlier dequeue was the `performance staleness` gate, not this PR: the queue run fetched `performance-data-v1` at 03:18:39, 24 seconds before the collector published the point for `b5a60ad` (03:19:03), so it saw the series 9 trunk commits behind after tonight's merge burst. The gate itself reports this as a repository-wide condition. The collector is healthy and the series is now 1 commit behind trunk, so I've re-queued; no changes to the PR were needed.

---

PR #2447, opened 2026-08-16
RUE-1544: verify dominance with the shared dominator tree

steveklabnik:
Fixes RUE-1544.

## Problem

`crates/rue-cfg/src/verify.rs` computed dominance privately, as a dense `blocks x blocks` `Vec<Vec<bool>>` iterated to a fixpoint, allocating a fresh blocks-sized `Vec` per block per sweep. The matrix alone is ~100MB at 10k blocks.

That ran on **every** `verify()`: at least twice per function (`Cfg::finish` and `Cfg::finish_after_optimization`), plus once per LICM preheader (`opt/loops.rs`) and once per accessor inline splice (`inline.rs`).

The crate already owned a near-linear Cooper–Harvey–Kennedy dominator tree in `crates/rue-cfg/src/dominators.rs` (RUE-914), used by `opt/forward.rs`, `opt/loops.rs`, and `opt/licm.rs`. The verifier's matrix was a second, slower computation of the same relation.

## What changed

* The verifier queries the shared `DominatorTree` instead of building a matrix. Its private reachability DFS and its private successor decoder go with it: the tree's own DFS answers reachability via a new `DominatorTree::is_reachable`, so the verifier makes one traversal where it used to make two plus a fixpoint.
* `DominatorTree::dominates` walked `b`'s idom chain, which is O(depth) — still quadratic across every operand use on a deep CFG. It now numbers the finished tree in preorder, records each subtree's last preorder number, and answers with an interval containment test in constant time. That is a strictly cheaper query for the existing optimizer consumers too.
* The tree is built after `verify_targets_and_slices()`, since it decodes every terminator and may only run once targets are known in bounds and payload slices valid.

Nothing outside `crates/rue-cfg/` is touched.

## Semantics

Unchanged — same accepts, same rejects, same messages, including the unreachable-block conventions the matrix encoded:

* an unreachable **use** is exempt from the dominance check (the matrix guarded the query on the use block's reachability);
* an unreachable **definition** dominates nothing, so a reachable use of one is still rejected with `... does not dominate the use`.

Two new verifier tests pin those two halves against the one graph shape that separates them. The hard-panic contract documented at the top of `verify.rs` (RUE-45) is untouched.

`DominatorTree` also gains an oracle test: over a reducible graph, an irreducible graph, and a switch fan-out, every block pair is checked against the definition of dominance — `b` must stop being reachable once `a` is deleted — which shares no code with the CHK fixpoint or the intervals. Mutating the interval bound by one fails six of the module's seven tests, so the oracle is not vacuous.

## Measurements

Generated single-function probe whose only axis is match-arm count (each arm lowers to one CFG block under a single `Switch`). Release compiler (`//platforms:release`), `-O3`, median of three samples, `phase_accounting.phase_ns.cfg_and_optimization`:

| match arms | before | after | speedup |
| --- | --- | --- | --- |
| 1000 | 14.1 ms | 2.7 ms | 5x |
| 2000 | 50.9 ms | 2.6 ms | 20x |
| 4000 | 199.7 ms | 4.5 ms | 44x |
| 8000 | 850.6 ms | 8.0 ms | 106x |
| 16000 | 3285.4 ms | 13.9 ms | 236x |

Before is 4x per doubling — clean O(blocks²). After is linear. Total compile wall time at 8000 arms falls from 1339 ms to 263 ms, and at 16000 arms from 4641 ms to 515 ms.

The probe generator is scratch tooling and is deliberately not committed; nothing under `performance/` changes, so no suite revision is involved.

## Codegen is unaffected

`examples/lattice/main.rue` at `-O3` compiles to a byte-identical executable before and after:

```
b893e76cfabed737b149d0e8c4d8527077dedd17da78418db20a28a7d30885e5  lattice-before
b893e76cfabed737b149d0e8c4d8527077dedd17da78418db20a28a7d30885e5  lattice-after
```

The 8k- and 16k-arm probe executables are byte-identical too, and both still print the expected `92`.

## Testing

* `scripts/rue unit cfg` — 258 passed, 0 failed.
* `./buck2 test //crates/rue-cfg/...` — unit tests, fmt check, clippy, and the debug-assert policy gate all pass.
* `scripts/rue quick` — 30 targets pass, 0 failures.
* `scripts/rue fmt`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QhUhrqAwp3CgP1AR7PDw1H)_

### DIFF
--- a/crates/rue-cfg/src/dominators.rs
+++ b/crates/rue-cfg/src/dominators.rs
@@ -10,10 +10,12 @@
 //! This is shared analysis infrastructure, not a pass: consumers recompute it
 //! on demand rather than caching it across mutations. The value-forwarding pass
 //! (RUE-914, `opt/forward.rs`) uses [`DominatorTree::dominates`] to turn sema's
-//! definite-initialization argument into an always-on checked invariant, and
+//! definite-initialization argument into an always-on checked invariant,
 //! natural-loop analysis (RUE-926, `opt/loops.rs`) and LICM (RUE-927,
 //! `opt/licm.rs`, run at `-O3` from `opt/mod.rs`) build the loop forest and its
-//! hoisting decisions on it.
+//! hoisting decisions on it, and the structural verifier (RUE-227,
+//! `verify.rs`) uses it for both reachability and the defined-before-use
+//! dominance check on every operand.
 //!
 //! ## Definitions
 //!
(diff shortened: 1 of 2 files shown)


steveklabnik (commented):
I reviewed the change at commit `60b753e3144cf21551041179dd0fb4e804fde12c` and found no actionable correctness issues. The verifier builds the shared dominator tree only after target and payload-slice validation; the reachability and unreachable-use behavior remains consistent with the previous matrix implementation; and the interval-query tests cover reducible, irreducible, and switch fan-out CFGs. The current CI run is still in progress.

PR #2448, opened 2026-08-16
[Linker] Resolve archives with a worklist and index Mach-O externals

steveklabnik:
Three bounded fixes in `crates/rue-linker`. The pipeline links roughly 1,280 single-function objects, so anything the linker repeats per object per pass is quadratic in the size of the link.

## Archive resolution (`linker.rs`, `add_archive`)

The fixed-point loop rebuilt its undefined-symbol list from scratch each round — rescanning **every** linked object's symbol table and pushing one owned `String` per undefined reference, with no deduplication.

The already-linked objects cannot change while archive members are being selected (members are added to `self` only after the loop), so their symbol tables are now read exactly once into a worklist, and each extracted member extends that worklist with its own undefined symbols — the transitive step the rescan existed to perform. Names are borrowed from `self`/the archive rather than cloned, and a `HashSet` dedups the queue while a `VecDeque` keeps insertion order, so extraction stays first-eligible in archive member order and independent of hash iteration order (RUE-848's property).

Deduplication also closes a real extraction hole rather than only shrinking a list. First-eligible selection skips members it has already pulled, so a second reference to a symbol with two providers extracted the *second* provider, which then collided as a duplicate definition. Two of the new tests fail on the old implementation with `DuplicateSymbol("shared")`:

- `test_archive_symbol_referenced_by_many_objects_pulls_one_member` — eight objects referencing one symbol, archive with two providers
- `test_archive_reference_satisfied_by_earlier_extraction_pulls_nothing` — a reference already satisfied by the member pulled in for an earlier reference

`test_archive_transitive_chain_walks_backwards` guards the rewrite itself: a five-deep chain whose providers all sit earlier in the archive than their users, reachable only by following each newly pulled member's own references.

## Mach-O object emission (`emit.rs`, `build_macho`)

External-symbol collection deduplicated with `extern_symbols.contains(...)` (O(S²) per object) and relocation emission resolved each symbol with `.iter().position(...)` (O(S) per relocation). Both now go through the `HashMap` index the ELF path at `emit.rs:229` already carries for exactly this reason.

The map is keyed on the *source* name rather than the underscored one: `add_macho_underscore` prepends exactly one `_` to every name, so both keys partition the relocations identically, and the underscored string is now built once per distinct symbol instead of again for every relocation naming it — the duplicated string building the brief called out. `extern_symbols` itself was only ever read for its length, so only the name offsets are retained.

`test_macho_relocation_symbol_resolution_round_trip` pins the index map end to end: an object whose relocations mix a self-reference (symbol index 0), interleaved string constants, repeated references to an already-registered external, and a late reference back to the first external is emitted, parsed back, and every relocation must name the symbol it was emitted for.

## Mach-O layout pre-pass (`linker.rs`, `link_macho` / `macho.rs`)

The pre-build layout builder received `merged_text.clone()` and `merged_data.clone()` — a copy of the entire linked image — for a computation that, per its own comment, reads only the lengths. `compute_dynamic_layout` now delegates to `dynamic_layout_for(ImageSizes { .. })`, and `link_macho` measures the merged image instead of copying it. Load-command sizes still come from the builder's own segments and commands.

## Verification

- `//crates/rue-linker:rue-linker-test` — 156 pass (152 before, 4 new).
- `scripts/rue quick` — 30 targets pass.
- Clippy and fmt checks for the crate pass; `scripts/rue fmt` applied.
- **Byte-identical output**, `examples/lattice/main.rue` compiled with the compiler built before and after the change:
  - x86-64-linux (ELF): `45784ce7c7cde992d7ea820912ca1692c05dc7582a367210d017b3765a9a89e7` both.
  - aarch64-macos (Mach-O), cross-compiled on Linux: `a29ad940222f28d879c769bc4696cbefcee02f87aa435687446f85a078278bd3` both.
- Timing on the same example (`--time-passes`, medians of three, host is noisy at this scale): `Link_archive_resolve` ~21ms → ~17ms on the ELF path. The rest is within run-to-run variance — the win here is complexity, not a headline number for this workload size.

**macOS is CI-validated.** Everything above ran on Linux; the Mach-O evidence is emitted bytes from a cross-compiled image, not a binary that was signed and executed. Please watch the `macos-15` leg of `native-platforms` (which also carries the baseline-bash half of the shell gate).

No Linear issue tracks this work, so none is cited; the change touches only `crates/rue-linker/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4B2t9mNjqWStgqVFFeRSN)_

### DIFF
--- a/crates/rue-linker/src/emit.rs
+++ b/crates/rue-linker/src/emit.rs
@@ -673,24 +673,35 @@ impl ObjectBuilder {
         }
 
         // External symbol names (for relocations)
-        // All symbols need underscore prefix for macOS
-        let mut extern_symbols: Vec<String> = Vec::new();
+        // All symbols need underscore prefix for macOS.
+        //
+        // Keyed by the *source* name rather than the Mach-O one, the same way
+        // `build_elf` indexes its externals: `add_macho_underscore` prepends
+        // exactly one `_` to every name, so both keys partition the relocations
+        // identically, and keying on the source name means the underscored
+        // string is built once per distinct symbol here instead of again for
+        // every relocation that names it. The map replaces a linear `contains`
+        // per relocation while building the table and a linear `position` per
+        // relocation while emitting them, both quadratic in an object's symbol
+        // count.
         let mut extern_name_offsets: Vec<usize> = Vec::new();
+        let mut extern_symbol_indices: HashMap<&str, usize> = HashMap::new();
         for reloc in &self.relocations {
             // Skip string symbols - they're handled via local symbols
             // String symbols use .rodata.strN format (possibly with @PAGE/@PAGEOFF suffix)
             if reloc.symbol.starts_with(".rodata.str") {
                 continue;
             }
-            // Always add underscore prefix for macOS
-            let macho_sym = crate::util::add_macho_underscore(&reloc.symbol);
-            if !extern_symbols.contains(&macho_sym) {
+            if !extern_symbol_indices.contains_key(reloc.symbol.as_str()) {
+                extern_symbol_indices.insert(reloc.symbol.as_str(), extern_name_offsets.len());
                 extern_name_offsets.push(strtab.len());
+                // Always add underscore prefix for macOS
+                let macho_sym = crate::util::add_macho_underscore(&reloc.symbol);
                 strtab.extend_from_slice(macho_sym.as_bytes());
                 strtab.push(0);
-                extern_symbols.push(macho_sym);
             }
         }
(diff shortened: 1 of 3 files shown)


steveklabnik (commented):
I reviewed the archive-resolution worklist, Mach-O external-symbol indexing, and the size-only Mach-O layout pre-pass. I did not find any actionable correctness issues in the changed code. The new tests cover repeated references, already-satisfied archive symbols, backward transitive extraction, and relocation-index round trips. CI is still in progress, including the macOS validation called out in the PR description.

PR #2449, opened 2026-08-16
ADR-0073 proposal: validation-certificate durability across append-only revisions

steveklabnik:
## What

Design document only — no compiler change. Proposes ADR-0073: letting validation certificates survive append-only input revisions so fresh-build validation costs O(edges) instead of O(import-depth × edges).

## Why

Discovery publishes one revision per frontier round and certificates are keyed to one exact revision, so each round expires every certificate and re-walks the green cone. Deterministic counters make the shape exact: `validation.certificate_misses` is 64²/256²/1024² on the corresponding module-chain fixtures, and a 1,024-module chain spends ~12.5s of a ~14.1s fresh build in discovery. Lattice (import depth ≈ 12) pays ~600k validation probes against ~34k claims on every fresh build in the ADR-0071 boundary. Follow-up promised in PR #2446, which removed the smaller of the two discovery quadratics.

## Shape of the proposal

- An explicit, directional certificate rule (same lineage, certificate not newer than request, equal validation epochs, exact node incarnation), implemented behind a narrowly named `extends_for_certificate` predicate with stated transitivity/monotonicity obligations — deliberately not reusing the namespace-binding compatibility relation on name resemblance.
- A validation epoch decided by mechanically enforced publication classes: discovery extensions preserve it by consuming the existing extension/mutation-validation proof plus an O(appended-keys) negative-observation check; toolchain acquisition, edits, removals, and any unknown class bump by default.
- A capture-by-key audit bounding the proof surface (module-set enumeration lives in query keys, so appends mint new nodes rather than staling old ones).
- A falsifier matrix including both directions of the single certificate slot: newer certificate rejected at an older pin, older certificate accepted forward, shadowing appends must bump, and a premerge ratio gate on the committed chain fixtures so `certificate_misses` regressing toward n² fails structurally.

The invariant structure, the directional gate, and the release-build enforcement requirement come out of an external design review of the initial sketch; the grounding in the existing extension-validation proof and the class table are how this repo's machinery meets those requirements.

## Validation

- `python3 scripts/validate-adrs.py` — registry valid, 74 records; README index regenerated with `--write`.
- Docs-only change; no compiler behavior affected.

---

### DIFF
--- a/docs/designs/0073-validation-certificate-durability.md
+++ b/docs/designs/0073-validation-certificate-durability.md
@@ -1,0 +1,269 @@
+---
+id: 0073
+title: "Validation-certificate durability across append-only revisions"
+status: proposal
+tags: [architecture, compiler, incremental, performance, query-engine]
+feature-flag: null
+created: 2026-08-16
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["ADR-0051", "ADR-0063", "ADR-0067", "ADR-0071", "RUE-1112", "RUE-1473"]
+---
+
+# ADR-0073: Validation-certificate durability across append-only revisions
+
+## Status
+
+Proposal. This is an internal query-engine design with no language-semantics
+change. It changes only the cost of red/green validation, never what
+invalidates what; that boundary is a falsifier, not an aspiration.
+
+## Summary
+
+A query terminal's validation certificate is currently keyed to one exact
+revision. Import discovery publishes one immutable input revision per
+frontier round, so every round expires every certificate and re-walks the
+accumulated green cone. Fresh-build validation therefore costs
+O(import-depth × graph edges), and a depth-n import chain costs O(n²): the
+deterministic counters record `certificate_misses` of exactly 64², 256², and
+1024² on the corresponding chain fixtures, and the 1,024-module chain spends
+roughly 12.5 s of a 14 s fresh build in discovery. The maintained Lattice
+workload (import depth ≈ 12) pays the same multiplier as roughly 600k
+validation probes against ~34k query claims on every fresh build measured
+under ADR-0071.
+
+Discovery rounds only ever add inputs. This ADR introduces a validation
+epoch that survives publications proven observationally monotone, a
+certificate rule that is explicitly directional, and mechanically enforced
+publication classes that decide the epoch in release builds. Fresh-build
+validation drops from O(depth × edges) to O(edges); a chain probe's
+`certificate_misses` falls from n² to O(n). Ordinary edit, removal, and
+retained-session invalidation behavior is unchanged.
+
+## Context
+
+### The mechanism today
+
+Red/green validation walks a terminal's recorded dependency edges and, on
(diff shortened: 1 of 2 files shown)


steveklabnik (commented):
The core design is strong: the directional gate, exact-incarnation requirement, release-build enforcement, conservative default bump, and falsifier matrix address the important safety cases.

I found one remaining correctness gap around the assumption of linear publication history. The query runtime does not currently make publication history globally linear: `publish_revision` accepts independent full views, and `publish_revision_overlay` permits more than one newer child of the same retained parent; the latter checks parent availability, compatibility equality, and `child.id > parent.id`, but not unique descent. Consequently, numeric ordering plus epoch equality is safe only if the new epoch machinery also guarantees that every epoch-preserving publication belongs to one unique extension chain and that all generic/full/sibling publication paths receive distinct epochs (or conservatively bump).

Please scope the section 2 claim to “certificate-eligible same-epoch history is linear,” state how that property is enforced at the runtime boundary, and add a sibling-overlay regression showing that a certificate from one child cannot validate the other. Otherwise `C.id <= T.id` can mistake a sibling for an ancestor. I would treat this as blocking for accepting the ADR because it determines whether revision ordering is actually a proof of ancestry.

steveklabnik:
Addressed in 6d75ef9. Confirmed the gap against the runtime source: `publish_revision` accepts independent full views, and `publish_revision_overlay` checks parent availability, compatibility equality, and `child.id > parent.id` but not unique descent, so siblings are representable and id ordering alone proves nothing about ancestry.

Section 2 now claims only the narrower, enforced property — certificate-eligible same-epoch history is linear — via a per-`(compatibility namespace, epoch)` extension head: a publication preserves the epoch only when it is a constrained-class extension whose parent is the current head, and every other publication (independent full view, non-head overlay, second child of an extended parent) receives a fresh or bumped epoch at the runtime publication boundary, in release builds. Transitivity of `extends_for_certificate` now follows from the head-extension rule rather than a global linearity assumption, and invariant clause 3 does the safety work in the sibling case since siblings never share an epoch.

The falsifier matrix gains the requested regression as gate-targeted test 3: two overlay children of one retained parent, certificate minted under the first, demanded under the second, rejection proven.

---

steveklabnik (commented):
Re-reviewed at 6d75ef9. The previous blocker is resolved.

Section 2 now states the right invariant: runtime history may branch, but certificate-eligible history within one `(compatibility namespace, epoch)` is a unique extension chain enforced at the publication boundary. Requiring the parent to be the pair’s current head makes epoch equality plus directional revision ordering a valid ancestry reduction; assigning independent views, non-head overlays, and sibling children fresh/bumped epochs closes the sibling-reuse hole. The added two-child overlay falsifier directly exercises that boundary.

I found no further issues with the ADR. The directional old-pin tests, release-build class enforcement, negative-observation handling, conservative default bump, capture-by-key audit, scaling gate, and byte-identity evidence form a suitably strong acceptance matrix.

PR #2450, opened 2026-08-16
Remove three quadratic costs from the backend

steveklabnik:
Fixes RUE-1335.

A single large straight-line function compiled with backend-phase time growing
as the square of its statement count. Three things in `rue-codegen` were
responsible. Each commit is independently byte-preserving.

## 1. Register coalescing swept `live_at` per merge (`regalloc.rs`)

`coalesce()` swept **every** `live_at` bitset on each successful merge to
substitute dst with src, so the number of full sweeps tracked the number of move
candidates: O(merges × instructions). Both factors scale with function size.

## 2. x86-64 peephole identity removal (`x86_64/peephole.rs`)

Pass 3 deleted with `Vec::remove` inside a loop, shifting the whole tail per
removal, and the FLAGS gate re-scanned that tail per candidate — O(removals × n)
twice over.

It now marks removals against the pristine vector and compacts once with
`retain`, matching what AArch64's pass 3 already does, with FLAGS-liveness
precomputed in a single backward pass (`compute_flags_dead`, the same technique
pass 1 already uses; recomputed here because passes 1 and 2 rewrite the stream).

Marking against the pristine vector is exactly what the in-place loop saw: it
only ever removed instructions *before* the one it was examining, so the gate's
forward scan never observed a removed instruction, and the tail at each decision
point was the pristine tail.

## 3. The `live_at` bitsets themselves (`liveness.rs`)

Fixing (1) by deferring the sweep left the backend still scaling at ~3.3x per
doubling, and a callgrind diff put the residual in the dense bitsets rather than
in anything that read them. `LivenessInfo::live_at` was a
`Vec<FixedBitSet>` sized instructions × vregs, so building it was quadratic on
its own — and nothing read its contents: the `live_at(idx)` accessor had no
callers, `compute_pressure` was exercised only by its own tests, and the field's
only live use was `.len()` as an instruction count.

So the field is gone, along with the two builders that fed it
(`compute_live_at`, `compute_acyclic_live_at`) and the pressure surface that
consumed it (`PressureInfo`, `compute_pressure` — the dead code RUE-1335 asked
to either wire up or delete). Coalescing loses its `live_at` maintenance
entirely rather than deferring it, which makes commit 1 a deletion. The
instruction count is now `LivenessInfo::instruction_count()`, backed by the
per-instruction `clobbers_at` table. A pressure-aware allocator that wants these
sets back should reintroduce them alongside its own consumer.

## Numbers

Generated single-function probe, `-O3 -j1`, release compiler, backend phase.
Runs are warm and interleaved across the three binaries, best of five — the
first run of any size on this host is several times slower than steady state, so
a naive before-then-after ordering overstates the win.

| statements | before | after 1+2 | after 1+2+3 | total |
| --- | --- | --- | --- | --- |
| 1000 | 120.7 ms | 24.5 ms | 19.2 ms | 6.3x |
| 2000 | 470.3 ms | 83.4 ms | 54.5 ms | 8.6x |
| 4000 | 2891.9 ms | 324.2 ms | 200.2 ms | 14.4x |

## Codegen is unchanged

Compiling with the before and after binaries and comparing executable sha256 —
identical at every stage of the series:

| check | result |
| --- | --- |
| `examples/lattice/main.rue` at -O0 / -O1 / -O2 / -O3 | match |
| probes n=1000 / 2000 / 4000 at -O3 | match |
| lattice + n=2000 on `aarch64-linux` and `aarch64-macos` | match |

The AArch64 checks matter because commit 3 touches the shared `liveness.rs`.

## Tests

- `scripts/rue unit codegen` — 718 passed, covering both backends. Two new
  peephole cases pin the FLAGS gate per instruction position: a redundant
  `mov r, r` is still removed ahead of a flags reader, and a flags-writing
  `xor r, 0` is kept between two removable neighbours. Four tests were deleted
  with the surface they exercised (two pressure tests, one `live_at` union test,
  one empty-info assertion folded into `instruction_count()`).
- `scripts/rue quick` — 30/30.
- `//crates/rue-codegen:rue-codegen-clippy` — pass.

## What is still superlinear

This does not make the backend linear: the probe still scales ~3.5x per doubling
(15.0 / 62.3 / 193.2 / 681.3 ms at n=1000…8000). A callgrind profile of the
final binary puts the residual in `compute_dataflow`, whose `live_in` table has
the same instructions × vregs shape that `live_at` had. The difference is that
`live_in` is load-bearing — `build_live_ranges` reads it — so making it cheap
means changing the liveness representation itself (sparse sets, or deriving
ranges without materializing per-instruction live-in), not deleting it. Tracked
separately as RUE-1545.

### DIFF
--- a/crates/rue-codegen/src/aarch64/liveness.rs
+++ b/crates/rue-codegen/src/aarch64/liveness.rs
@@ -492,8 +492,8 @@ mod tests {
         let info = analyze(&mir);
 
         assert!(info.ranges.is_empty());
-        assert!(info.live_at.is_empty());
         assert!(info.clobbers_at.is_empty());
+        assert_eq!(info.instruction_count(), 0);
     }
 
     #[test]
--- a/crates/rue-codegen/src/liveness.rs
+++ b/crates/rue-codegen/src/liveness.rs
@@ -403,7 +403,6 @@ where
         return (
             LivenessInfo {
                 ranges: IndexMap::new(),
-                live_at: Vec::new(),
                 clobbers_at: Vec::new(),
                 non_returning_at: Vec::new(),
                 vreg_classes,
@@ -480,22 +479,13 @@ where
         }
     });
 
-    // Step 6: Compute live_at for each instruction (union of live_in and live_out)
-    let live_at = match dataflow {
-        DataflowSets::Acyclic { live_in } => {
-            compute_acyclic_live_at(live_in, &successors, vreg_count as usize)
-        }
-        DataflowSets::Cyclic { live_in, live_out } => compute_live_at(live_in, &live_out),
-    };
-
-    // Step 7: Collect clobbers and the never-returning call sites (RUE-1224)
+    // Step 6: Collect clobbers and the never-returning call sites (RUE-1224)
     let clobbers_at: Vec<Vec<R>> = instructions.iter().map(|i| get_clobbers(i)).collect();
     let non_returning_at: Vec<bool> = instructions.iter().map(&get_non_returning).collect();
 
     (
         LivenessInfo {
             ranges,
-            live_at,
             clobbers_at,
             non_returning_at,
             vreg_classes,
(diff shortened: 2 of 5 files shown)


steveklabnik (commented):
I reviewed the changes against trunk and found no actionable correctness issues.

The new x86-64 pass-3 implementation preserves the prior FLAGS-gating decisions while replacing repeated Vec::remove/tail scans with one backward liveness pass and a single compaction. The live_at/pressure surface also has no remaining in-repo consumers, while the load-bearing live_in data remains intact.

The reported unit, quick, and clippy checks cover the affected paths; the GitHub CI run for this commit is still in progress.

PR #2451, opened 2026-08-16
Decorate warning sort keys and index unused-function lookups

steveklabnik:
Two warning-path hot spots in `crates/rue-compiler/src/session.rs`, both presentation-only.

## The sort key was built inside the comparator

`rooted_cfg_with_cancellation` sorted warnings with a comparator that computed the whole key for *both* operands of *every* comparison: a linear `graph.modules` scan for the span's file id, plus `warning.to_string()`, plus `format!("{:?}", warning.diagnostic())` — two heap-allocating renders per side per comparison.

Now the file-id → module-path map is built once, each warning is decorated with its key once, the decorated vector is sorted, and the keys are stripped. The sort stays stable and the key tuple is unchanged, so the order feeding `dedup()` is the same.

## Unused-function collection scanned twice per candidate

`rooted_unused_function_warnings` did a linear `graph.modules` scan per candidate declaration, then a linear AST-items scan matching `function.span == locator.declaration_span`. Modules are now keyed by module id once, and each module's span → function index is built on first use, so a program whose candidates touch one module never indexes the rest. The per-candidate work is one map lookup and one span lookup.

Both indexes keep the ties the scans kept: the first module wins a duplicated file id or module id, and the first item wins a duplicated span.

## Validation

Compiled output and diagnostics are unchanged, checked against a compiler built from the base commit:

| fixture | warnings | result |
| --- | --- | --- |
| 4 modules, mixed variable/function warnings | 14 | stderr and binary byte-identical |
| 100 modules × 20 dead functions | 2000 | stderr and binary byte-identical |
| 400 modules × 20 dead functions | 8000 | stderr and binary byte-identical |
| 400 modules, body + function warnings | 3203 | stderr and binary byte-identical |

Suites: `scripts/rue ui` (250 passed), `scripts/rue unit compiler` (878 passed), `scripts/rue quick` (30 passed), the per-crate clippy and fmt gates, and `scripts/rue fmt`.

A new session test, `warnings_are_ordered_by_module_then_span_across_modules`, pins the cross-module order the UI cases only cover per module: a root importing `alpha.rue` and `zeta.rue`, whose own path sorts between them, so module-path order and the root-first order that discovery follows disagree. It asserts the module grouping, ascending spans within a module, which identifier each warning names, and that a second compile presents the identical sequence.

## Measurements

Callgrind instruction counts, same fixture and same build configuration:

| fixture | before | after |
| --- | --- | --- |
| whole compile, 400 modules / 8000 warnings | 21.13G | 20.73G |
| `rooted_unused_function_warnings` there | 339M | 150M |
| whole compile, 100 modules / 2000 warnings | 4.843G | 4.820G |

The saving grows faster than the input: 4× the modules and warnings gives roughly 17× the instructions saved, which is the removed modules × warnings term. Wall-clock difference on these fixtures is inside run-to-run noise on this host.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LqQoTDHMo4JTus7K5KpLiv)_

### DIFF
--- a/crates/rue-compiler/src/session.rs
+++ b/crates/rue-compiler/src/session.rs
@@ -1,6 +1,6 @@
 //! In-process canonical parse, merge, and RIR query orchestration.
 
-use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
 use std::sync::{Arc, Mutex};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


steveklabnik:
The `fmt` job is red on infrastructure, not on formatting: it died in the dotslash install step with `curl: (35) Recv failure: Connection reset by peer` before any check ran ([job log](https://github.com/rue-language/rue/actions/runs/31967002211/job/95213586064)). `clippy`, `actionlint`, `rust-project`, `affected targets`, `CI contract`, `asan` and the digest-sensitivity job are green on the same commit, and locally `scripts/rue fmt` leaves the tree clean with `//crates/rue-compiler:rue-compiler-fmt-check` passing.

A re-run clears it, but this account gets a 403 on `rerun-failed-jobs`, so it needs a maintainer re-run or the next push to the branch.

---

steveklabnik (commented):
I reviewed the indexing changes and did not find any blocking correctness issues. The decorated warning sort preserves the previous stable tuple and duplicate behavior, and the lazy per-module span index preserves the first-match behavior of the old scans. Verification: `scripts/rue unit compiler` passes on this commit (878 passed, 0 failed), including the new cross-module warning-order test; the PR diff is also whitespace-clean.

PR #2453, opened 2026-08-16
Carry validation certificates across strictly additive revisions

steveklabnik:
## What

Implements accepted ADR-0073. Two changes:

**`rue-query`** — every published revision gets a validation epoch. A full view starts a fresh epoch and becomes its compatibility namespace's extension head; `publish_revision_overlay` continues the parent's epoch only when the parent is that head **and** the delta strictly adds new leaf identities (checked mechanically against the parent's effective view, bounded by `OVERLAY_COMPACTION_DEPTH`). Sibling children and re-stamping overlays start fresh epochs, so certificate-eligible same-epoch history is one linear additive chain by construction, exactly as ADR-0073 §2 requires. The certificate gate (`extends_for_certificate`) then accepts a certificate at its exact revision, or forward along the same epoch when the certified cone never observed a missing leaf.

The cone bit is the load-bearing soundness piece: `optional_input` records stamp-0 negative observations, and an added leaf changes the meaning of exactly the cones that observed its absence. The bit is composed once at terminal publication (frame's own stamp-0 inputs ∪ observed dependency terminals' bits, conservative on abort-prefix replays) and travels on terminals and certificates. Present-observing cones are provably still green under any additive extension; impure cones (the per-round `resolve-import` probes and the trusted-module demand probe — the only two production `optional_input` users) keep today's exact-revision behavior. This makes the ADR's publication-class table fully mechanical: toolchain acquisition and discovery appends need no trusted class declarations at all.

**`rue-compiler`** — the aggregate `accepted-import-topology` leaf is keyed by frontier round instead of re-stamped under one `"current"` identity (nothing in the production pipeline observes it), so discovery overlays are strictly additive at the runtime boundary and preserve their epoch.

## Measured (release ThinLTO, `-O3 -j1`, x86-64 Linux container, same-output paired runs)

| probe | `certificate_misses` before → after | wall before → after |
| --- | --- | --- |
| 64-module chain | 4,096 (= 64²) → **1** | ~neutral |
| 256-module chain | 65,536 (= 256²) → **1** | 1,024 → 974 ms |
| 1,024-module chain | 1,048,576 (= 1024²) → **1** | **~14.1 s → ~10.4 s** |
| frozen Lattice | 588 → 10 | neutral within host noise |

Executable hashes are byte-identical everywhere: Lattice reproduces `b893e76c…85e5`, the 1,024-chain reproduces `802bf2e1…9f4d`.

The remaining superlinear chain cost has a different owner: the frontier loop re-dispatches `resolve-import` for every plan occurrence each round (O(rounds × plan) requests at ~8µs fixed request overhead — `validation.traversals` still grows ~n²). That is a dispatch-volume change, independent of certificates, and should be its own bounded issue; it is also why Lattice (import depth ≈ 12) sees only a small direct win here.

## Falsifier tests (ADR-0073 acceptance matrix)

New in the `rue-query` suite, alongside the 170 existing tests (174 total, all passing):

- `additive_overlay_carries_certificates_without_a_cone_walk` — zero certificate misses, zero re-demands across a strict head extension.
- `missing_leaf_observation_blocks_certificate_carry` — an impure cone is re-demanded, never carried; the satisfied probe recomputes and observes the added leaf.
- `sibling_overlay_children_never_share_certificates` — two children of one parent; the second never rides the first's certificates.
- `certificates_are_directional_across_the_epoch_chain` — newer certificate rejected at an older pin; after old-pin revalidation overwrites the slot, the older certificate is accepted forward with zero misses.

Deferred to follow-ups: the premerge ratio gate on the committed `scale_modules` fixtures (needs a CI-tier decision on where a compile-and-assert-counters test belongs), and the frontier dispatch-volume fix above.

## Validation

- `rue-query` test binary: 174/174 pass
- `scripts/rue quick`: 30/30 pass
- `scripts/rue fmt`: clean
- Paired same-output hash checks as above

---

### DIFF
--- a/crates/rue-compiler/src/revisioned_query_database.rs
+++ b/crates/rue-compiler/src/revisioned_query_database.rs
@@ -3082,8 +3082,18 @@ fn import_context_input() -> InputIdentity {
     InputIdentity::new("import-discovery-context", "current")
 }
 
-fn accepted_import_topology_input() -> InputIdentity {
-    InputIdentity::new("accepted-import-topology", "current")
+/// The aggregate accepted-import-topology leaf, keyed by the frontier round
+/// that produced it. Per-round keying makes a grown topology a NEW leaf
+/// identity instead of a re-stamp of one "current" identity, so a discovery
+/// overlay stays strictly additive at the runtime boundary and preserves its
+/// validation epoch (ADR-0073). Consumers that want the newest aggregate
+/// observe the exact round they were computed against; nothing in the
+/// production pipeline observes this leaf today.
+fn accepted_import_topology_input(frontier_round: u64) -> InputIdentity {
+    InputIdentity::new(
+        "accepted-import-topology",
+        format!("round-{frontier_round}"),
+    )
 }
 
 fn accepted_read_input(module: &ModuleId) -> InputIdentity {
@@ -3106,7 +3116,7 @@ fn import_observation_input(request: &ImportDiscoveryRequest) -> InputIdentity {
 
 #[cfg(test)]
 fn test_import_graph_input() -> InputIdentity {
-    accepted_import_topology_input()
+    accepted_import_topology_input(0)
 }
 
 fn import_input_error(message: impl Into<String>) -> CompileError {
@@ -19432,7 +19442,10 @@ impl RevisionedQueryDatabase {
             } = &mut *store;
             let accepted_topology_stamp =
                 exact_value_stamp(next_stamp, topology_stamps, &accepted_topology);
-            leaves.push((accepted_import_topology_input(), accepted_topology_stamp));
+            leaves.push((
+                accepted_import_topology_input(frontier_round),
+                accepted_topology_stamp,
+            ));
             retain_stamp_value(topology_stamps, &accepted_topology);
             leaves.push((
                 import_context_input(),
@@ -19689,7 +19702,7 @@ impl RevisionedQueryDatabase {
                 // the parent stamp plus this overlay's sorted fact delta, so
                 // lookup and retention stay O(delta) without a whole-ledger scan.
                 let stamp = exact_value_stamp(next_stamp, topology_stamps, &accepted_topology);
-                leaves.push((accepted_import_topology_input(), stamp));
+                leaves.push((accepted_import_topology_input(frontier_round), stamp));
                 retain_stamp_value(topology_stamps, &accepted_topology);
                 stamp
             } else {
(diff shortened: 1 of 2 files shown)


steveklabnik (commented):
I found one blocking soundness issue.

**[P1] Make cone purity participate in red/green identity — `crates/rue-query/src/lib.rs`, `QueryFamily::publish` around the stamp decision (current head lines 6885–6894).**

`cone_missing_observation` is stored on each new terminal, but the decision to reuse the previous stamp compares only terminal kind, outcome, and semantic diagnostics. The bit can therefore change while the node keeps the same stamp.

That creates an unsound sequence across epochs:

1. In revision A, child C has no negative observation, returns value `0`, and publishes stamp S with `cone_missing_observation = false`; parent P observes `(C, S)` and also records false.
2. A full publication starts a fresh epoch. Under revision B, an ordinary edit makes C execute `optional_input(x)`, observe x absent, and still return the same value/diagnostics `0`. C therefore keeps stamp S even though its cone bit is now true.
3. P validates at B: its retained dependency still matches `(C, S)`, so P remains green. `mark_terminal_validated` certifies P in B’s epoch using P’s retained terminal bit, which is still false; validation never folds the current child terminal’s bit back into P.
4. An additive revision C adds x. P’s certificate now passes the cross-revision gate and returns the stale result without walking the child whose absence was satisfied.

The new `missing_leaf_observation_blocks_certificate_carry` test starts with the child already impure, so it does not exercise this same-value false→true metadata transition.

Please either include `terminal.cone_missing_observation == cone_missing_observation` in the stamp-reuse equality (so a purity transition changes the child stamp and invalidates parents), or make recursive validation return and compose the current cone bit instead of only the stamp. Add a regression that performs the four-step sequence above and proves the final append recomputes to the newly present value.

I did not find another issue in the epoch-head or directional-gate mechanics; the blocker is specifically that the transitive safety metadata is not currently part of the red/green observation seen by parents.

steveklabnik:
Fixed in 827c4fe, taking the first of the two options: `cone_missing_observation` now participates in the stamp-reuse equality in `publish`, so an equal-value publication whose cone purity changed mints a new stamp, dirties parents through the ordinary `(node, stamp)` observation, and they recompute and inherit the current bit. Composition stays publication-time only; validation still returns just the stamp.

The regression (`purity_transition_with_an_equal_value_changes_the_stamp`) runs the exact four-step sequence: pure child at revision A; an equal-value edit at B that turns the cone impure, asserting the child's stamp changes; then the additive revision satisfying the probed absence, asserting the parent **recomputes to the newly present value** rather than carrying the stale result.

Full `rue-query` suite is 175/175, quick suite green, and the paired evidence re-verified on the rebuilt release compiler: Lattice reproduces `b893e76c…85e5` with certificate misses still 10, the 1,024-chain reproduces `802bf2e1…9f4d` with misses still 1 — the equality change costs nothing in these builds because no equal-value purity transition occurs in them; it exists precisely for the sequence the review constructed.

---

PR #2464, opened 2026-08-17
Propose ADR-0074 and ADR-0075: structural identity and wave-granular discovery

steveklabnik:
Records the two semantics-adjacent contract changes Steve approved for the speed campaign, with falsifiers spelled out before any implementation lands (same discipline as ADR-0073).

**ADR-0074 — Structural node identity.** `NodeIdentity` is today a pair of formatted display strings: ordering, hashing, the RUE-1381 published dependency order, and the retained-memory charge all consume the text, which is why the 2026-08-16 lazy-display prototype removed zero of the 32,912 eager materializations (that measured rejection is recorded in the audit note merged in #2462). The ADR gives every `QueryKey` a content-derived 128-bit stable hash; order/equality/hash become `(family, stable_hash)` with lazily-formatted text as the cold collision tiebreaker; the retained charge is re-denominated in structural units with a one-time budget recalibration; lazy display formatting then actually works. Falsifiers cover cross-run and cross-worker-count determinism, forced-collision totality, byte-identical diagnostics/executables, and RSS neutrality. One-time rebaseline: counters encoding the current text order shift once.

**ADR-0075 — Wave-granular discovery.** A depth-n import chain takes n discovery rounds *by contract* — one hop per revision publication. With per-round costs now delta-scoped (ADR-0073, #2457), the round count itself is the last O(depth) structural factor. The ADR makes a round resolve transitively (read/parse/resolve newly discovered sources within the same round), publishing one revision per wave: O(1) discovery revisions per chain, ~11 publications dropped from every fresh Lattice build. The closing whole-plan re-root is replaced by a cumulative fail-closed exhaustion witness with the re-root as the unprovable-coverage fallback. The invalidation coarsening (mid-wave mutation → wave-granular fail-closed re-run; batch stamp verification at publish) is enumerated with falsifiers, including a mid-wave-mutation test hook requirement and premerge-gated revision counts on the chain fixtures.

Sequencing: ADR-0074 implementation is `rue-query`-local and can start immediately; ADR-0075 implementation starts after the in-flight delta-scoping change lands (same files).

---

### DIFF
--- a/docs/designs/0074-structural-node-identity.md
+++ b/docs/designs/0074-structural-node-identity.md
@@ -1,7 +1,7 @@
 ---
 id: 0074
 title: "Structural node identity for dependency order and retained charge"
-status: proposed
+status: proposal
 tags: [architecture, compiler, performance, query-engine]
 feature-flag: null
 created: 2026-08-17
--- a/docs/designs/0074-structural-node-identity.md
+++ b/docs/designs/0074-structural-node-identity.md
@@ -1,0 +1,139 @@
+---
+id: 0074
+title: "Structural node identity for dependency order and retained charge"
+status: proposed
+tags: [architecture, compiler, performance, query-engine]
+feature-flag: null
+created: 2026-08-17
+accepted:
+implemented:
+spec-sections: []
+superseded-by:
+relates: ["ADR-0063", "ADR-0071", "ADR-0073", "RUE-1381"]
+---
+
+# ADR-0074: Structural node identity for dependency order and retained charge
+
+## Status
+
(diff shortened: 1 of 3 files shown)


steveklabnik:
Review follow-up: ADR-0074’s collision-safety contract was underspecified. `QueryKey::stable_identity()` may collide; the existing `rue-query` `CollidingKey` fixture has unequal keys with identical display text, so a forced stable-hash collision cannot safely use display text as the sole fallback. I opened draft follow-up PR #2465 to require a deterministic structural collision witness and include that case in the falsifier contract. It has not been queued for merge.

PR #2465, opened 2026-08-17
Clarify structural identity collision handling

steveklabnik:
Follow-up to #2464.

ADR-0074 originally specified formatted display text as the collision fallback for the proposed structural node identity. That is insufficient because `QueryKey::stable_identity()` is explicitly allowed to collide; the existing `rue-query` `CollidingKey` fixture has two unequal keys with identical display text. Under a forced stable-hash collision, a text-only fallback would make those identities compare equal.

This follow-up requires a deterministic structural collision witness (or exact typed-key equality in the collision bucket), makes that witness part of canonical ordering, and adds the same-display-text fixture to the falsifier contract.

Validation: `python3 scripts/validate-adrs.py` passes (`ADR registry valid: 76 records`).

This PR is intentionally draft and has not been queued for merge.

### DIFF
--- a/docs/designs/0074-structural-node-identity.md
+++ b/docs/designs/0074-structural-node-identity.md
@@ -52,14 +52,18 @@ falsified) lazy formatting on top:
    **stable key hash**, computed from the typed key's fields with a fixed,
    keyed-with-a-compile-time-constant hasher — never from the formatted
    string, and never from addresses, allocation order, or scheduling.
-2. `NodeIdentity` ordering, equality, and hashing become
-   `(family, stable_hash)` with formatted text as a lazily-computed, cold
-   tiebreaker that is only reachable on a 128-bit hash collision between
-   distinct keys of the same family.
+2. `NodeIdentity` ordering, equality, and hashing use `(family, stable_hash)`
+   on the fast path. If two unequal typed keys in one family share that hash,
+   the runtime must consult a deterministic structural collision witness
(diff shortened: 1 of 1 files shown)


steveklabnik:
Verified against the code and endorsed — the original text-tiebreaker contract was unsound exactly as described. `revisions_and_key_identities_fail_closed_on_conflicts` (`crates/rue-query/src/lib.rs:18362`) constructs two unequal `CollidingKey`s whose `stable_identity()` both render `"collision"`, so under a forced stable-hash collision a text-only fallback would conflate distinct keys. The structural-witness requirement is the right fix, and making it part of canonical ordering keeps the determinism falsifier intact.

One consequence worth recording for the implementation: that same fixture currently pins the *opposite* display-level behavior — `assert_eq!(first.node(), second.node())` at `:18421`, i.e. today two colliding-text keys deliberately share one display `NodeIdentity` while the typed memo map keeps them apart. Under the amended contract their stable hashes differ, so `NodeIdentity` equality becomes strictly finer and that assertion legitimately flips to `assert_ne!`. That's a correct, expected one-time behavior change (published observations of such pairs become two identities instead of one), and the implementation should update the fixture assertion deliberately with a comment, not discover it as a surprise failure. I've relayed the amended contract to the in-flight ADR-0074 implementation so it builds against this version.

Promoting this from draft and queueing it — the amendment should land before the implementation PR does.

---

PR #2495, opened 2026-08-18
RUE-1561: share recursive ownership-property results in the semantic provider

steveklabnik:
## Problem

`type_carries_linear`, `type_has_drop_glue`, and `type_is_copy` in `SemanticNucleusTypeProvider` each walk the durable type graph independently, each with a fresh `BTreeSet` cycle-breaking stack, so an aggregate mentioned more than once was re-walked and its signature re-resolved every time.

## Change

Each walker now goes through a memoizing entry point backed by a provider-local `BTreeMap<StableDefinitionKey, OwnershipProperties>`. Nominal definitions are the only keys stored, since reaching one is what costs a signature resolution; everything else goes straight to the walk.

The three properties stay separate `Option` fields rather than one computed bundle, because they are asked for independently and each costs its own traversal — filling one must not force the other two.

### What may be stored

This is the part worth reviewing. Each walker breaks a recursive type by answering **provisionally** for a key already on its own stack — `DoesNotCarry`, `false`, and `true` respectively. A result reached through such an answer is a property of *that stack*, not of the type, and storing it would let a later independent query read a wrong answer.

`OwnershipWalk` carries that condition back up as taint:

- a subtree measures its own taint (`mem::replace(&mut walk.tainted, false)`) rather than inheriting the caller's, so one recursive branch cannot suppress memoization of an unrelated sibling;
- the subtree's taint is then folded back into the caller's;
- a tainted result is returned but never stored.

`LinearOwnershipFact::Deferred` and every `Err` are likewise never stored: they say the signature was not resolvable yet, which a later request may answer differently. Signature-reentry and cycle-abort handling is untouched.

`resolved_signature` takes `&self` and records its dependency through the query runtime, so a memo hit that skips it cannot drop an edge the first walk already observed — the dependency sets it feeds are sets, so the multiplicity change is not observable.

## Measurements

Honest summary: **correct and byte-identical, but the win is small.** Reporting it so the complexity/benefit trade is visible.

Emitted executables are byte-identical before and after:

| workload | sha256 |
| -- | -- |
| `examples/lattice/main.rue` `-O3` | `b893e76c…` both |
| aggregate-heavy probe (120 fns over a 4-level nested aggregate) | identical |
| call-heavy probe (60 fns, 60 call sites) | identical |

Provider analysis time is unchanged within noise (5 runs each, sorted): aggregate-heavy 243.1–293.0 ms before vs 243.8–282.5 ms after; call-heavy 230.3–294.0 ms before vs 213.4–258.8 ms after. The spread is ±20%, so nothing here is a signal.

Instrumented memo counters explain why. These walkers are driven by ownership gates, not by every body:

| workload | walks | memo lookups | reuses |
| -- | -- | -- | -- |
| `examples/lattice/main.rue` | 60 | 28 | 3 |
| gate-heavy program (`@require_droppable` over a repeated aggregate graph) | 1 | 7 | 4 |
| aggregate-heavy probe | 0 | 0 | — |
| call-heavy probe | 0 | 0 | — |

The redundancy the issue describes is real — 4 of 7 lookups are reuses where the walkers are actually exercised — but Lattice avoids 3 traversals in a ~1.7 s compile. The taint discipline is the cost side of that trade, and it is subtle enough that a future edit has to preserve it deliberately. Happy to close this unmerged instead if the trade doesn't look worth it; the measurement is the point either way.

Worth noting: `tainted` never fired on any workload measured, because by-value recursion is rejected upstream as infinite-size (E0483) and pointers do not recurse in these walks. The guard is for correctness if that path ever becomes reachable, not for a case seen today.

## Tests

New `ownership_property_memo_preserves_decisions_across_repeats_and_recursion`, driving `@require_droppable` over four type graphs:

- repeated non-linear aggregates (`Leaf`/`Mid`/`Top`, each mentioned three times) stay droppable;
- linearity survives a reused answer — `Mid` carries a linear field and is mentioned twice, so a memo that stored the wrong answer for the second mention would let it pass;
- mutually recursive aggregates through `ptr const`, where only one of the pair owns the linear field, asserted in both directions so the two cannot share one answer.

`scripts/rue unit compiler` 893/893, `scripts/rue quick` 30/30, `scripts/rue ui` 250/250, `scripts/rue spec` 2346/2346.

Fixes RUE-1561

---
_Generated by [Claude Code](https://claude.ai/code/session_01GEt7kDT3SQnCVfGTnb8qP6)_


steveklabnik:
Blocked on stuck CI infrastructure, not on this change. Recording it since I can't clear it myself.

## What's wrong

Workflow run [32099376665](https://github.com/rue-language/rue/actions/runs/32099376665), on the original head `f118d024`, has been `in_progress` since **04:30:09Z** with `updated_at` frozen at 04:30:48 — no job has reported in over three hours. It holds this PR's `concurrency` group (`CI-refs/pull/2495/merge`), so every later run queues behind it and never starts:

| run | head | created | status |
| -- | -- | -- | -- |
| 32099376665 | `f118d024` | 04:30:09 | **in_progress (wedged)** |
| 32106436059 | `12780115` | 06:19:26 | cancelled |
| 32108591942 | `44cbfc5f` | 06:49:31 | cancelled |
| 32111872804 | `44cbfc5f` | 07:31:50 | pending, queued behind the wedged run |

`cancel-in-progress: true` cancelled the two later runs but has never managed to cancel the wedged one, which is what points at this being stuck on GitHub's side rather than a workflow-config problem.

## What I tried

- Merged current trunk and re-pushed (`44cbfc5f`) — new run, same queue.
- `POST /actions/runs/32099376665/cancel` — `403 Resource not accessible by integration`; this token has no `actions: write`.
- Closed and reopened the PR, which normally cancels a PR's in-flight runs. It cancelled run 32108591942 and left the wedged one running.

Auto-merge is re-enabled (the close cleared it), so this merges on its own once a run actually completes.

## What clears it

Either cancel run 32099376665 from the Actions UI, or wait — GitHub's 6-hour job limit should reap it by ~10:30Z, after which the queued run starts on its own.

## The change itself

Unaffected and verified locally at `44cbfc5f` against current trunk: builds clean, `ownership_property_memo_preserves_decisions_across_repeats_and_recursion` passes, `scripts/rue quick` 30/30. The measurements in the description still stand — including that the win here is small, so closing this unmerged remains a perfectly good outcome if the complexity trade doesn't look worth it.

---

PR #2239, opened 2026-08-11
ADR-0073: agent-first test runner on the query graph

DorianListens:
Refs RUE-506.

Draft ADR proposing `rue test`, re-grounding the RUE-506 design capture against the compiler as it exists today. Intended for several rounds of iteration before acceptance; nothing is ratified.

## What the proposal commits to

- **Tests are language items**: `test "name" { ... }` blocks (contextual keyword, preview-gated as `test_declarations`), analyzed as ordinary demand-driven roots. Placement is the visibility model — tests beside the code see private items, tests in importing modules prove the public API. Executable requests never root them. Bodies are `()`-typed, and `?` in a test body's immediate block gets unwrap-and-report semantics: the failure arm emits a structured `unhandled_error` record — the error payload rendered by a compiler-synthesized structural printer, the `?` site's span attached — and traps, so attribution matches `@assert`, each `?` site stands alone (no identical-error-type constraint, no return-type annotation on the block), and the position it defines is a compile error today. Orphan-test detection is manifest-gated and backed by a bounded candidate-acquisition step in the ADR-0063 host input protocol: out-of-closure manifest entries are read and published as revisioned inputs (or typed absent/unreadable outcomes) and consumed by a parse-only query that never mints semantic roots.
- **`rue test` is the driver's first subcommand**, emitting a versioned NDJSON event stream on stdout as the primary surface (per ADR-0061 §6 schema policy); the human renderer is a consumer of the same stream. Discovery-without-execution via `--list`, stable IDs, repro argv on every failure, asymmetric verbosity, byte-safe bounded capture. `capability_summary` is present from event-schema v1.0 with an explicit `unavailable` status until Phase 3 populates it, so the MVP never contradicts its own zero-claims posture. The structured failure channel is a dedicated inherited pipe (recommended over stderr framing, which arbitrary user bytes could forge), budgeted separately from user streams.
- **Execution is a contract, not a mechanism**: MVP is one linked test image per target (compiler-synthesized dispatcher `main`, per-function `CodegenUnit` reuse) plus one process per test with process-group timeout/kill. Two inventories are pinned as contract values: the loader-visible one (constant `argv[0]`, fixed-width selector, pinned environment vector, run-constant image spelling — initial-stack consumption becomes deterministic, making the pinned `RLIMIT_STACK` a real boundary) and the test-visible one (what `std.env` observes after the dispatcher normalizes argv).
[...]

PR #2566, opened 2026-08-23
RUE-506: ADR-0083 — rue test MVP (test declarations, runner, event protocol)

DorianListens:
Refs RUE-506. Successor to #2239, scoped down per the review consensus there: this is the MVP ADR, and only the MVP ADR. The document stands alone — the process history is two sentences in Status, and reading #2239 is not required.

Rebased onto current trunk and renumbered: trunk claimed 0081 (incremental linking contract) and 0082 (Buck2) while this was in review, so the document is now **ADR-0083**. Marked **accepted for the MVP phase**.

## Maintainer calls, as ratified

All four acceptance-level calls were ruled on review 2026-08-23. Each is recorded at its site in the body and summarized under Open Questions; the Status section now names them as the scope of acceptance:

- **Tests are language items** — `test "name" { ... }` blocks, not a `@test` directive on a function (§1).
- **Discovery is the import closure** for the MVP, with the unimported-test-file warning; directory contents do not become compiler inputs. The convention-based alternative sits under Rejected alternatives, and the package-model end state stays a deferred question (§1).
- **`--filter` narrows the run set, never the analysis root set** (§2), so a filtered run's verdicts match the same tests' verdicts in a full run.
- **The structured failure channel is a dedicated inherited pipe** (§5.1), with the ABI call under ADR-0055.

Acceptance is explicitly scoped: it does not settle the lower-impact calls Open Questions marks decidable within their phase (exit codes, `@assert` stabilization, the `skipped` verdict's producer, the `scripts/rue test` homonym, naming), nor the memo-database-pressure spike Phase 2 requires. The deferred layers of §6 remain separate ADRs.

## What the MVP commits to

- **Test declarations**: contextual keyword, preview-gated as `test_declarations`, placement-is-visibility, tests rooted only by test requests. `?` in test bodies gets unwrap-and-report semantics with the skipped-destructor consequence accepted explicitly.
- **`rue test` as the driver's first subcommand**, with worked CLI examples in the text: NDJSON event stream v1.0, byte-safe bounded capture, asymmetric verbosity, stable IDs, repro argv on every failure, and `capability_summary` present from v1.0 with an explicit `unavailable` status.
- **Execution as a contract**: one test image per target, one process per test, both the loader-visible and test-visible inventories pinned to exact values.
- **Phase 2.5 structured assertion payloads** — agent-first in content, not just transport.
- **An explicit compiler-vs-build boundary section**: everything in the ADR is compiler behavior; build integration supplies exactly one optional input (`--test-candidates` for the orphan warning).

## What this defers, and where it went

Capability inference, hermetic verdict caching/selection, scheduling/flake policy, and the public provider protocol are deferred to focused follow-up ADRs — deferred, not rejected. §6 records what the MVP does now so each lands additively.

The full design capture stays on #2239 (closed unmerged, five review rounds). Follow-up issues seeded with the extracted sections: RUE-1621 (capability inference), RUE-1622 (verdict caching, blocked by 1621), RUE-1623 (scheduling, partially blocked by 1621), RUE-1624 (provider protocol) — Linear project "rue test follow-ups". MVP phases are RUE-1618/1619/1620 in project "rue test MVP", authorized to proceed by this acceptance.

Registry and doc-link gates pass (84 records).

### DIFF
--- a/docs/designs/0081-rue-test-mvp.md
+++ b/docs/designs/0081-rue-test-mvp.md
@@ -9,7 +9,7 @@ accepted:
 implemented:
 spec-sections: []
 superseded-by:
-relates: ["RUE-506", "RUE-505", "RUE-504", "RUE-438", "ADR-0063", "ADR-0061", "ADR-0058", "ADR-0055", "ADR-0064", "ADR-0027", "ADR-0025", "ADR-0069"]
+relates: ["RUE-506", "RUE-505", "RUE-504", "RUE-438", "ADR-0063", "ADR-0061", "ADR-0058", "ADR-0055", "ADR-0064", "ADR-0038", "ADR-0027", "ADR-0025", "ADR-0069", "ADR-0005"]
 ---
 
 # ADR-0081: `rue test` MVP: test declarations, runner, and event protocol
@@ -156,7 +156,8 @@ longer is:
   observation is best-effort — network and time are invisible to it — so the
   cache is unsound in exactly the corners users get burned by. Steal:
   cache-as-default UX, `test2json` as stream-first output. Avoid: observed
-  hermeticity; Rue verifies statically, at item granularity.
+  hermeticity; Rue's deferred answer (§6) is static verification, at item
+  granularity.
 - **Zig**: `test "name" { }` blocks as language items, discovered by the
   compiler; the build runner drives the test binary over a stdin/stdout binary
   protocol (execute-by-index, structured results); inferred error sets are the
@@ -178,8 +179,8 @@ longer is:
 - **Bazel**: the Test Encyclopedia states the hermeticity ideal — declared
   inputs only, `TEST_TMPDIR`, pinned env — but explicitly does not enforce it,
   and its caching is target-granular. Steal: the environment contract and
-  result-caching semantics. Rue's point of departure: enforcement, at item
-  granularity.
+  result-caching semantics. Rue's point of departure (deferred, §6):
+  enforcement, at item granularity.
 - **Buck2 / tpx**: tests handed to an external runner through an explicit
   protocol boundary — the runner is a client of the build system, not a
   subroutine. Validates RUE-506's "protocol, not a trait" conclusion, which
@@ -435,7 +436,7 @@ The driver grows its first subcommand:
 ```
 rue test <root.rue> [--list] [--filter <pattern>]... [--format human|json]
          [--jobs N] [--timeout-ms N] [--shard K/N] [--target <t>] [-O<n>]
-         [--seed N] [--keep-going] ...
+         [--seed N] ...
 ```
 
 - **Dispatch rule**: the driver enters test mode when the first argument
@@ -470,7 +471,10 @@ rue test <root.rue> [--list] [--filter <pattern>]... [--format human|json]
   deferred work (§6) and are not producible in the MVP, whose whole-run
   compile failure is exit code `2`. A failure record is data: failure kind
   (`assert` / `unhandled_error` / `trap:<class>` / `exit` / `signal` /
-  `timeout` / `output_overflow` / `ice`), the pinned runtime message (the
+  `timeout` / `output_overflow`; `ice` is reserved alongside the per-test
+  `compile_error` verdict, on the same grounds — in the MVP a compiler crash
+  is a whole-run failure on stderr, not a per-test event), the pinned
+  runtime message (the
   abort-only runtime's fixed stderr strings are machine-recognizable by
   construction), exit code or signal, and a source location — in the MVP, the
   test declaration's span, except `unhandled_error`, whose record carries the
@@ -714,12 +718,14 @@ The MVP mechanism:
   choices (shuffle order, scratch naming). Making `@random_*` seedable in
   test builds — lowering to a seeded PRNG per test process — is a
   runtime/codegen change whose maintainer call is deferred with scheduling
-  and caching (§6): it would make `random`-observing tests deterministic and
-  hence cacheable, at the cost of test-mode divergence from production
-  entropy.
-- Test execution order within a run is shuffled by seed by default (verified
-  isolation makes order dependence impossible for hermetic tests; shuffling
-  keeps everyone else honest, and the seed makes any surprise reproducible).
+  (§6; its cache-key interaction travels with RUE-1622): it would make
+  `random`-observing tests deterministic and hence cacheable, at the cost of
+  test-mode divergence from production entropy.
+- Test execution order within a run is shuffled by seed by default
+  (shuffling keeps order dependence visible, and the seed makes any surprise
+  reproducible; once capability inference lands (§6), verified isolation
+  makes order dependence impossible for hermetic tests and shuffling keeps
+  everyone else honest).
 
 ### 5. Extensibility is tiered, in-language first, with no privileged built-ins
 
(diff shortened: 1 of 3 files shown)


DorianListens:
Pushed `345daade`: fixes from an adversarial review pass (Fable subagent) over the distillation. All nine findings were remnants of the pre-narrowing scope rather than new design content; the grounding spot-checks (E0503/E0505, abort-101, `rue-test-runner` mechanics, both manifest scripts, the 11 value-taking flags, the 46-helper manifest, ADR-0061/0063 citations) all verified clean against current trunk.

**The four that mattered:**

1. **§6 scheduling bullet miscounted its own deferral list** — it named five items ("…and the seedable `@random_*` maintainer call") while keeping the original's "two of its four items require capability inference," making the dependency unrecoverable. Now four items with the gated two named explicitly (`--reruns`-for-non-hermetic, hermetic-mismatch reporting) and the seedable-`@random_*` call attached as its own sentence.
2. **§4's shuffle bullet claimed present-tense verified isolation** the MVP disclaims — the one un-rewired hermeticity claim left in a Decision section. Now: shuffling keeps order dependence visible today; the impossible-for-hermetic-tests claim arrives with inference.
3. **Per-test `compile_error` verdicts were falsely gated on capability inference** by the blanket "Requires capability inference" on the RUE-1622 bullet — §3's mechanism (independent per-test roots, exclusion not stubbing) needs no summaries and no cache. Now scoped: caching/selection items require inference; the `compile_error` mechanism may land independently.
4. **The seedable-`@random_*` call was deferred to two different places** (§4 said scheduling *and caching*; §6 said RUE-1623). Now consistent: the call is RUE-1623's; its cache-key interaction travels with RUE-1622.

**Smaller:** the Go/Bazel prior-art bullets no longer assert "Rue verifies/enforces" in present tense; Phase 2's bold note becomes "the minimum usable runner" so "the MVP" consistently means Phases 1–2.5 as the title and Summary use it; `ice` is reserved alongside `compile_error`/`cached_pass` per the document's own unproducible-verdict principle (in the MVP a compiler crash is a whole-run failure on stderr, not a per-test event); the dangling HyRTS citation (it supported the dropped selection-economics argument) is removed; ADR-0038 and ADR-0005 join `relates`, both load-bearing in the body; undefined `--keep-going` leaves the synopsis.

Registry and doc-link gates still pass (82 records).

DorianListens:
Pushed `4426d161`: a full editorial pass per maintainer notes, 1,259 → 846 lines. Four changes of substance beyond compression:

1. **The document now stands alone.** Process history is two sentences in Status; Context is reduced to the facts the design is built on plus one-lesson-each prior art.
2. **The UX is shown, not described**: worked `rue test` invocations with illustrative human and NDJSON output open §2, and §1 shows the import-wiring idiom as code with its tradeoff stated plainly.
3. **A new top-level boundary section** pins compiler-CLI vs build-integration: everything specified is compiler behavior; the build supplies exactly one optional input (`--test-candidates`).
4. **File discovery is now an explicit maintainer call** (before Phase 2): import-closure + orphan warning as drafted, vs a closure-anchored naming convention that auto-roots conventional test files — bounded probe over closure directories, not a walk; kills the Zig wiring papercut for sibling tests and makes orphan detection self-contained, at the cost of a spec-pinned name pattern and (until per-test `compile_error` lands) whole-run blast radius for a broken conventional file. The package-model end state (declared sources as canonical inventory, auto-rooted) is recorded under deferred questions.

Every contract value, reserved schema surface, and adversarial-review fix from `345daade` is preserved. Gates pass.

DorianListens:
Pushed `2e7be7a8`: one ruling taken, and Open Questions reorganized by impact.

**Ruled: no implicit preview enable.** `rue test` requires `--preview test_declarations` like every other preview feature. The deciding observation: the gate covers a parser change, so any request whose closure contains test items — executable builds included, which parse test items for the warnings scan — needs the flag to compile at all. Auto-enabling in test mode alone would leave those same files failing ordinary builds while making `rue test` the first flag to bypass ADR-0005's explicit opt-in, for no net convenience. §1's preview-gate bullet now records the decision and rationale; the open question is removed.

**Open Questions now splits into two groups.** "Needs a ruling before acceptance": the test declaration surface, the file discovery mechanism, and `--filter` root-vs-run semantics — the three calls that shape the language surface and discovery machinery. "Lower-impact decisions (decidable within their phase)": failure-channel mechanism, `@assert`/exit-code stabilization, the 0/1/2/3 contract, the `skipped` verdict, the `scripts/rue test` homonym, and naming.

steveklabnik (dismissed):
Codex review, requested by Steve.

The overall MVP boundary is much clearer after the editorial pass, and I recommend resolving the three acceptance-level questions as follows: keep the test "name" { ... } surface; use import-closure discovery for the MVP (with the optional declared candidate inventory warning) rather than adding directory contents as compiler inputs; and have --filter narrow the run set, not the analysis root set. The dedicated inherited failure pipe is also the right Phase 2 mechanism.

I found two correctness/sequencing gaps that should be resolved in the ADR:

1. Phase 1 says test-body ? becomes legal and test bodies semantically analyze, while the failure-arm lowering is deferred to Phase 2. Current semantic analysis does not have a separable legality-only step: analyze_try immediately builds the Option/Result match and early-return AIR against the enclosing return type. A ()-typed test body therefore cannot analyze successfully until the test-specific failure arm exists. Move legality and lowering together into Phase 2, or specify a real Phase 1 lowering that makes the claimed emit-level analysis possible.

2. The accepted std.exit(0) blind spot makes a test that exits before its assertions report as a pass. With one process per test and the dedicated parent-child channel already in the MVP, require a terminal completion record that only the dispatcher writes after the test body returns normally. Exit 0/EOF without that record should be a runner failure. This is false-positive test evidence, so it should not be deferred as optional hygiene.

The ADR-registry and documentation-link gates pass at this head.

— Codex

DorianListens:
Both gaps resolved and all four rulings recorded, as `f245ecac1`.

**Gap 1 — the Phase 1/2 `?` split.** You're right that there's no separable legality-only step: `analyze_try` builds the `Option`/`Result` match and early return against the enclosing return type as it analyzes, so a `()`-typed body can't analyze successfully until the test-specific failure arm exists. Taken as recommended: legality and lowering move together into Phase 2, Phase 1 keeps today's compile error for test-body `?`, and §1 states the mechanism reason so the constraint doesn't get re-derived at implementation time.

**Gap 2 — the `std.exit(0)` blind spot.** Agreed it's false-positive evidence, not hygiene debt. The dispatcher now writes a terminal completion record on the dedicated channel after the body returns normally, and exit 0 with end-of-stream but no record is a failure with new kind `incomplete`. The pass definition (§1), the event sketch (§2), the execution contract (§3), and Phase 2's item all carry it. One honest caveat is stated where the mechanism lives: the channel is not a security boundary, so deliberate forgery remains possible — the defect closed is the accidental early exit.

**Rulings recorded**: `test "name" { ... }` blocks; import-closure discovery for the MVP with directory contents staying out of compiler inputs (the convention alternative is preserved under Rejected alternatives with the package-model revisit named); `--filter` narrows the run set, never the root set (§2 now states the verdict-stability rationale inline); dedicated inherited pipe. Open Questions' front group is now the ruling record, and the Status section reflects it.

Registry and doc-link gates pass at this head.

DorianListens:
Pre-merge read-through against current trunk. The grounding claims check out (the eleven value-taking driver flags, `rue-test-runner`'s 10 s default and process-group kill, the live `fn test` in `std/bitset.rue`, spec 4.15:3/4, 10.3, 4.13:5b, identifier-only directive args, the runtime's argv/envp capture, no clock API), the branch sits on current trunk, and the registry gains the 0082 row it was missing. Merging as accepted. Four observations recorded for the phases rather than the document; none changes the rulings.

1. **Preview-gate placement is a choice the text does not name.** ADR-0005 and AGENTS.md put preview gates in semantic analysis via `require_preview()`, and executable requests never analyze test bodies under §1. With the standard placement only `rue test` would need `--preview test_declarations`. §1 instead gates at the parser, so every closure containing a test item needs the flag for ordinary builds: std cannot carry `test` blocks until stabilization, and every `rue_program` example that adds tests must set `preview_features`. Worth recording as a deliberate deviation with its reason when Phase 1 lands (noted on RUE-1618).

2. **Per-test `compile_error` is stranded behind a blocked issue.** §6 says it needs no capability inference and may land independently, but it lives inside RUE-1622, which is blocked by RUE-1621. For agent loops this is the sharpest edge in the MVP: one broken unrelated test file turns every filtered run into exit 2. Recommend splitting it into its own issue queued directly after Phase 2 (noted on RUE-1622).

3. **Repro argv should filter on the full stable ID.** The §2 worked example filters on the bare name, but names are unique per module, not per program, so a repro for one test can run several. Pin in the Phase 2 schema doc (noted on RUE-1619).

4. **RUE-1618/1619/1620 are still Backlog.** The Status section authorizes them, and AGENTS.md lets autonomous work claim only Todo, so RUE-1618 should move to Todo once this merges.

PR #2608, opened 2026-08-23
RUE-1543: gate on every declared baseline resolving, retired epochs included

DorianListens:
Refs RUE-1543. Split out of #2444 per the coordination note there: this is the independent integrity safeguard, separated from the ADR amendments so it can land without waiting on any policy ruling.

## The silent failure mode this closes

A wrong baseline address fails silently today — for a *retired* epoch. `derive` publishes no index and no ratios for the affected platform while still plotting every series, and `staleness-inputs` deliberately keeps retired epochs out of the data `derive` reads (RUE-1542), so `unindexed()` can never see the loss. Six of the nine declared baselines sit on retired epochs 2 and 5, squarely in that blind spot.

This exists independent of any compaction decision: a bad edit to `performance/manifest.toml` or a corrupted `index.json` entry hits it today. If #2444's append-based re-encode is later accepted, every address in the store moves and all nine pins get re-pinned — at which point this gate stops being merely useful and becomes the stated acceptance condition. It is deliberately mergeable first.

## Mechanism

`rue-bench check-baselines --manifest … --index …` resolves every baseline declared in the manifest against `index.json` — every epoch, retired included, requiring a record of the baseline's own epoch and platform, which is exactly where `derive` looks it up — and exits 3 naming the epoch on any failure. The CI staleness job runs it against the `index.json` it has already checked out (one pass over a file the step is holding anyway; the selection path stays narrowed per RUE-1542). `validate-ci-gate` pins the responsibility so it cannot be silently dropped from the workflow.

Index-level rather than selection-level on review guidance in #2444: selecting every baseline-bearing epoch into the derive input would restore the cost RUE-1542 removed — on 2026-08-18, 1,437 of 1,440 records against the 321 the gate then read; at `44b5d5164` (2026-08-23), 1,616 of 1,619 against 500.

## Verification

- Seven unit tests over the resolution rule (missing record, wrong epoch, wrong platform, retired epoch, no-baseline state) plus the report path end to end: a simulated compaction — live epoch resolving, retired epoch pinned to a pre-compaction address — exits `NOT_APPENDABLE` (3) naming exactly the retired epoch.
- Against the live store at `44b5d5164` (2026-08-23): 9/9 declared baselines resolve, exit 0.
- Full `rue-bench` suite: 165 passed, 2 ignored, 0 failed; `validate-ci-gate` and its test suite pass with the new responsibility under the RUE-1266 lane structure.

https://claude.ai/code/session_01NPBpmZU3DpwkumLHNwHsgu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,6 +653,16 @@ jobs:
             git reset --quiet
           fi
           echo "reading $(wc -l < "$SELECTED") record(s) from the live epoch"
+          # Baseline resolution is asked here rather than in the derived data,
+          # because the selection above deliberately excludes retired epochs and
+          # a retired epoch's baseline is exactly the one whose loss is silent:
+          # `derive` publishes no index and no ratios while still plotting every
+          # series, and `unindexed()` iterates live epochs only. index.json is
+          # already checked out, so this covers every epoch for one pass over a
+          # file the step is holding anyway.
+          "$BENCH" check-baselines \
+            --manifest performance/manifest.toml \
+            --index "$DATA/index.json"
           DERIVED="$(mktemp)"
           "$BENCH" derive \
             --manifest performance/manifest.toml \
--- a/crates/rue-bench/src/check_baselines.rs
+++ b/crates/rue-bench/src/check_baselines.rs
@@ -1,0 +1,281 @@
+//! Answer "does every declared baseline still name a record that exists?".
+//!
+//! `derive` resolves an epoch's baseline by address among that epoch's own
+//! records (`derive.rs:1316`). A miss is not an error there: the epoch simply
+//! publishes no index and no workload ratios while still plotting every
+//! per-workload series, so the dashboard keeps moving and loses the one number
+//! it exists to publish. `validate-performance-stall.py`'s `unindexed()` gate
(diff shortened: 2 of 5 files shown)


DorianListens:
Pushed `be347c2af` from an adversarial review pass (Fable subagent). The substantive finding: the PR body claimed a simulated-compaction unit test naming the retired epoch and exiting 3, but all six tests exercised only the pure `unresolved()` rule — nothing covered `report()`, the exit code, or the message. That test now exists (live epoch resolving beside a retired epoch pinned to a pre-compaction address; asserts exit `NOT_APPENDABLE` and that exactly the retired epoch is named). Also fixed: the dangling "the compaction" comment gets its RUE-1543 antecedent for trunk readers, the module doc's dated selection-cost figure no longer says "now", and the body's store figures are stated per-date (08-18: 1,437/1,440 vs 321; 08-23: 1,616/1,619 vs 500).

Verified clean in the same pass, for the record: retired-epoch coverage via `Manifest::epochs()`, no vacuous-pass path except the documented zero-baselines state, the `staleness_inputs.rs` diff is pure visibility/extraction with byte-identical selection, CI wiring under `set -euo pipefail` with `$BENCH`/`$DATA` in scope ahead of `derive`, and 9/9 live baselines re-verified independently against the manifest and index.

PR #2613, opened 2026-08-23
RUE-1707, RUE-1708, RUE-1709: fix std traps at argument extremes

steveklabnik:
Three std entry points failed on values their signatures accept. All three were reproduced on an unmodified checkout first, and every fix was re-verified by hand on the integrated tree.

## RUE-1707 — `to_radix` did not validate its base

`std/fmt.rue` documented "base 2..=16" and enforced nothing:

| call | before | now |
|---|---|---|
| `to_radix(5, 1)` | **infinite loop** (`m /= 1` never decreases `m`) | `panic: to_radix: base must be 2..=16` |
| `to_radix(5, 0)` | division-by-zero trap | same guard panic |
| `to_radix(16, 17)` | index-out-of-bounds trap | same guard panic |
| `to_radix(255, 17)` | silently printed `"f0"` | same guard panic |

The last row is the worst of them: bases ≥ 17 only misbehave when a digit exceeds 15, so misuse was data-dependent and could pass unnoticed. An entry guard turns a hang, two unrelated traps and a silent wrong answer into one diagnosable message. `to_radix(255, 16)` still prints `ff`.

## RUE-1708 — overflow traps at type extremes in `std/math.rue`

`is_prime`'s bound `i * i <= n` overflowed for a prime near the type maximum: the loop only exits once `i` passes `isqrt(n)`, and the square traps before the comparison can end it. `is_prime(i32, 2147483647)` trapped instead of returning `true`. The bound is now `i <= n / i` — the same overflow-free form `isqrt` already uses — which fixes every `T` at once; `i >= 2` throughout, so it never divides by zero.

`gcd` took `|a|` and `|b|` up front, trapping for a `T::MIN` operand even when the answer was representable (`gcd(i64::MIN, 2) == 2`). Euclid now runs on the signed values with the sign normalised at the end. Rue's `%` truncates toward zero, so remainder magnitude still strictly decreases and termination is unchanged. Magnitude-1 operands are answered up front because `T::MIN % -1` traps on quotient overflow, and that is the only pairing the loop could reach with `T::MIN` as dividend. `gcd` now panics only when the true gcd is `|T::MIN|`, which no signed `T` can hold.

That rewrite also fixes `gcd` at **unsigned** `T`, where it never compiled at all. Unary `-` is rejected outright on an unsigned type (E0801) even in a branch unsigned `T` can never reach, so the old body failed to compile for every unsigned instantiation while its doc promised any `T`. Verified against pristine trunk `std/`:

```
error: [E0801]: cannot apply unary operator `-` to type 'u64'
  --> std/math.rue:60:28
```

The negations are spelled `0 - x` for that reason.

`lcm` rejects a non-zero `T::MIN` operand up front: `|lcm(T::MIN, x)|` is `2^(bits-1)` times an odd factor for every non-zero `x` and is never representable. That check is also what makes the closing negation safe.

## RUE-1709 — `IntMap` trapped on key `i64::MIN`

`_slot` hashed negative keys as `@intCast(0 - k)`, which overflows at `i64::MIN` and made that single value unusable as a key on a map documented to take any `i64`. `_key_bits` reinterprets the two's-complement bit pattern instead, spelling `|k| - 1` as `0 - (k + 1)` so every intermediate stays in range.

The hash itself moves into one module-level `_hash`. `_grow_to` carried an **inline copy**, and while it did, a hash fix applied to `_slot` alone would place keys by the new hash and re-place them by the old one on growth, corrupting every later lookup. The extraction is part of the fix, not tidying — worker verification deliberately tested the detector by reverting `_grow_to` alone and watching the growth roundtrip drop from 400 to 225.

## Verified by hand on this tree

```
to_radix(5,1)            -> panic: to_radix: base must be 2..=16, exit 101   (was: hang)
to_radix(5,0) / (255,17) -> same guard panic                                  (was: trap / silent "f0")
to_radix(255,16)         -> ff                                                (control)
is_prime(i32, 2147483647)-> true                                              (was: overflow trap)
gcd(i64, i64::MIN, 2)    -> 2                                                 (was: overflow trap)
gcd(u64, 48, 18)         -> 6                                                 (was: E0801, did not compile)
MIN key + 64 inserts forcing growth -> min_after_growth=7, len=66             (was: trap)
```

Suites: `cli std_fmt_radix` 9/9, `cli std_collections` 5/5, `cli std_core` 9/9, `scripts/rue fmt` clean. Full `./test.sh` result is posted below once it completes; this PR stays a draft until then.

## One deliberate omission

`intmap_key_i64_min_full_lifecycle` is left **without** `differential_opt`, with an in-file comment explaining why, because of a pre-existing **-O2/-O3 miscompile** filed as RUE-1758: an `Option` `None` returned from a struct method is read as `Some(filler)` when a runtime-conditional `StrBuf` concat precedes the match. It reproduces on pristine trunk `std/` with an ordinary key 42 — no `i64::MIN`, no growth — so marking the case would report an optimizer bug as a std-library regression at the wrong address. The other two new IntMap cases do carry `differential_opt` and pass at all four levels.

Also registers 13 oracle model-gap entries that the fail-closed CLI corpus audit required.

Fixes: RUE-1707
Fixes: RUE-1708
Fixes: RUE-1709

---

### DIFF
--- a/crates/rue-cli-tests/cases/std_collections.toml
+++ b/crates/rue-cli-tests/cases/std_collections.toml
@@ -155,3 +155,226 @@ fn main() -> i32 {
     42
 }
 """ }]
+
+# ── IntMap key extremes (RUE-1709) ──────────────────────────────────────────
+#
+# IntMap is documented to take any `i64` key, but until now every case above
+# used small NON-NEGATIVE keys — the suite had no negative-key coverage at all.
+# That left two things unpinned:
+#
+#   1. `_slot` hashed a negative key by its MAGNITUDE, `@intCast(0 - k)`, and
+#      `0 - i64::MIN` overflows. `@int_min(i64)` was a single-value landmine:
+#      an ordinary -5 key worked, so nothing looked wrong until exactly that
+#      key showed up. The hash now reinterprets the key's two's-complement bit
+#      pattern, which has no such hole.
+#   2. `_grow_to` carried a SECOND, inline copy of that same hash. A fix
+#      applied to `_slot` alone would place keys by the new hash and re-place
+#      them by the old one on the next growth — every lookup after a growth
+#      silently wrong. The hash now lives in one module-level `_hash`, and the
+#      growth case below is what would catch it diverging again.
+
+# Insert / lookup / overwrite / remove / re-insert, all at i64::MIN, which
+# previously trapped with "integer overflow" on the very first insert.
+#
+# NOT `differential_opt`, and deliberately so: this case's own shape trips a
+# SEPARATE, PRE-EXISTING optimizer bug that has nothing to do with the key. On
+# trunk's unmodified `std/intmap.rue`, at -O2 and -O3 only,
+#
+#     let mut m = M.new(0);
+#     m.insert(42, 7);
+#     println("contains=" + std.fmt.bool_to_string(m.contains(42)));
+#     match m.get(99) { O.Some(v) => ..., O.None => ... }   // takes Some(0)!
+#
+# reads a `None` returned by `IntMap.get` as `Some(filler)`. It needs no
+# negative key, no i64::MIN and no growth — an ordinary key 42 reproduces it —
+# and it disappears if the `bool_to_string(<call>)` concatenation before the
+# lookup is removed. Marking this case `differential_opt` would report that
+# optimizer bug as a std-library regression at the wrong address, so the
+# opt-level sweep is left off HERE and carried by the two cases below, whose
+# assertions do not sit downstream of that shape. See the meta note filed with
+# RUE-1709 for the standalone repro.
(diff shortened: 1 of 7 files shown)


steveklabnik:

```
Tests finished: Pass 231. Fail 0. Timeout 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
=== TEST SUITE: PASSED ===
TESTSH_EXIT=0
```

Run in the foreground and unpiped, with the exit code captured to a file rather than read from a pipeline's status. A first attempt was killed mid-run while compiling `caldera` and wrote no exit marker; this is a clean rerun, not that run's output.

Agrees with CI on this head (`269dcda3`), which was already green across all 23 checks.

---

PR #2626, opened 2026-08-23
RUE-1711: reject fs paths holding a NUL byte instead of truncating them

steveklabnik:
Fixes RUE-1711.

## The bug

`_path_to_cbuf` copied a `StrBuf` path into a NUL-terminated buffer without looking at the bytes it copied. A path carrying a NUL of its own has no C-string spelling, so the terminator it appends is not the first one the kernel meets and the syscall reads the **prefix**.

Verified on `1d657d4` with the real std, before the change:

| call | result | what actually happened |
| --- | --- | --- |
| `fs.remove("victim\0decoy")` | `Ok(())` | `victim` was deleted |
| `fs.create_dir_all("nulparent\0x/y")` | `Ok(())` | `nulparent` was created |
| `fs.read_dir("listed\0decoy")` | `Ok(entries)` | listed `listed` |

Nothing about any of these is visible to the caller — each reports success against a name the caller never wrote — and a program deriving paths from untrusted bytes, which is what byte-string paths invite, can be steered into it.

## The change

Reject such a path as `FileError.InvalidInput`, the answer Rust's `CString`-based `std::fs` gives the same input. The rule lives in one predicate, `_path_has_interior_nul`, that every path-taking entry point consults *before* marshalling, so the file the prefix names is never touched:

- `File._open_with` — covers `open` / `create` / `append`, and `read_dir` / `walk` reach the kernel through it
- `_metadata_at` — covers `fs.metadata` and the `DT_UNKNOWN` kind fallback
- `rename`, on **both** operands — a clean `from` does not license a `to` the kernel would truncate
- `remove`, `create_dir`, `create_dir_all`, `remove_dir`

`create_dir_all` checks ahead of its prefix loop rather than inside it. It is the one operation whose documented non-atomicity would otherwise let a rejected path leave real directories behind, and before this its loop was exactly what minted `nulparent`.

`_mkdirat_raw` is left unguarded and documented as such: both its callers check first, and every path reaching it from `create_dir_all` is a prefix of one already checked.

## Tests

Three CLI cases, each failing on the prior std with the operation reporting success (exit codes 63, 61, 62 respectively):

- `fs_interior_nul_path_rejected` — the `remove` repro plus every other entry point, asserting the victim file survives and the rejected renames did not half-happen
- `fs_interior_nul_create_dir_all_creates_nothing` — rejection precedes any `mkdirat`, so not even the first component appears
- `read_dir_interior_nul_path_rejected` — `read_dir` and `walk`, and the directory still lists by its real name

The NUL is pushed as a byte rather than written as `\0`: these sources are TOML basic strings, where `\0` is not a valid escape.

`scripts/rue premerge` — 200 passed, 0 failed. `scripts/rue cli fs_` — 35 passed. `//crates/rue-oracle-diff:oracle-diff-test-action` — passes.

## Notes

ADR-0057 §2 gets the decision recorded alongside the existing path-marshalling paragraph it qualifies.

The second commit registers the three new cases in the oracle-diff model-gap inventory. All three build their paths with `StrBuf`, so they reach the same `@byte_copy` gap the rest of the std.fs group already carries, and an unregistered gap is a harness failure rather than silent debt. The debt is incidental to what they test — the rejection they assert happens in Rue, above the syscall.

Scope is `std.fs` only. `std.c`'s `free_c_string` has its own interior-NUL problem, tracked separately as RUE-1710, and is untouched here.

### DIFF
--- a/crates/rue-cli-tests/cases/fs_file_io.toml
+++ b/crates/rue-cli-tests/cases/fs_file_io.toml
@@ -1,7 +1,8 @@
 # V1 FOLLOW-UPS (ADR-0057 "Future Work"): the cases whose names begin `fs_seek_`,
 # `fs_stat_`/`fs_metadata_`, `fs_rename_`, `fs_remove_`, and `fs_mkdir_` exercise
 # seek/tell, fstat/newfstatat metadata, rename, unlink, and directory
-# create/remove. The `fs_mkdir_` group also covers `create_dir_all` (RUE-995):
+# create/remove. The `fs_interior_nul_` pair covers the path PRECONDITION those
+# entry points share (RUE-1711) rather than any one syscall. The `fs_mkdir_` group also covers `create_dir_all` (RUE-995):
 # nested creation, idempotent re-creation, the separator shapes its byte scan has
 # to collapse, and the EEXIST-but-not-a-directory case.
 # They are UNGATED, like the v0 group, because each one issues a syscall whose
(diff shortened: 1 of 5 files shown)


steveklabnik:
Closing as superseded — #2625 landed the same fix first and RUE-1711 is Done. Three of us raced this issue; #2627 was closed for the same reason.

The merge queue dequeued this with `MERGE_CONFLICT` against the already-landed change, which is the correct outcome. Trunk's `_path_has_nul` guards the same set of entry points this did, so nothing here is missing from trunk.

One thing that is worth noting rather than losing: neither merged change covers `read_dir`/`walk` with a test — `fs_read_dir.toml` has no NUL case. Both entry points *are* guarded, but only transitively through `File._open_with`, so a future refactor giving `read_dir` its own open path would regress silently. This PR had a `read_dir_interior_nul_path_rejected` case (verified failing against the pre-fix std) if anyone wants to pick it up.

---

steveklabnik:
Superseded by #2625, which merged the same fix for RUE-1711 (`c12ca68`, `46f1b58`). Three sessions picked this issue up in parallel; #2627 was the third and is likewise closed unmerged.

Closing rather than rebasing: the branch is conflicting now, and what landed covers the same ground — the NUL check at every marshalling site, `create_dir_all` guarded ahead of its prefix loop so a rejected path creates no parents, both `rename` operands, and the ADR-0057 §2 record.

One difference worth noting for anyone reading later: this PR asserted `read_dir`/`walk` in a case of their own, whereas #2625 folded them into its all-entry-points case. Same coverage, different packaging. RUE-1711 is Done.

---

PR #2627, opened 2026-08-23
RUE-1711: reject fs paths holding an interior NUL byte

steveklabnik:
## Summary

- reject byte 0 anywhere in an `std.fs` path with `FileError.InvalidInput`, before the path is marshalled
- check at the marshalling call sites, so every path-taking entry point is covered — open/create/append, metadata, rename (either operand), remove, create_dir, create_dir_all, remove_dir, read_dir, walk
- record the path-contract refinement as ADR-0057 section 2a

## Problem

`_path_to_cbuf` copies a `StrBuf` into a NUL-appended buffer and hands the kernel that pointer, which reads it as a C string — so a path holding byte 0 named only the bytes before it. The operation succeeded, reported nothing unusual, and landed on a path the caller never wrote:

```rue
// victim.dat exists
fs.remove(borrow sneaky)   // "victim.dat" + 0 + "totally_different_file"
// -> Ok(()), and victim.dat is gone
```

Byte-string paths invite exactly the program that hits this: one assembling a path from bytes it does not control.

## Approach

The check is at the marshalling call sites rather than inside `_path_to_cbuf`, because that function returns a bare pointer with no way to spell a failure. `_mkdirat_raw` reports it on its raw errno channel as `-EINVAL`; 22 agrees across Linux and macOS and both tables already normalize it to `InvalidInput`, so its callers get the same error the `Result`-returning entry points return directly.

Two cases a "just reject it" fix can still get wrong, both covered:

- **A trailing NUL is rejected too**, though its C-string form is a valid path — the buffer the kernel reads is not the byte string that was passed, and the caller can spell that path unambiguously already.
- **`create_dir_all` checks the whole path once up front**, so a rejected path creates none of the parents preceding the offending component, rather than creating them and failing at the bad one.

## Validation

- `scripts/rue premerge`: 200 passed, 0 failed
- `scripts/rue cli nul`: 7 passed (the 3 new cases plus the existing NUL-named ones)
- `scripts/rue fmt`, `buck2 test //:adr-registry-validation`
- The three new CLI cases each **fail against the pre-fix std**, verified by reverting `std/fs.rue` and re-running: the behavioral case exits 1 having removed the victim file, and the two coverage cases report the truncated call's own error instead of `InvalidInput`.

Fixes RUE-1711

---
_Generated by [Claude Code](https://claude.ai/code/session_01KcQ73d3FmQYaxAUh6Lzi9C)_

### DIFF
--- a/crates/rue-cli-tests/cases/fs_file_io.toml
+++ b/crates/rue-cli-tests/cases/fs_file_io.toml
@@ -1013,3 +1013,131 @@ fn main() -> i32 {
 """ }]
 stdout = "1\n2\n"
 exit_code = 0
+
+# INTERIOR NUL (RUE-1711). Paths are byte-string `StrBuf`s, and the kernel reads
+# the marshalled buffer as a C string — so a path holding byte 0 anywhere would
+# name only the bytes before it. This case is the behavioral one: `remove` on
+# "victim.dat" + 0 + "totally_different_file" must NOT unlink `victim.dat`. It is
+# a real regression test, not a shape assertion — before the check it exited 1,
+# having removed a file the caller never named.
+[[case]]
+name = "fs_nul_path_does_not_remove_the_truncated_prefix"
+env = { RUE_STD_PATH = "${REAL_STD}" }
+files = [{ path = "main.rue", source = """
+const std = @import("std");
+
+fn main() -> i32 {
+    let F = std.fs.File;
+    let FE = std.fs.FileError;
+    let R = std.result.Result(F, FE);
+    let CR = std.result.Result((), FE);
+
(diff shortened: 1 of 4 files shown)


steveklabnik:
Closing as a duplicate: three sessions picked RUE-1711 concurrently and produced the same fix. #2626 is the one to land — Steve picked it. #2625 is the third.

One thing from here that #2626 may not cover, worth a look before it merges: a **trailing** NUL (`"trailing.dat"` + byte 0). Its C-string form is a valid path, so a check written as "reject an *interior* NUL" accepts it and silently operates on `trailing.dat`. This PR rejected it on the same rule and asserted so in `fs_nul_path_rejected_by_every_entry_point`.

---

PR #2628, opened 2026-08-23
RUE-1758, RUE-1699: fix the -O2 lowering-order miscompile; make type identity linear

steveklabnik:
Two independent fixes in disjoint files. Both were reproduced on an unmodified checkout first, and both were re-verified by hand on this integrated tree.

## RUE-1758 (Urgent) — `None` read as `Some(filler)` at -O2/-O3

A real `None` returned from a struct method took the `Some` arm and printed uninitialized memory. Silent wrong-answer codegen on both backends.

**It is not the optimizer's transforms and not the standard library.** `terminator_plan::lower_cfg` visited blocks in CFG *arena* order, while lowering materializes a value into whichever block first *demands* it. The -O2 inline splice moves everything after a call into a fresh continuation block carrying the **highest** id, while the call block's own successors keep low ids. Arena order therefore reached the continuation last: `bb1` lowered `bb6`'s `call get` into itself, and `bb6`'s `switch` was left comparing a vreg never defined on that path. Whichever garbage the compare saw decided the arm.

Every hypothesis in the original report was **refuted** during diagnosis: the StrBuf concatenation, std involvement, and slot-reuse/dead-store elimination. The actual requirement is only that a call small enough for the -O2 inliner sits between the method call and the match.

`Cfg::block_lowering_order()` walks the dominator tree topologically, taking the lowest available block id. Sound because the CFG verifier already requires every definition to dominate its uses, so a dominator-respecting order always reaches a definition before any demand for it. It coincides with arena order whenever arena order was already valid — codegen for correct programs is unchanged, verified byte-identical at -O1 on both x86-64 and AArch64. The order is shared code, so both backends are fixed together; both carried the identical defect.

The regression corpus is **std-free and reduced**: a hand-written enum, a hand-written struct method, one inlinable free function. It covers `Option(bool)` and a three-slot struct payload (so the defect is not payload-width specific), and every case was checked to fail against a rebuilt pre-fix compiler.

`intmap_key_i64_min_full_lifecycle` carries `differential_opt` again — it was withheld pending this fix — and its explanatory comment is gone.

## RUE-1699 — anonymous-type identity was exponential in nesting depth

`AnonymousNominalKey` carried both a `producer` and its own `arguments`, and the producer of `Pair(T)`'s result is itself a specialization carrying those same arguments — so each nesting level embedded the inner identity twice. Hashing, ordering, cloning and identity remapping were all O(2^k).

`Pair^16(i32)` took 15.6s; `Pair^20` never finished, pegged at 100% CPU with no diagnostic. Worse, `MAX_SPECIALIZATION_ROUNDS` could never fire — a self-growing `f(comptime T: type, …) { f(Pair(T), …) }` hung forever instead of reporting the depth-64 error, because per-level work doubled long before round 64. A guard that cannot fire is not a guard.

Memoization was tried first and **rejected**: it bought a 3× constant factor and left the curve doubling, because three further whole-tree walks remained (durable encoding, retained charge, structural equality) and the FNV-1a name digest cannot be memoized at all — its state is not composable across subtrees.

The field is removed instead. Every mint site was checked to build the producer and the arguments from one source, so the duplicated copy was never independent information.

| depth | before | after |
|---|---|---|
| K=16 | 15.6s | — |
| K=20 | hang (>180s, no diagnostic) | 0.17s |
| K=60 | — | 0.77s |
| depth-64 E1200 | hung forever | fires in 1.85s |

**The equivalence relation is unchanged**, which is the part that matters — this is a correctness change wearing a perf costume. Verified by running: types differing in argument, in nesting depth, or in **producer** (`Pair(i32)` vs a structurally identical `Twin(i32)`) remain distinct, and independent spellings of one nesting remain the same type. Removing a field shifts every `__anon_struct_{digest}` name once, so the pinned durable-encoding table and two standalone digest pins were re-derived, and the deliberate digest collision in the lookalike case was re-checked by hand to confirm it still collides.

## Verified by hand on this tree

```
RUE-1758  -O0/-O1/-O2/-O3 -> get=none        (was: get=0 at -O2/-O3)
          key present, -O0/-O2 -> get=7      (Some path still correct)
RUE-1699  K=10 0.12s | K=16 0.13s | K=20 0.07s
          self-growing recursion -> E1200 depth-64, no hang
          sameness across spellings -> compiles, exit 7
          distinct by argument / depth / producer -> all rejected E0206
```

`scripts/rue fmt` clean. Full `./test.sh` result will be posted as a comment; this stays a draft until it is green.

Fixes: RUE-1758
Fixes: RUE-1699

---


steveklabnik:

```
=== TEST SUITE: PASSED ===
TESTSH_EXIT=0
```

Run in the foreground and unpiped, with the exit code captured to a file rather than read from a pipeline's status. 234 targets rather than the usual 231 — the two new regression corpora (`inlined_continuation_lowering_order.toml`, `anonymous_identity_nesting.toml`) add targets.

Agrees with CI on this head (`6c71d96b`), green across all 23 checks. Two of those are worth naming, because they speak directly to the riskiest claims here: **compiler reproducibility** and **rue_program digest sensitivity** both pass, which is independent evidence that the block-ordering change leaves correct programs' codegen unchanged and that removing the identity field disturbed nothing beyond the single intended digest shift.

---

PR #2630, opened 2026-08-23
RUE-1636, RUE-1654, RUE-1710: keep an integer's inference class at one width, and stop deriving a free size by scanning

steveklabnik:
Three bugs from one fix cycle, in disjoint crates. Every claim below was re-verified by execution on this branch rather than inherited from the worker notes.

**Rebased onto current trunk. This PR no longer carries RUE-1711** — see "What changed on rebase" at the bottom.

## RUE-1636 — the unifier lost type-variable equality

Integer match patterns were typed as the bare `InferType::IntLiteral` terminal. Binding the scrutinee stored that absorbing terminal into its substitution slot, so variables already chained to the scrutinee path-compressed onto it and the union-find class stopped having a single representative. A later concrete type then upgraded only the queried variable; its siblings defaulted to i32.

Patterns are now typed as marked int-literal *variables* (`fresh_int_literal_var`), the way integer literal expressions already were, so binding extends the chain instead of terminating it.

Three spellings, three different symptoms — all reproduced here before and after:

| variant | before | after |
| --- | --- | --- |
| `match c {3=>{},_=>{}}` then return i64 | `E0800` literal out of range for i32 | compiles |
| `let b = 0 - 5;` then `a + b + c` | `E9000` CFG verification failed | compiles |
| `let b = 2;` then `a` | silent `add %8:i64, %9:i32` | consistent AIR |

The third is the one that matters most: it reached codegen and worked only by accident of the 8-byte slot model.

## RUE-1654 — AIR validation did not check operand agreement

Which is why that third variant shipped silently. `Air::finish` now enforces that arithmetic, bitwise, shift and logical operands agree with each other and with the result; that comparisons agree and yield `bool`; and that a `PlaceWrite` matches its place. `!` and the error type are the only escapes, reusing what `Type::can_coerce_to` already encodes.

Implemented **before** the RUE-1636 fix, deliberately, so the guard could be shown to fire on the bad AIR and then go quiet once the inference bug was gone.

Integer/integer *comparison* operands are exempt for now. Two producers remain, both named in a comment in `finish` and both filed rather than papered over: RUE-1777 (the slice bounds check's `Lt(index, u64 len)`) and RUE-1778 (a literal compared against a struct field whose base is still an inference variable). Fix both and the exemption can be deleted. Arithmetic is enforced with no exemption at all.

Turning the guard on found a third producer, which **is** fixed here: indexing a base whose type is an interned array or the slice struct dropped an element type the base already carried, making `s[0].v * 4294967296` on a slice a bogus `E0800`. That was a user-visible reject-valid, not just an AIR smell.

## RUE-1710 — free_c_string freed with a scan-derived size

`owned_c_string` allocated `len + 1` bytes; `free_c_string` recovered the size by scanning to the first NUL. An interior NUL freed a 4 KiB block under size 3, and since the allocator classifies by the size it is handed, the block joined the 8-byte free list and came straight back out as an 8-byte allocation. Corruption, not a leak — and observable only by pointer identity, which is filed separately as RUE-1779.

`owned_c_string` now fails fast, with a public `has_interior_nul` predicate so a caller holding untrusted bytes keeps the decision. The scan in `free_c_string` is therefore exact by construction and its signature is unchanged. The doc comment that asserted the opposite is corrected.

## The cross-mechanism case, and why it is load-bearing

Added at integration time, and the rebase made the argument for it stronger rather than weaker.

Trunk's `_path_has_nul` — part of the RUE-1711 fix that landed while this was in flight — scans with `if b == 0`: the **bare literal** spelling, compared against a `u8` read out of an `Option(u8)` match. That is exactly the shape RUE-1636 governs, so a shipped security check depends on the inference fix in this PR.

The case pins that spelling directly and confirms it emits `%15 : u8 = const 0` and an honest `eq %14:u8, %15:u8`. It is asserted directly rather than left to the guard, because the RUE-1654 guard would **not** catch a regression there — it exempts integer/integer comparisons pending RUE-1777 and RUE-1778.

Verified in both directions: trunk's full fs suite runs green against these inference changes (34 passed, including the new NUL-rejection cases).

## What changed on rebase

RUE-1711 landed independently on trunk (#2625) while this was in flight. That fix rejects a NUL-bearing path *before* marshalling rather than inside `_path_to_cbuf`, which keeps the rejection free of both the allocation and the syscall, and lets `_mkdirat_raw` answer on its raw-errno channel.

It subsumes what was here, so this branch drops its own `std/fs.rue` change and its five `fs_interior_nul` cases rather than carrying a second spelling of the same guard. Trunk's coverage was checked for gaps first: its cases cover both `rename` operands and the no-parents-created guarantee, and the surrounding `fs_file_io` cases already serve as the NUL-free control. Nothing was lost worth re-adding.

## Verification

`./test.sh` → `Tests finished: Pass 234. Fail 0. Timeout 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0`, exit 0 on the pre-rebase tree; re-running on the rebased tree, with the affected sections already confirmed green individually (`integer_class_width` 15/15, `std_c_strings` 5/5, `fs_` 34/34).

Both backends: no codegen or CFG files changed, so there is no target split to ship; `--emit asm --target aarch64-linux` was sanity-checked on the repros regardless.

New coverage: 15 CLI cases in `integer_class_width`, 5 in `std_c_strings`, 5 spec cases under `4.7:13`/`4.1:3`/`4.7:23`, and a `ValidatedAir` unit test per new rule arm. The std.c cases were mutation-checked with `std/c.rue` stashed: 3/5 fail, the survivors being deliberate controls.

## Filed, not fixed here

RUE-1777 and RUE-1778 (the two remaining comparison-agreement producers — fixing both lets the RUE-1654 rule tighten to plain `operands_agree`), RUE-1779 (the allocator cannot detect a wrong-size-class free), RUE-1780 (no `std.fs` path type, so the NUL invariant is procedural rather than structural), RUE-1781 (`free_c_string(null)` faults), RUE-1782 (a signal-killed `test.sh` prints `TEST SUITE: PASSED`).

Fixes RUE-1636
Fixes RUE-1654
Fixes RUE-1710

### DIFF
--- a/crates/rue-air/src/inference/generate.rs
+++ b/crates/rue-air/src/inference/generate.rs
@@ -636,6 +636,37 @@ impl<'a> ConstraintGenerator<'a> {
         }
     }
 
+    /// The element type of an indexable operand whose type is *already*
+    /// concrete at constraint-generation time: an interned fixed array
+    /// `[T; N]`, or the synthetic slice struct `[T]` whose first field is the
+    /// `*const T` half of the fat pointer (ADR-0043, RUE-322). An
+    /// `InferType::Array` base is handled structurally by the caller; this
+    /// covers the interned forms — a slice parameter, an annotated binding, a
+    /// call result — that reach the generator as `Concrete`.
+    ///
+    /// Without it, indexing a concrete slice degraded to a fresh variable, so
+    /// an integer literal sharing an operator with the element (`e.i.a * 10`)
+    /// had nothing to take its width from and defaulted to i32, while sema
+    /// typed the operator from the element's real width. That is exactly the
+    /// mixed-width AIR the operator-agreement check in `inst.rs` rejects
+    /// (RUE-1636 family; the check found this producer).
+    fn concrete_element_type(&self, ty: &InferType) -> Option<Type> {
+        let ty = ty.as_concrete()?;
+        if let Some(array_id) = ty.as_array() {
+            return Some(self.type_pool.array_def(array_id).0);
+        }
+        let id = ty.as_struct()?;
+        // `str`/`Str(N)` share the view representation but index as bytes;
+        // `is_string_indexable_type` answers those before this is consulted.
+        if !crate::types::is_slice_struct_name(&self.type_pool.struct_def(id).name) {
+            return None;
+        }
+        match self.type_pool.struct_def(id).fields.first()?.ty.kind() {
+            TypeKind::PtrConst(ptr_id) => Some(self.type_pool.ptr_const_def(ptr_id)),
+            _ => None,
+        }
+    }
+
     /// Provide file-level constant types (name -> declared type) for `VarRef`
     /// resolution. See the `const_types` field for details (RUE-142).
     pub fn with_const_types(mut self, const_types: &'a AHashMap<(FileId, Spur), Type>) -> Self {
@@ -2905,12 +2936,19 @@ impl<'a> ConstraintGenerator<'a> {
                 } else {
                     match &base_info.ty {
                         InferType::Array { element, .. } => (**element).clone(),
-                        _ => {
-                            // Base might be a type variable that will resolve to an array.
-                            // Use a fresh variable for the element type.
-                            let result_var = self.fresh_var();
-                            InferType::Var(result_var)
-                        }
+                        // An interned array or slice type carries its element
+                        // type too; publish it rather than letting a literal
+                        // beside the element default to i32 (see
+                        // `concrete_element_type`).
+                        _ => match self.concrete_element_type(&base_info.ty) {
+                            Some(element_type) => self.type_to_infer(element_type),
+                            None => {
+                                // Base might be a type variable that will resolve to an array.
+                                // Use a fresh variable for the element type.
+                                let result_var = self.fresh_var();
+                                InferType::Var(result_var)
+                            }
+                        },
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


steveklabnik:
CI is red on `test (linux-x64-cli-shard-3)`, and the aggregate `CI success` gate fails downstream of it. I am **not** pushing a fix, and here is exactly why.

**The failure**

```
---- cli.watch::change_during_compile_publishes_only_newest_revision ----
timed out waiting for watch event "change-detected" (count 1)
watch process status: Some(ExitStatus(unix_wait_status(9)))
--- watch stdout ---
Watching main.rue for changes
Compiled main.rue -> prog in 296 ms (target: x86-64-linux, linker: internal)
Compiled main.rue -> prog in 312 ms (target: x86-64-linux, linker: internal)
```

Both compiles **succeeded**. The failure is in watch-event delivery, and `unix_wait_status(9)` is the harness SIGKILLing the watch process after its own timeout.

**Why it is not this diff**

This PR touches integer inference, AIR validation, `std/c.rue` and test cases. It touches nothing in watch orchestration, file observation, or the driver loop.

I checked the one causal link that would have made it mine. The RUE-1654 guard added here runs an operand-agreement pass over every AIR instruction, so in principle it could slow compilation and tip a timing-sensitive test. Measured on this tree: the same trivial program compiles in **29-31ms warm**, against the runner's ~300ms. The guard's cost is proportional to instruction count on a five-instruction program — two orders of magnitude below the 250ms window this test depends on.

The case is `kind = "cancel"` with `compile_delay_ms = 250`, firing two edits into that injected window. On this runner the compiles took 296ms and 312ms, both longer than the 250ms window meant to contain them, so the margin was spent before the edits landed.

**Verification on this tree**

- Full `./test.sh`: `Tests finished: Pass 234. Fail 0. Timeout 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0`, exit 0.
- `scripts/rue cli watch` run three times: 4/4 passing each time, ~1.0s per run — including the case that failed here.

Both of those ran while this host was also running two other build workers, so the passes are under load, not on an idle machine.

**Why I am not patching it here**

The fix belongs in the watch test harness, which is unrelated to this change; patching it in this PR would widen the diff into another subsystem. Filed as RUE-1783 with a proposed direction: make the injected delay block until the harness signals its edits have been written, rather than sleeping a fixed duration that has to out-race a real compile. The invariant the case depends on — the injected delay must exceed a real compile on the slowest supported runner — is currently implicit next to a constant calibrated on fast hardware.

I could not re-run the job to confirm: the re-run API returns 403 for me. So this needs either a maintainer re-run, or RUE-1783 landing first.

I am keeping this PR watched and will re-check until it is green, merged, or closed.

---

PR #2642, opened 2026-08-24
RUE-1700: type comptime self-call reduction

steveklabnik:
**Rescoped.** This PR originally carried RUE-1766, RUE-1761 and RUE-1700. While it was open, trunk absorbed the first two independently, so it has been rebuilt on fresh trunk as RUE-1700 alone. See the note at the bottom for what happened to the other two and how it was verified.

## The bug

A comptime-recursive function's value-returning self call opted out of the canonical durable reduction and fell back to the request-local evaluator, whose environment carries no resolved types. The callee's body therefore ran under the untyped fallback despite its parameter and return types being declared — and it failed in both directions.

**Spurious rejection.** With body `if n == 0 { ~0 } else { f(n - 1) }`:

```
f(0)  ->  255            (typed path)
f(1)  ->  E1200: value -1 is out of range for type u8
```

because the recursive step computed `~0` untyped as `-1`.

**Silent acceptance — the worse half.** With body `if n == 200 { (n + 100) - 100 } else { f(n + 1) }`:

```
f(200) ->  correctly rejected: the result 300 does not fit in u8
f(199) ->  compiled and ran, exit 200
```

`f(199)` evaluates the *identical* body through the self call, and the compile error spec 4.14:4 mandates was skipped with no diagnostic.

## The fix

The opt-out's recorded justification was that a self call would form a same-key query cycle. **It does not** — a comptime call is keyed by declaration *and arguments*, so `f(1)` reducing `f(0)` is a different key. That claim was the only thing holding the untyped path open.

Value-returning self calls now take the canonical typed query, which already shares `integer_semantics::IntegerType`, and the untyped fallback is **deleted rather than patched**, per RUE-1750's direction of collapsing evaluators rather than multiplying them. A reduction the durable evaluator cannot represent still falls back locally; `-> type` self calls keep their prior contract exactly.

## Verification

Verified by hand on this tree, rebuilt against current trunk:

- `f(199)` → rejected with E1200; `f(1)` → compiles to 255.
- Worker differential table: 48 four-way groups (`~ << >> + - *` × 8 widths) asserting direct == 1-hop == 3-hop == runtime, 0 mismatches. That program does not compile at all on base.
- Controls still rejected through the self call: 30/30 at depths 0/1/2, plus 18/18 "recovering shape" controls — **10 of which previously compiled and ran a would-panic operation**.
- A plain `a + b` control does *not* catch the silent direction; it is rejected incidentally via E0800. Only the recovering shape — overflow that comes back into range — exposes it. Worth knowing for anyone extending this coverage.
- Both runaway shapes still reach their depth diagnostic rather than hanging.
- Full `./test.sh` result posted below; this stays a draft until it is green.

## Note for RUE-1767

The depth cap is unchanged, but the counter a self call consumes moves, so reachable depth is one greater (failing at 50 rather than 49). A relaxation, recorded there rather than absorbed silently.

## What happened to RUE-1766 and RUE-1761

Both were fixed on trunk by other work while this PR was open, which is why they are gone from it rather than merged:

- **RUE-1766** — fixed by `299e8931`, which touched the same five files this PR's version did. Verified: the repro now gets `E0430`.
- **RUE-1761** — fixed by `a9876b011` ("RUE-1786: check the call-loan frame where the mutation happens, not where the argument is"). No commit names RUE-1761, but that generalization subsumes the receiver path. Verified by running the receiver repro on `6784806a` (use-after-free: prints `5` for a freed buffer, then reads through the stale pointer) and on `d0f9a8e66` (rejected). This PR's only delta from trunk is comptime-side, so that rejection is trunk's behavior, not this branch's.

The duplicated RUE-1766 work was dropped rather than merged as a second fix for the same bug.

Fixes: RUE-1700
Part-of: RUE-1750

### DIFF
--- a/crates/rue-air/src/sema/provider_body_host.rs
+++ b/crates/rue-air/src/sema/provider_body_host.rs
@@ -4506,15 +4506,30 @@ where
             definition
         };
         let signature = DurableCallableSource::function(&self.source, &definition)?;
-        // Runtime-valued self calls are request-local and use the ordinary
-        // evaluator. A `-> type` self call, however, must ask the canonical
-        // comptime query: a same-key recursion is a typed query cycle that the
-        // required reduction below turns into an E1200 at this call site.
-        if name == self.function_symbol
-            && !matches!(signature.result, crate::SemanticImportType::ComptimeType)
-        {
-            return None;
-        }
+        // A self call asks the canonical comptime query like any other call.
+        // The query is keyed by declaration *plus arguments*, so the recursive
+        // step of a terminating comptime recursion (`f(1)` reducing `f(0)`) is
+        // a different key, not a cycle; only a genuinely non-terminating
+        // recursion revisits a key, and the cycle abort inside
+        // `reduce_comptime_call` turns that into the depth diagnostic.
+        //
+        // A value-returning self call used to opt out here and fall back to the
+        // request-local evaluator, which reduces the callee body with no
+        // resolved types and so did its arithmetic under the untyped-i64
+        // fallback: `~0` at u8 came back as -1 and was rejected, and an
+        // overflowing `n + 100` at u8 silently produced a wrapped result
+        // instead of the compile error 4.14:4 requires (RUE-1700). The durable
+        // evaluator knows the declared operand types, so routing through it
+        // makes a value reached through a self call equal the same value
+        // computed directly.
+        let self_call = name == self.function_symbol;
+        let required_type_reduction =
+            matches!(signature.result, crate::SemanticImportType::ComptimeType);
+        // A `-> type` self call already asked the canonical query and had no
+        // local fallback; only the value-returning self call, which is the one
+        // that used to opt out entirely, gains one. Keeping the two apart is
+        // what makes this change a strict addition for type constructors.
+        let value_self_call = self_call && !required_type_reduction;
         let type_arguments = signature
             .parameters
             .iter()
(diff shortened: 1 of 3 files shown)


steveklabnik:
Full suite result, as the description promised — plus a note on CI.

```
=== TEST SUITE: PASSED ===
TESTSH_EXIT=0
```

Run in the foreground and unpiped on head `ee62b629`, with the exit code captured to a file rather than read from a pipeline's status.

## Three CI-driven fixes since the first push

The initial head failed `premerge`, which runs both differential corpora, with five problems — one spec disagreement, three CLI disagreements, and one unregistered model gap:

- `f05001a1` — the spec case became compile-only.
- `6bbc66c4` — the three CLI cases became compile-only, and the `ByteCopy` model gap the audit named by hand was registered.
- `ee62b629` — those four cases still declared `stdout`/`exit_code`, which RUE-1717 now rejects rather than silently discarding; removing the fields also fixed `zero-filter-harness-tests`, which was failing only because the spec and CLI harnesses aborted on the validation error before they could explain an unmatched filter.

All four disagreements share one root cause, worth reading on its own: the oracle models by-reference parameters as copy-in / copy-out, and its module doc calls that *"observably identical to by-reference under the law of exclusivity"*. That equivalence fails whenever the same root is aliased — copy-in snapshots at call time, so a nested mutation is invisible. On `a.scalar(a.bump())` the oracle computes 1 where the compiler computes 2. The core calculus sides with the compiler: its evaluation-context grammar admits only a by-value argument as a redex and states a by-reference argument is "a place, not reduced".

That is a pre-existing oracle defect, not a regression here — the shape already compiled on trunk, and no case had exercised it. Filed as RUE-1778; the four cases carry comments naming the stdout assertions to restore once the oracle models aliased by-reference arguments.

## CI note

GitHub created no workflow runs for the last three pushes (`f05001a1`, `6bbc66c4`, `ee62b629`) — 25 runs exist for this branch and the newest is on the original head. `workflow_dispatch` is not available to this integration (403). Rather than push an empty commit to force a `synchronize`, this relies on the merge queue's `merge_group` run, which validates the exact tree that lands and blocks the merge if it fails.

Locally verified on `ee62b629` beyond the full suite: both oracle-diff lanes (`LANES_EXIT=0`, zero disagreements, zero harness failures), `spec-traceability`, and `zero-filter-harness-tests`.

---

steveklabnik:
Correction to the comment above: the head is now `3741c90c`, not `ee62b629`.

The suite verdict quoted above (`TESTSH_EXIT=0`, Pass 234) was measured on `ee62b629`. The one commit since is text-only — the compile-only comments cited `RUE-1778`, an identifier I picked before filing; the oracle defect is actually **RUE-1789**. It changes a `description` string and three comment lines, no case source, no assertions, no compiler code.

The oracle issue is now filed as [RUE-1789](https://linear.app/steve-klabnik/issue/RUE-1789), with the four downgraded cases and their stdout assertions listed so they can be restored when it lands.

---

steveklabnik:

```
=== TEST SUITE: PASSED ===
TESTSH_EXIT=0
```

Run in the foreground and unpiped on head `36f5f274a`, exit code captured to a file rather than read from a pipeline's status.

CI agrees on the same head: 24 success, 2 skipped, including `CI success`, `premerge`, both oracle-diff corpora, all four CLI shards, asan, valgrind, compiler reproducibility, digest sensitivity, and both arm64 natives.

---

PR #2643, opened 2026-08-24
RUE-1788: subtract each corpus action together with its test target

steveklabnik:
Fixes RUE-1788. Completes what RUE-1511 (PR #2405) set out to do — that issue stays Done; this is the follow-up, not a reopening.

## The bug

RUE-1511 removed the two oracle-diff corpus actions from the premerge build scope, on this acceptance criterion:

> `premerge (linux-x64)` no longer builds either oracle-diff corpus action, on both the narrowed and unnarrowed paths.

It has been building them ever since. `cached_corpus_suite` emits each corpus as a **pair** (corpus.bzl): a `_corpus_action` genrule that runs the harness, and a thin `sh_test` carrying `args = ["$(location :NAME-action)"]`. So `NAME` depends on `NAME-action`, and subtracting only the action leaves the build reaching it through that edge:

```
$ ./buck2 uquery "deps(//crates/rue-oracle-diff:oracle-diff-test, 1)"
root//crates/rue-oracle-diff:oracle-diff-test-action     <-- pulled back in

$ scripts/affected-targets build-scope | grep oracle-diff   # before
//crates/rue-oracle-diff:oracle-diff-test                <-- still in scope
//crates/rue-oracle-diff:oracle-diff-spec-test           <-- still in scope
```

Building the action *runs the corpus*, which is the whole reason the subtraction exists.

## Why nothing caught it

Both the acceptance check and this script's tests asserted that no `-action` label appeared in the output. That stayed true throughout — the leak was a label the assertion did not name.

The stub made it unobservable regardless. Its `//crates/...` answer listed no corpus targets at all and ignored the `except set(...)` it was handed, so the unnarrowed case returned the same two targets no matter what the script subtracted.

## The fix

Emit both halves of each deferred pair. Both call sites keep their shape — the unnarrowed `except set(...)` and the narrowed `grep -Fxv` now subtract the same larger set — so the narrowed and unnarrowed spellings cannot diverge, which is the failure mode RUE-1511 called out.

The pair is the complete closure by construction: the macro is the only producer of a `_corpus_action` and gives each exactly one consumer. `cquery rdeps` confirms it on the current graph.

## Verification

End to end, against the dependency closure rather than the printed list — the latter is what the bug walked around:

```
$ ./buck2 cquery "kind('_corpus_action', deps(set(<203-target scope>)))"
(empty)

# add back the two test targets the old code left in:
$ ./buck2 cquery "kind('_corpus_action', deps(set(<scope> :oracle-diff-test :oracle-diff-spec-test)))"
root//crates/rue-oracle-diff:oracle-diff-spec-test-action
root//crates/rue-oracle-diff:oracle-diff-test-action
```

Gates: `//:affected-targets-tool-tests` 101 checks pass and **4 fail against the previous behavior**; `scripts/rue premerge` 200 passed, 0 failed; `validate-shell-bash-baseline.py` clean; `validate-ci-gate.py` valid.

Coverage is unchanged where it matters: `test (linux-x64-oracle-diff)` and `test (linux-x64-oracle-diff-spec)` both ran and passed on this PR, which also exercised the new determinator code, since the diff is that determinator.

### On the wall-time claim

Not established here, and worth saying plainly. `premerge (linux-x64)` took 103s on this PR against 299s on #2625's green run, but those two diffs invalidate very different amounts of cache — #2625 changed `std/fs.rue`, this one changes shell scripts — so the delta is directional, not a measurement. RUE-1511's ~240s came from a median over 169 cold-compiler runs, and confirming the recovery needs that same methodology rather than one comparison.

## Tests

The stub now models the real graph — each suite as a pair, and an `except set(...)` it actually honours — and the assertions are exact scopes rather than an absent suffix. A fix that removed only the `-action` label passes every other case in the file and fails the two new ones.

The fail-closed property is unchanged and still covered: an action **no** required lane owns keeps both halves, because the corpus must still execute somewhere.

## Docs

AGENTS.md called `scripts/rue premerge` the "canonical required-CI tier". It is a tier — it selects on `rue_test_tier_premerge`, and the oracle-diff corpora are `tier = "slow"` while their lanes gate the merge queue. That gap survives this fix, so the testing section now says what to run when a change touches corpus cases. This is the gap that cost RUE-1711 (#2625) a CI round.

## Known limit

The subtraction still names targets rather than deriving the closure from the graph: a hand-written dependency on a corpus action, outside the macro, would evade it. `rdeps` in `uquery` cannot express the closure here (it fails traversing unconfigured toolchain deps), and `cquery` is too costly for this path — so the construction argument above is what the rule rests on.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,7 @@ scripts/rue build                    # build the compiler and print its path
 scripts/rue exec prog.rue            # compile and run a quick program
 RUE="$(scripts/rue-bin)"; "$RUE" main.rue -o out
 scripts/rue quick                    # fast unit suite
-scripts/rue premerge                 # canonical required-CI tier
+scripts/rue premerge                 # canonical premerge test tier
 scripts/rue slow                     # canonical scheduled slow tier
 scripts/rue stress                   # canonical opt-in stress tier
 scripts/rue all                      # canonical union of all tiers
@@ -226,6 +226,17 @@ Use proportionate validation:
    architecture-specific suites.
 3. Run the broad/full suite once when the change is cross-cutting or high-risk.
    CI is the final full-platform authority.
+
+`scripts/rue premerge` is a tier, not a preview of required CI. It selects on
+`rue_test_tier_premerge`, so a required lane whose target sits in another tier
+is outside what it can report — the oracle-diff corpora are `tier = "slow"`
+while `test (linux-x64-oracle-diff*)` gates the merge queue. Adding a CLI or
+spec case is the common way to land in that gap: a case reaching a construct
+the oracle does not model is a harness failure until its gap is registered in
+`crates/rue-oracle-diff/src/model_gaps/`, and the audit prints the exact entry
+to add. Run `./buck2 test //crates/rue-oracle-diff:oracle-diff-test` when a
+change adds or edits corpus cases; it reproduces that failure in about two
+minutes (RUE-1711 cost a CI round for want of it).
 
 The generated-oracle smoke target has a fixed two-second child-process budget
 and can time out under heavy parallel load on the local macOS host. If failures
--- a/scripts/affected-targets
+++ b/scripts/affected-targets
@@ -45,7 +45,8 @@
 #   build-scope        Print the targets a `//crates/...` lane may build, with
 #                      or without an impacted list. Narrowing must stay inside
 #                      the lane's own scope, and neither spelling of that scope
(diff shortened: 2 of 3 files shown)


steveklabnik:
Superseded by #2645. This concurrent draft became visible after implementation began; #2645 derives the full exclusion from the configured Buck graph, covers arbitrary reverse dependents, and preserves fail-open behavior when the graph query is unavailable.

PR #2681, opened 2026-08-24
RUE-1762: describe the call-loan nesting rule the compiler implements

steveklabnik:
6.1:36 ended by promising that nested loans are conflict-free:

> …likewise a loan nested inside another argument (such as `f(inout x, g(borrow x))`) ends before the outer call begins and does not conflict.

That describes programs the compiler now rejects. A call's loan of a root spans the evaluation of its own argument list, so a nested exclusive loan can overlap it — which is the whole basis of the RUE-1696 / RUE-1766 / RUE-1786 use-after-free fixes.

## The rule was derived by running it, not from the ticket

The issue's stated fix direction predates two changes and is no longer accurate. It says to condition the exemption on a **shared** enclosing loan; RUE-1766 broadened that to a materialized view in **either** mode, and RUE-1786 moved the check to the mutation site. Writing the prose from the ticket would have encoded a rule the compiler stopped implementing two commits ago.

Measured on `a5dfbf16f`, nested `inout` of the same root:

| enclosing loan | result |
|---|---|
| `borrow str` — view-materialized, shared | reject `E0430` |
| `inout str` — view-materialized, exclusive | reject `E0430` |
| `borrow StrBuf` — address-passed, shared | reject `E0430` |
| `inout StrBuf` — address-passed, exclusive | **accept** |
| autoref `borrow self` receiver | reject `E0430` |
| autoref `inout self` receiver | **accept** |
| nested access inside a block-expression argument | reject `E0430` |

Controls: a nested *shared* loan never conflicts (`f(inout x, g(borrow x))` — 6.1:36's own example — still compiles), and a nested loan of a different root never conflicts.

So the condition is two-pronged: a nested exclusive loan conflicts when the enclosing loan is **shared**, or **view-materialized in either mode**. It does not conflict under an address-passed exclusive loan, where the nested write reaches the storage the enclosing callee reads — ordinary sequenced mutation, which the error-accumulator idiom depends on.

## What changed

**6.1:30** gains the nesting rule it had already de-facto acquired — nine cases cite it for exactly this — stated with why `inout str` conflicts despite being exclusive, and why the address-passed idiom stays legal.

**6.1:36**'s sentence is qualified rather than dropped: its own case (outer `inout`, nested `borrow`) is still correct, so it now scopes itself to nested *shared* loans and defers the exclusive case to 6.1:30.

No new rule ids, so traceability is unchanged.

## It closes a coverage hole, not just a prose gap

The amended prose asserts the rule holds through a by-reference receiver and inside a block expression. Neither spelling had a case — all nine existing ones pass the nested loan as a plain argument, and the receiver path was entirely unchecked until RUE-1786. Three cases now pin the receiver rejection, the address-passed receiver that stays legal, and the block-expression spelling.

## Verification

- `scripts/rue spec 6.1` — 153 passed, 0 failed, with all three new cases confirmed executed rather than silently filtered.
- `//:spec-traceability` — passes.
- `scripts/rue fmt` — clean.
- Full `./test.sh` result posted below once it completes; this PR stays a draft until then.

Fixes: RUE-1762

---


steveklabnik:

```
=== TEST SUITE: PASSED ===
TESTSH_EXIT=0
```

Run in the foreground and unpiped on head `4c3b1611a`, exit code captured to a file rather than read from a pipeline's status.

CI agrees on the same head: 24 success, 2 skipped, including `CI success`, `premerge`, and both `oracle-diff` lanes — confirming the compile-only change resolved the disagreement.

## Note on the one CI failure this PR saw

The first push failed `oracle-diff-spec` with a single disagreement on `nested_inout_under_address_passed_inout_self_receiver_ok` (expected exit 0, oracle got 1). That was predicted before the push and is [RUE-1789](https://linear.app/steve-klabnik/issue/RUE-1789), not a wrong case: the oracle models by-reference parameters as copy-in / copy-out, so it reads the receiver before the nested mutation and computes 0 where the compiler computes 1. The core calculus sides with the compiler — a by-reference argument is a place, not reduced.

That case is now `compile_only`, which still pins what 6.1:30 needs from it (the shape is *accepted*), with its `exit_code` removed since RUE-1717 rejects that pairing. A comment names the assertion to restore.

This brings the count of cases whose runtime assertions RUE-1789 suppresses to five — four from the RUE-1766 work plus this one. Every new case touching by-reference aliasing now arrives unable to check its own value, which is worth weighing when prioritising that issue.

---

PR #2825, opened 2026-08-30
RUE-1845: stop re-parsing the embedded runtime archive to validate it

steveklabnik:
Addresses RUE-1845 for the system-linker path. See **Scope**; referenced without a
closing keyword.

## Problem

The system-linker path writes the embedded runtime bytes to disk **verbatim** and
never needs the parsed archive — but it called the archive-returning validation
entry point, which decodes every member (headers, symbol tables, relocations, and
a `Vec` per section) and then dropped the result.

Only the *verdict* is memoized. So every link after the first decoded several
megabytes purely to re-derive a value already sitting in the `OnceLock`, and paid
it again on every warm rebuild in a retained `--watch` session, where the embedded
bytes cannot change within the process.

## Change

A validation-only entry point that consults the memo **before** parsing, keyed on
the same `std::ptr::eq` physical-identity test the memo itself uses — so a
caller-supplied archive that merely compares equal is still fully validated.

Both validation variants now take their per-target memo from one
`embedded_runtime_validation` helper instead of repeating the three-arm mapping
(it appeared twice).

## Scope

The **internal-link** path still needs the parsed archive to feed
`add_archive_with_cancellation`, so its parse remains.

Memoizing that one is what the issue's main proposal describes, and it is not a
memo: it needs `Archive.objects` held as `Arc` so `add_archive` can
take `&Archive` and clone only the selected members. Neither `Archive` nor
`ObjectFile` derives `Clone` today, and `add_archive_with_cancellation` consumes
its argument and moves members out — so it is a reshape of the linker's archive
API, in the code path every single link goes through. I did not want to take that
on in the same change as a memo lookup.

## Guard

A `RUNTIME_ARCHIVE_PARSES` counter incremented where the decode actually happens,
plus `validating_the_embedded_runtime_stops_reparsing_it`: after the verdict is
memoized, repeated validation must not decode the archive again. The test primes
the verdict first and asserts only the steady state, so it does not depend on what
else ran earlier in the process.

**I verified it bites** by bypassing the memo consult — the test failed with
`validating the embedded archive re-parsed it after the verdict was memoized` —
and passed again once restored.

## Validation

On the first commit (`4e8ec4430`):

- `scripts/rue unit rue-compiler` — 1133 passed, 0 failed.
- `scripts/rue cli abi` — 63 passed, 0 failed. Linking is exercised by the real
  binary against real files, which is the coverage that matters for this path.
- `rue-compiler-clippy`, `rue-compiler-fmt-check`,
  `rue-compiler-debug-assert-check` green.

On the current head (`c6e08a3f3`, after merging trunk):

- `scripts/rue premerge` — 203 pass, 0 fail, suite PASSED.
- `scripts/rue quick` — 33 pass, 0 fail.
- `buck2 test //crates/rue-compiler:{rue-compiler-clippy,rue-compiler-fmt-check,rue-compiler-debug-assert-check}`
  — 3 pass, 0 fail.

Validation behavior is unchanged: the same verdict is produced for the same bytes,
and a non-embedded archive still takes the full parse-and-validate path.

## Why the branch was merged with trunk

The first run's `test (linux-x64-cli-shard-0)` died on a remote action cache
`403 Forbidden` before any test body ran (`NO TESTS RAN`; the failing targets were
toolchain archives and std genrules this diff does not touch), and the cancelled
siblings followed from it. The branch was two commits behind trunk, so merging the
current base is a real update the PR needed anyway — and it re-runs CI from a fresh
head rather than replaying the poisoned one.

### DIFF
--- a/crates/rue-compiler/src/linking.rs
+++ b/crates/rue-compiler/src/linking.rs
@@ -444,6 +444,11 @@ static RUNTIME_AARCH64_LINUX_VALIDATION: std::sync::OnceLock<Result<(), String>>
     std::sync::OnceLock::new();
 static RUNTIME_AARCH64_MACOS_VALIDATION: std::sync::OnceLock<Result<(), String>> =
     std::sync::OnceLock::new();
+/// Times the embedded runtime archive has actually been decoded. Parsing it
+/// materializes every member — headers, symbol tables, relocations, and a
+/// `Vec<u8>` per section — so this count is the real cost being avoided
+/// (RUE-1845).
+static RUNTIME_ARCHIVE_PARSES: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
 
 /// Return the embedded rue-runtime archive matching `target`.
 pub(crate) fn runtime_for_target(target: Target) -> &'static [u8] {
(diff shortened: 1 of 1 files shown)


steveklabnik:
`test (linux-x64-cli-shard-0)` is red on `4e8ec44`, and it is not this PR's failure.

Every failed action in that job is a remote-cache error, not a test:

```
Internal error (stage: remote_action_cache): Remote Execution Error on REClientBuilder
Error: (Failed to query capabilities of remote: ... status: 403 Forbidden)
Action failed: toolchains//rust:rust-std-x86_64-unknown-linux-gnu (http_archive)
Action failed: toolchains//rust:rustc-linux-x86_64 (http_archive)
Action failed: root//:std-{arraybuf,rawbuf,sort,strbuf,strings}-source (genrule)
1 BUILDS FAILED
NO TESTS RAN
```

The job died fetching the Rust toolchain and building the std sources — before any test body ran — and the failing targets are toolchain archives and std genrules that this diff does not touch. This change is confined to `crates/rue-compiler/src/linking.rs`.

There is no fix to port: the remote action cache returned `403 Forbidden`, which is a CI credential/service condition rather than anything reachable from this PR's scope, so I cannot make it robust here either.

I do not have permission to re-run the job (`rerun-failed-jobs` returns 403 for me), so I am not spending the confirming re-run. Someone with Actions write access re-running `linux-x64-cli-shard-0` should be enough; if the remote cache is still refusing, this will reproduce identically on any PR touching anything.

Locally on this commit: `scripts/rue unit rue-compiler` 1133 passed, `scripts/rue cli abi` 63 passed, `scripts/rue quick` 33 pass, and the three `rue-compiler` gates green.

I will keep watching this PR until it is green, merged or closed.

---

PR #2841, opened 2026-08-31
docs: survey the optimizer passes against LLVM and GCC prior art

steveklabnik:
Adds `docs/notes/optimizer-prior-art-survey.md` (plus its notes-index row; `scripts/validate-doc-links.py` passes). Docs only — no code changes.

## Why

RUE-1865 observed that LICM materializes loop preheaders mid-sweep where both LLVM (`LoopSimplify`) and GCC (`loop_optimizer_init (LOOPS_NORMAL)`) establish them as a precondition. That looked like an instance of a broader pattern, so this surveys the whole `crates/rue-cfg/src/opt/` pipeline against both compilers on the axes that motivated the question: correctness, capability, speed, and allocations.

## What it says

The organizing observation is that LLVM and GCC separate canonicalization, analysis, and transformation with an explicit preservation contract between them, where each Rue pass owns all three privately. ADR-0054's recompute-per-pass rule is the deliberate form of that choice; the note argues the ordering consequence — canonicalization is what *lets* a transform preserve an analysis, so RUE-1865 should land before any analysis caching is considered, not after.

Thirteen findings, each citing the Rue source line and the corresponding LLVM and GCC mechanism. The highest-value ones:

- **LICM's memory gate is whole-body.** A `Load` is non-invariant if the loop body holds *any* observable-effect instruction. `Alloc` is in that set and `CfgBuilder` lowers every `let` to an `Alloc`, so one local variable in a loop disables memory-read hoisting for the whole loop. `slot_facts.rs` already owns the per-slot write/escape classification that would replace the veto, and is already shared by `constopt` and `forward` — LICM is the third consumer that needs it and the one not using it. (It does need a loop-scoped query the module does not yet expose.)
- **RUE-934's stated blocker looks wrong, but the replacement is bigger than "must-execute".** That issue defers guarded trapping-op hoisting until loop rotation exists. GCC does not require rotation — `movement_possibility` returns `MOVE_PRESERVE_EXECUTION` and `determine_max_movement` clamps to `ALWAYS_EXECUTED_IN(bb)`, computed over the dominator tree, and `ensure_preheader` already guarantees the structural precondition. But Rue's panics are defined rather than UB, so a hoist must preserve the whole observable prefix: not only *which* trap fires, but *what has already happened* when it fires — hoisting above a `Store` or `Drop` makes the trap fire with that effect not yet performed. Must-execute plus trap-and-effect ordering is the rule, and that pairing is the design work. Flagged as wanting an ADR-0054 amendment.
- **The optimizer's work counters never reach production.** `opt/mod.rs:428` discards `licm::run`'s ten-field `Stats`; the published `OptimizationStats` carries only unrolling and growth. RUE-1843's own acceptance counter cannot show up anywhere a person would look. This is why the rest of the note is unmeasured, and it is the cheapest item in it.

Also recorded: `ensure_preheader` still holds the transactional whole-CFG clone that RUE-1663/RUE-1842 removed from the five sibling passes; unrolling refuses innermost loops merely for having a parent; CSE is block-local where `EarlyCSE` is dominator-scoped; constant propagation is not conditional (no SCCP); post-unroll cleanup runs `constopt → simplify → dce` and skips `forward`/`peephole`/`cse`; and LICM has no cost model where GCC has `stmt_cost`/`LIM_EXPENSIVE`.

There is a closing section on divergences that are deliberate and should stay — the `may_trap`/`has_observable_side_effect` split is better factored than either production compiler's, and `peephole`'s refusal to strength-reduce `x * 2^k` follows correctly from Rue's defined-panic semantics.

## Cross-audit

A second independent audit of the same trunk revision reviewed the first draft and corrected four claims; all four checked out against the source and are folded in (second commit):

- Finding 1 called `unroll`'s restart loop "the same defect" as LICM's while also noting that unrolling clones blocks. Those contradict, and the second is the true one: canonical preheaders make LICM instruction-motion-only but buy unroll's restarts nothing. Unroll needs batching or incremental forest update, tracked separately — RUE-1865 should **not** absorb it.
- Finding 3's ordering clause covered trap order but not observable effects (above).
- Finding 5 recommended debug-gating `ensure_preheader`'s verification. `verify.rs:15` says the guard must fire in release per RUE-45, `rue-cli-tests` carries a release-mode job to catch exactly that gating, and finding 11 defends the same always-on posture. Corrected to: drop the clone, keep the check.
- Finding 10 claimed LLVM and GCC size analysis scratch to the working set. Their `DominatorTree`, `LoopInfo` and `MemorySSA` are function-sized too; the advantage is reuse, flat representation, and explicit invalidation.

Findings 12 and 13 came from that audit, as did the reusable def-use index folded under finding 10.

## Scope and limits

- **Nothing here is measured.** Every performance claim is a structural argument about asymptotics or allocation counts. Finding 12 is why, and is the first thing that should change.
- No issues were filed. The note ends with a suggested issue set ordered by value, including three items that are comments on existing issues (RUE-1865, RUE-934, RUE-917) rather than new ones.

## Testing

`scripts/validate-doc-links.py` → `doc links valid: notes index complete, no dead references`. No code changed, so no test tier was run.


steveklabnik:
Closing without merging — the survey's content has been migrated into Linear as tracked, self-contained work items, so the note in `docs/notes/` is redundant and this branch is being deleted.

Fifteen issues filed under the Rue team (RUE-1866 through RUE-1880), each carrying its own problem statement, Rue source citations, LLVM/GCC prior art, proposed fix, and validation criterion — plus scope-correcting comments on the three existing issues the survey touched (RUE-1865, RUE-934, RUE-917).

The ordering the note argued for is now encoded as Linear blocking relations rather than a prose queue:

- RUE-1866 (drop `ensure_preheader`'s whole-CFG clone) → blocks RUE-1865, because canonicalization multiplies the calls to it.
- RUE-1865 → blocks RUE-1879 (analysis-preservation contract), because canonical form is what makes preservation a structural fact rather than a promise.
- RUE-1867 (ADR-0054 amendment on must-execute plus trap/effect ordering) → blocks RUE-934.
- RUE-1868 (publish the optimizer's work counters) → blocks RUE-1878 (LICM profitability), which is a measurement question.
- RUE-917 (mem2reg) → blocks RUE-1880 (LICM sinking).

Everything else is unblocked and parallelizable.

---

PR #2898, opened 2026-09-03
RUE-1937: link through zig cc on Linux

steveklabnik:
Closes RUE-1937.

## What changed

- `toolchains/cxx_tools.bzl`: `cxx_tools_with_linker` is replaced by two rules. `zig_cxx_tools` builds a `CxxToolsInfo` (compiler_type `clang`, archiver_type `gnu`, linker_type `gnu`) whose tools are `cmd_script` wrappers around `toolchains//:zig`'s `ZigToolchainInfo.zig` RunInfo, so the whole Zig tree is a hidden input and Zig's caches land under `BUCK_SCRATCH_PATH` through the existing cache wrapper. `exec_cxx_tools` forwards a selectable `CxxToolsInfo` from the exec-configured provider target.
- `toolchains/BUCK`: `toolchains//:zig-cxx-tools` (target `x86_64-linux-gnu.2.17` / `aarch64-linux-gnu.2.17` selected on `prelude//cpu:*`, `compatible_with` Linux) and `:cxx-tools` selecting it on `prelude//os:linux` and the prelude's `path_clang_tools` on `prelude//os:macos`. The `root//constraints:remote-execution` → `cc` linker override is gone; the constraint stays (nothing else selects on it; `platforms/remote_cache.bzl` still inserts it and its comment is updated). Header comment updated.
- Linker is `zig cc`, not `zig c++`: `zig c++` compiles libc++ from source into the per-action cache on first use (measured: minutes plus 5 MB of warnings), and Rust links need no C++ runtime. `zig c++` remains the `cxx_compiler`.
- The prelude's `-fuse-ld=lld` (added for every Linux non-g++ link in `_cxx_toolchain_from_cxx_tools_info`) is accepted by `zig cc` as an unused argument; buck2 surfaces no warning for it.
- `toolchains/zig/runtime-debug-discard.ld` + `runtime_debug_discard_test.sh`: see "Reproducibility" below.
- `docs/process/build-cache.md`: status note and item 6 under "What it took" describe the new state.

## glibc floor

`2.17`, as in `third-party/reindeer_rules.bzl`. It linked first time; the highest symbol version the compiler references is `GLIBC_2.17` (debug and release). No raise was needed.

## Reproducibility (the one non-obvious part)

The first run of `scripts/check-reproducible-compiler.sh` failed: the two compilers differed by 752 bytes in `.debug_info`/`.debug_line`/`.debug_str`. Zig compiles the glibc start files, libunwind (what `-lgcc_s` resolves to), and compiler_rt from source during the link, with `-gdwarf-4`, and clang records the action's cwd (the checkout root) as `DW_AT_comp_dir` of those 12 CUs. Zig offers no knob for only its runtime objects: `-g0`, `-fdebug-prefix-map`, `-ffile-prefix-map`, and `PWD` do not reach those compiles (verified), and every strip spelling (`-Wl,--strip-debug`, `-Wl,-S`, `-s`) sets the root module's `strip`, which turns into LLD `-s` and drops the symbol table too (Zig source: `main.zig` `-Wl` handling, `Compilation/Config.zig` `debug_format`, `link/Elf.zig`). `zig objcopy --strip-debug` is unimplemented for executables in 0.16.

The fix is an LLD linker script the linker wrapper passes via `-Wl,-T,…`: `/DISCARD/` of `.debug_*` from inputs matching `*zig-global-cache/*` and `*zig-local-cache/*` (the only inputs under those directories are the objects Zig built during the link), with `INSERT AFTER .text;` so LLD's default layout is untouched. Verified on a probe: output byte-identical across two working directories, `.symtab` kept, the caller's own DWARF kept, program headers and section order identical to a plain link (the non-INSERT form changes the layout and the binary does not run). The debug compiler now has 28 CUs, all rustc's prebuilt sysroot (`/rustc/...`); the prelude builds Rue's own crates without DWARF, so nothing first-party was ever involved. A policy test ties the script's patterns to the cache directory names in `defs.bzl`.

## ldd before / after (Linux x86_64 debug `//crates/rue:rue`)

Before (origin/trunk, host clang 18):
```
linux-vdso.so.1
libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
/lib64/ld-linux-x86-64.so.2
```
After (zig cc, `x86_64-linux-gnu.2.17`):
```
linux-vdso.so.1
libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
/lib64/ld-linux-x86-64.so.2
libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0
libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2
```
`libgcc_s.so.1` is gone (Zig's libunwind is linked statically); `libpthread.so.0` and `libdl.so.2` appear as separate `DT_NEEDED` entries because glibc 2.17 still shipped them as separate libraries (glibc ≥ 2.34 provides them as stubs, so this resolves everywhere). The release binary shows the same set.

## Verification (Linux x86_64, 4 cores)

| Command | Result |
| --- | --- |
| `./buck2 build //crates/rue:rue` | BUILD SUCCEEDED; `rue --version` → `rue 0.1.0`; `scripts/rue exec examples/first/stats.rue` runs |
| `./buck2 cquery "deps(//crates/rue:rue)" \| grep -i -E 'clang\|zig'` | `toolchains//:zig-cxx-tools`, `toolchains//:zig`, `toolchains//zig:hermetic-linux-x86_64`, `toolchains//zig:dist-linux-x86_64`, `toolchains//zig:runtime-debug-discard.ld`; no `path_clang_tools` (the only `clang` hit is the prelude's internal `prelude//cxx/tools:clang_tidy_wrapper` python tool) |
| `ldd "$(scripts/rue-bin)"` | above |
| `scripts/rue quick` | Pass 35, Fail 0 (1165 unit tests) |
| `./buck2 test //fixtures/rue-program: //:large-example-caldera-canary //:reproducible-programs` | Pass 6, Fail 0 |
| `scripts/check-reproducible-compiler.sh` | `Rue compiler rebuild is byte-reproducible`, sha256 `3cd0d42aed797f0b84d3e0c580d8ae152d4e7e58e45962b36e5ee5fe88b8626b` |
| `./buck2 test toolchains//...` | Pass 24, Fail 0 (includes the new `runtime-debug-discard-test`) |
| `./buck2 build //crates/rue:rue --target-platforms //platforms:release` | BUILD SUCCEEDED (ThinLTO link through zig cc); binary runs, `GLIBC_2.17` max |
| `python3 scripts/validate-shell-bash-baseline.py` | valid (new test script is Bash 3.2-clean) |
| `strings -a rue \| grep <checkout>` | 0 hits |

## Cost worth knowing

A fresh Zig cache costs about 17 s of CPU per link action on this machine (compiler_rt, libunwind, and the glibc start files are compiled from source into the per-action scratch; a warm cache links in 25 ms). With one cache per action that cost recurs on every Rust link (binaries, tests, build scripts). Measured here: `scripts/rue quick` 300–407 s on 4 cores (no trunk baseline was timed), a full cold compiler build under 3 min, the release build 336–397 s. If that matters, a follow-up could pre-populate the Zig cache from a Buck action.

## Not verified here

- macOS: unchanged in code (`path_clang_tools` + `-ld_classic`), but not built here.
- Linux aarch64: the `aarch64-linux-gnu.2.17` arm is selected on `prelude//cpu:arm64` and mirrors the mimalloc archive; not built here (CI's `ubuntu-24.04-arm` lanes will).
- Remote execution: the worker now links with the same Zig tools; the merge-queue canary is the check.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BmNTj9stFgZSNZihSxoTDS)_


steveklabnik:
CI status on 18ee4f0: every lane passed except `test (linux-x64-cli-shard-2)`, which failed before any test ran while materializing `rustc` from the BuildBuddy action cache:

```
Internal error (stage: materialize_inputs_failed): Error materializing artifact at path `buck-out/v2/art/toolchains/.../rustc/bin/rustc`
Error: (Failed to make BatchReadBlobs request: code: 'Internal error', message: "Unexpected EOF decoding stream.")
```

That is a cache blob read of the pinned Rust toolchain, an input this diff does not touch; the other three shards were cancelled by the matrix fail-fast. The same shards passed on the neighbouring PR minutes earlier. This integration cannot re-run jobs (403), so the shards get their fresh run when the branch is updated onto trunk after #2896 merges.

---

PR #2915, opened 2026-09-04
Finish and stabilize floating point (ADR-0065 phases 7–10)

steveklabnik:
## Summary

Completes ADR-0065 (f32/f64, IEEE-754) after RUE-1071/RUE-1072 and RUE-1073, and stabilizes the feature: the `floats` preview gate is removed. One linear commit rebased onto current trunk.

### Backends (phases 5/6 leftovers)
- Unsigned int→float and float→unsigned conversions on both targets (x86-64 halve-with-sticky-bit for u64→float and the 2^63 offset path for float→u64; AArch64 `ucvtf`/`fcvtzu`).
- Precise `@float_to_int` guards: succeeds iff `MIN-1 < x < MAX+1`, otherwise the integer-overflow trap. Covers NaN and both infinities.
- `@total_cmp` (IEEE totalOrder via a sign-folded bit key), `@sqrt`, `@floor`, `@ceil`, `@trunc`, `@round` (ties away).
- Decision: x86-64 directed rounding uses SSE4.1 `roundsd`/`roundss`, the one instruction above the SSE2 baseline the backend emits. Ties-away `@round` is a truncate-and-adjust sequence.
- Aggregate slots are classified by their leaf type through one shared mapper, with an FP return bank beside the GP one and typed slot traffic. This fixes an ICE for a let-bound struct with a float field crossing a call (both targets) and a silent miscompile writing a multi-slot float aggregate through an `inout` parameter. Sixteen CLI cases cover the shapes.

### Runtime formatting (RUE-1074 on top of RUE-1073)
- `@dbg` and `@to_string` accept floats through the width-explicit `__rue_to_string_float` helper from #2917, plus one matching `__rue_dbg_float(bits, width)` that formats through the same authority. Output is the shortest round-trip decimal via zmij: positional for decimal exponents in -5..=15 (f64) / -6..=12 (f32), otherwise `d.ddde±XX`; `NaN`, `inf`, `-inf`, `-0.0`.
- Sema emits an internal `BitCast` (validated float↔int at equal width) plus the width discriminator.

### Comptime floats
- The shared comptime engine evaluates `+ - * /`, negation, and comparisons at the operation's width (annotated type, else f64) with a canonical positive NaN. Mixed f32/f64 operands and float `%` are rejected. Amends ADR-0065's arbitrary-precision framing so comptime and runtime agree.
- An out-of-range source literal is rejected on the const path as it already was on the let path.
- `comptime_float` is not nameable, mirroring `comptime_int` (spec 3.12:3): a named `comptime_float` const previously read as 0.0 at runtime and a `comptime_float` parameter ICE'd; every annotation position is now E0204 and the E0475 help suggests `f64`. A captured comptime float takes its width from use, like a captured integer.

### Oracle
- Models float arithmetic, IEEE comparison (including aggregate equality), conversions with traps, total order, rounding, and formatting; the float model gaps are retired.

### Inference
- `f32`/`f64` are identifiers, not keywords, so a float type passed as a `comptime T: type` argument reached inference as a variable reference and a `-> T` return stayed the `type` placeholder; module-qualified generic calls never substituted the return type at all and ICE'd against a float literal. Both call shapes now reduce the declared return type through one helper.

### Constant folding
- Float constants were folded as raw bits and the integer identities x−x=0 and x·0=0 were applied to floats, misfolding `NaN == NaN`, `-0.0 == 0.0`, `inf - inf`, `inf * 0.0` at -O1 and above. Gated on the operand type.

### Spec (RUE-1075) and stabilization (RUE-1076)
- New chapter 3.12, threaded through tokens, literals, arithmetic and comparison operators, the intrinsic inventory, comptime, runtime behaviour, and the grammar appendix; every paragraph cited by spec cases. ADR-0065 marked implemented with an amendment recording the implementation decisions.
- `PreviewFeature::Floats` and E1109 retired per the ADR-0005 procedure; `%` on floats keeps a permanent diagnostic pointing at `std.math.rem`.

### std and examples
- `std.math`: `sqrt floor ceil trunc round fabs inf nan is_nan is_inf is_finite rem total_cmp` (generic over the float type; `rem` is an exact fmod). `std.fmt.float_to_string`. `examples/floats.rue`, and the `examples/ruelex` dogfood lexer learned float literals so the frontend differential stays green.

### Observable, target-defined behaviour
- The sign bit of a NaN produced by an operation differs between x86-64 (negative) and AArch64 (positive). It is visible only through `@total_cmp` or bit inspection.

## Test plan
- [x] `scripts/rue premerge` green on the pre-rebase tree; the only CLI failures are environmental to the container (no IPv6, uid 0). Re-run on the rebased head in progress.
- [x] Unit suites: air, codegen, compiler, cfg, runtime, runtime-abi, builtins, oracle, oracle-diff, error, parser
- [x] Spec suite, traceability, machine index, ADR registry, doc links, oracle-diff corpus, frontend differential, runtime archive validators
- [ ] CI on this PR: the pull_request event has not produced a workflow run for this branch and manual dispatch is refused for this integration; needs a nudge from the repo side.

https://claude.ai/code/session_01FDypeYDnXe1JKoS27PLgQC


steveklabnik:
CI on 1727ef6: `test (linux-x64-spec)` failed before any test ran. Buck could not materialize the Zig toolchain from BuildBuddy (`BatchReadBlobs ... missing grpc-status trailer, stream was terminated without a final status`), and the workflow's built-in retry from RUE-1949 hit the same truncation, so the spec runner never linked and the other test jobs were cancelled by fail-fast. Nothing in this diff touches the toolchain or the cache path. The same suite (`//:spec-tests`, plus traceability, the oracle-diff corpus, and premerge) passes locally on this head. I don't have permission to re-run the failed jobs from here, so this needs a re-run from the repo side; I'll keep watching the PR.

---

PR #2941, opened 2026-09-04
RUE-2003: keep hermetic toolchain distributions out of the remote CAS

DorianListens:
Fixes RUE-2003.

CI jobs have been dying in their first minute since 2026-09-03 16:34 UTC with `materialize_inputs_failed` and a truncated BuildBuddy `BatchReadBlobs` response, and the merge queue has been ejecting approved PRs. In every failing job the artifact is one of exactly four action digests, and all four are hermetic toolchain distributions: the Zig x86_64 and aarch64 trees, `rustc`, and `rust-std`. They are the only artifacts we make of tens of thousands of tiny files (Zig alone is 19,546 files and 341 MiB, all but two under the batch threshold), so materializing one is the only time we ask BuildBuddy for a batch of thousands of blobs. The failures land seconds into a build after a few hundred KiB, per-lane traffic is unchanged across the onset, RUE-1934 generates the client config byte for byte identically, and the same warm-from-CAS operation succeeded on 2026-08-31 at 483 MiB down. This is the service's handling of one large response, not our volume and not our config. RUE-1937 only exposed it by re-keying every Linux link so cached links became local misses that need a toolchain on disk; the intermittency is whether a lane gets an action-cache hit for the tree (materialize from CAS, can be truncated) or a miss (extract the tarball locally, cannot).

`toolchain_distribution` replaces `http_archive` for all twelve distributions: the archive comes from its origin URL through `download_file`, and the extraction sets `allow_cache_upload = False`, so the trees never enter the CAS and the failing operation cannot occur. The provisioned config also bounds one `BatchReadBlobs` response, since buck2's own CAS retry ignores the `Unknown`/`Internal` codes this arrives as. RUE-1949's replay is generalized from the Zig path to any artifact (3 of the 23 surveyed failures never qualified for it) with its measured 5-of-20 rescue rate recorded rather than implied, kept because a merge-group run has no operator. `scripts/ci-timed` reports buck2's `Network` counters per step so this is measurable next time.

The cost is measured and deliberate: a lane that already needed a toolchain is unchanged (clippy 273 to 296 MiB); a lane that did not now pays about 237 MiB and 10 to 20 s, about 134 s and 2.8 GiB across a run, because an action's output digest keys its consumers and buck2 can only learn it by running the extraction. Dropping the one `allow_cache_upload = False` line restores cache-served trees if BuildBuddy confirms a fix on their side.

Evidence: two independent CI runs on this branch, 33889388662 and 33890275789, both green with zero `materialize_inputs_failed`, and per-lane `Network: Down` identical across the two runs (a deterministic fetch of the pinned tarballs). The remote-execution canary only runs in the merge group and is exercised for the first time there; it already passed `--no-remote-cache`, which executes the same extraction on the worker. Local: the twelve converted extractions invalidated nothing downstream, and the build-sharing and wrapper-script suites (230 checks) pass, including a new rustc-tree retry fixture. Follow-ups filed: RUE-2009 (silent no-cache daemons).

### DIFF
--- a/.buckconfig.local.example
+++ b/.buckconfig.local.example
@@ -33,12 +33,18 @@
 #     when remote is enabled — even for cache-only use)
 #   - default_allow_cache_upload=true (the upload-enabled executor is one gate;
 #     ordinary Rust actions defer to this separate OSS-default-off gate)
+#
+# max_total_batch_size is not one of those four: it affects no hit rate, only
+# how much one BatchReadBlobs response may carry. See "Toolchain distributions
+# do not travel through the CAS" in docs/process/build-cache.md (RUE-2003).
 
 [buck2_re_client]
 engine_address = grpc://remote.buildbuddy.io
 action_cache_address = grpc://remote.buildbuddy.io
 cas_address = grpc://remote.buildbuddy.io
 tls = true
+max_total_batch_size = 1000000
+max_decoding_message_size = 16000000
 http_headers = x-buildbuddy-api-key:<YOUR_BUILDBUDDY_API_KEY>
 
 [buck2]
--- a/buck2
+++ b/buck2
@@ -100,31 +100,44 @@ if [[ "$uses_remote_cache" -eq 1 ]] && [[ "$execution_command" -eq 1 ]] && [[ "$
     [[ "$inserted" -eq 1 ]] || args+=(--prefer-local)
 fi
 
-# BuildBuddy occasionally terminates a large BatchReadBlobs response while
-# Buck is materializing the Zig executable. The failure happens before the
-# requested action runs, and immediately replaying the same command has proved
-# sufficient. Keep the exception deliberately narrower than a general remote
-# cache retry: every observed invariant must be present in the first attempt's
-# stderr, and the replay calls DotSlash directly so there can only be one retry.
-is_truncated_zig_materialization() {
-    local line previous_is_zig_failure=0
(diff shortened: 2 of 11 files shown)


DorianListens:
Post-merge evidence, for the record.

Merge-group run 33892616080 succeeded, so `remote execution (linux-x64)`, the one job that only runs in the queue, has now executed the converted extraction on the worker.

Cache probe 33889466642 (cold build, `buck2 clean`, warm rebuild) against the 2026-08-31 baseline 33386522916:

| | baseline | after |
|---|---|---|
| cold cache hits / commands | 28%, 1033 (746 local) | 27%, 1035 (752 local) |
| cold network | up 537 MiB, down 205 MiB | up 450 MiB, down 246 MiB |
| warm cache hits / commands | 97%, 1033 (36 local) | 96%, 1035 (40 local) |
| warm network | up 494 KiB, down 483 MiB | up 451 KiB, down 507 MiB |
| `materialize_inputs_failed` | (was failing) | 0 |

Cold uploads fell by 87 MiB, which is the extraction results no longer entering the CAS; the four extra warm local actions are the always-executing extractions. Warm download did not fall: the toolchain bytes moved from the CAS to the origin CDNs, and the probe's warm phase materializes the whole release graph from the CAS regardless, so treat that figure as flat. What changed for the bug is the shape of the request, not the byte total: no lane issues a batch read for a 20,000-blob tree any more.

Four green runs on the change with zero truncation: 33889388662, 33890275789, 33891853019 (pull_request), 33889466642 (probe). Open on the BuildBuddy side: their advertised `max_batch_total_size_bytes` and the frontend's response-size cap; if they name a limit, set `max_total_batch_size` to it and consider restoring cache-served trees.

PR #2961, opened 2026-09-05
Rewrite the tutorial around the current language, with every snippet run

### LINKED ISSUE #2387: Compilation error when following official tutorial
```
const std = @import("std");

fn add(a: i32, b: i32) -> i32 {
    a + b
}

fn is_positive(n: i32) -> bool {
    n > 0
}

fn main() -> i32 {
    let sum = add(3, 4);
    println("sum = " + @to_string(sum));

    if is_positive(sum) {
        println("sum is positive");
    }
    if is_positive(-5) {
        println("-5 is positive");
    } else {
        println("-5 is not positive");
    }

    sum
}
```

error: [E0204]: unknown type 'StrBuf'
  --> hello.rue:13:24
   |
13 |     println("sum = " + @to_string(sum));
   |                        ^^^^^^^^^^^^^^^
   |

I followed the official docs step by step, but the build failed. Here is the error log.

steveklabnik:
Fixes #2387.

## What was wrong

The reported `unknown type 'StrBuf'` had two causes in the tutorial itself:

- Every complete program in the structs and borrow/inout chapters used `@to_string` without importing `std`. Those fences were unmarked, so the snippet checker never compiled them: 15 of the 18 unmarked programs failed, 13 with exactly the reported error.
- Chapter 2 taught calling the compiler binary directly without `RUE_STD_PATH`, which `scripts/rue exec` sets and a direct invocation does not. The variable was mentioned once, in the last paragraph of the last chapter.

Beyond the bug, the tutorial had frozen in July while the language gained `for` loops, methods, payload enums, slices, `Result` with `?`, compound assignment, floats, `test` declarations, and most of the standard library.

## Commits

Each commit stands on its own:

1. **Minimal fix for the reported bug**: add the imports, mark those fences `check`, export `RUE_STD_PATH` beside `RUE` in every manual-compile block.
2. **Checker upgrade** (`scripts/check-tutorial-snippets.py`): a `run` action that executes the program and compares stdout with the following ```` ```text ```` fence (with `stdin=` and `exit=` attributes), `file=` fences for multi-file chapters, and unmarked `rue` fences become an error. Covered by new cases in `scripts/test-tutorial-snippets.py`.
3. **The rewrite**: sixteen chapters, starting with a "Why Rue" chapter on the design principles and ending with a reverse-Polish calculator built in one file, then split into a module with tests. Every fence is verified: 69 programs compiled and run with output compared, 3 test-bearing programs compiled, 14 compile-fail programs checked for their diagnostic code.
4. **Process doc**: `docs/process/tutorial.md` gets the new outline, the fence markers, and a list of compiler limitations the chapters work around. Also fixes a stale comment in `examples/welcome.rue`.

## Things to review

- The "Why Rue" chapter states principles from the ADRs and the Sept 3 design-principle issues (RUE-1884 to RUE-1914). Its "Where things stand" section says some are guarantees today and others are the direction under discussion; please check that framing.
- Chapter URLs change (for example `/tutorial/01-installation/` becomes `/tutorial/02-getting-started/`). Nothing in the repository links to the old URLs.
- The site was not built locally (zola is not provisioned in the sandbox); CI should cover the Zola links.

## Compiler issues found along the way

Writing verified examples surfaced four compiler problems, worked around in the text and recorded in the process doc:

- ICE (E9000, "enum payload ... was already dropped on a reaching path") on `?` inside a function returning `Result(T, StrBuf)`.
- `for` over a `borrow StrBuf` parameter is rejected with E0429, contrary to spec 4.8:26.
- A qualified enum path inside a payload pattern (`R.Err(E.A(b))`) does not parse.
- `inout` from an immutable `let` binding is accepted for free functions but rejected for `inout self` methods.

## Validation

- `scripts/check-tutorial-snippets.py`: 86 snippets, all pass.
- `python3 scripts/test-tutorial-snippets.py`: 16 tests pass.
- `./buck2 test //:tutorial-snippet-tests //:tutorial-snippet-tool-tests`: pass.

https://claude.ai/code/session_01EAGCE4A6jUzqkjHTo6ocLq

---
_Generated by [Claude Code](https://claude.ai/code/session_01EAGCE4A6jUzqkjHTo6ocLq)_


website/content/tutorial/01-why-rue.md
```diff
@@ -0,0 +1,158 @@
...
+
+You already know that `sort` changes `scores` and `total` only reads it. The
+call site says so, and the compiler makes sure it is telling the truth. A
+function that wants to mutate its argument must declare the parameter `inout`,
+and every caller must write `inout` too. Compare a language where mutation is
+invisible at the call:
+
+```python
+values.sort()      # does this change values? you have to know
+```
```
steveklabnik: This is the strongest paragraph in the chapter, and it is also the one place where the tutorial says something the language does not do. "Every caller must write `inout` too" is true for free functions and false for methods: spec 6.4:25 says the receiver mode is *not* written at the call site, so `scores.sort()` in Rue reads exactly like the Python line below it. Chapter 6 (line 171) and chapter 8 (line 349) both admit this later, and chapter 15 line 244 already has the honest phrasing ("`inout` in its declaration and, for free functions, at its call site").

Two options, and I filed RUE-2067 to decide between them at the language level:

1. Keep the claim and change the language: a prefix mark on the receiver place (`inout scores.sort();`), Hylo-style. Chaining is not the obstacle it looks like, because 6.4:27 already forbids an `inout self` or `borrow self` receiver that is not a place, so a mark would only ever appear once at the head of an expression.
2. Keep autoref and soften this paragraph: functions carry the mode at both ends; methods carry it in the signature, and the compiler requires a `let mut` receiver. In that case the Python example has to go, because it is literally Rue's own method syntax.

Either way, this paragraph and chapters 6, 8, and 15 should say the same thing.

---

website/content/tutorial/01-why-rue.md
```diff
@@ -0,0 +1,158 @@
...
+
+The same principle shows up all over the language:
+
+- **Failure is visible.** A function that can fail returns an `Option` or a
+  `Result`. The `?` operator marks every place a function might return early.
+  There are no exceptions, so a call can never jump out of your function
+  without a mark on the line.
+- **Nothing converts silently.** An `i32` never becomes an `i64` on its own; you
+  write `@intCast`. An integer never becomes a string; you write `@to_string`.
+  Every conversion is a named operation you can see.
```
steveklabnik: "Nothing converts silently" is true for numbers but the tutorial itself relies on one silent conversion: a string literal becomes a `StrBuf` wherever one is expected (spec 3.7:44). Chapter 10 line 54 does it with `let owned: StrBuf = "heap text"`, and chapter 11 line 56 passes `"42"` straight to a `StrBuf` parameter. `println` accepting both `str` and `StrBuf` will also read as overloading to someone who just read the next bullet.

Suggest narrowing the claim to what holds: no implicit numeric conversions, and name the one contextual rule for string literals so the reader is not surprised in chapter 10.

---

website/content/tutorial/01-why-rue.md
```diff
@@ -0,0 +1,158 @@
...
+```python
+values.sort()      # does this change values? you have to know
+```
+
+The same principle shows up all over the language:
+
+- **Failure is visible.** A function that can fail returns an `Option` or a
+  `Result`. The `?` operator marks every place a function might return early.
+  There are no exceptions, so a call can never jump out of your function
+  without a mark on the line.
```
steveklabnik: Nit: "a call can never jump out of your function without a mark on the line" is not quite right, because a trap does exactly that (an overflow inside a callee ends the program mid-line, unmarked). What is true is that a call can never *return to your caller* without a mark. Chapter 11 draws this distinction well; this sentence should match it.

---

website/content/tutorial/01-why-rue.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Where things stand
+
+Rue is a young language, and the honest status is that these principles are
+ahead of the implementation in places. Some of the points above are stated
+guarantees today; others are the direction the design is being checked against,
+recorded in the project's
+[design records](https://github.com/rue-language/rue/tree/trunk/docs/designs)
+and issue tracker. The tutorial says which is which whenever it matters, and the
+[last chapter](@/tutorial/16-whats-next.md) lists what is missing.
```
steveklabnik, line 172: On the framing question from the PR description: this section promises that "the tutorial says which is which whenever it matters," but nothing above is actually marked. Having read the whole tutorial against the spec, the items in this chapter that are not fully true today are:

- the call-site mode claim (methods, see my comment on line 35);
- "nothing converts silently" (string-literal contextual typing, see line 51);
- "Programs are deterministic" (line 140) is one I could not verify. `std.rand` and the map types exist; if their iteration order and seeding are guaranteed today, say so, otherwise this belongs in the "direction" bucket.

Everything else in the list (no prelude, no overloading, order independence, traps, `?`) I checked and it holds. I would either add a one-word marker to the two or three aspirational items or list them here explicitly, so "says which is which" is literally true.

---

website/content/tutorial/06-structs-and-methods.md
```diff
@@ -0,0 +1,255 @@
...
+    println(@to_string(p.manhattan()) + " " + @to_string(o.manhattan()));
+    0
+}
+```
+
+```text
+7 0
+```
+
+Method calls chain the way you would expect: `Point.new(1, 2).manhattan()`.
```
steveklabnik: This example does not compile. `manhattan` takes `borrow self`, and spec 6.4:27 requires a by-reference receiver to be a place, which a call result is not. I checked with the trunk compiler:

```text
error: [E0427]: borrow argument must be a variable, field, or array element
  --> chain.rue:13:24
   |
13 |     println(@to_string(Point.new(1, 2).manhattan()));
   |                        ^^^^^^^^^^^^^^^
```

It slipped through because it is prose, not a fence. Either bind first (`let p = Point.new(1, 2); p.manhattan()`) or chain a by-value method that returns `Self`, which is the kind of chain Rue actually supports today. Worth a sentence either way, since "chain the way you would expect" is exactly the expectation the language does not meet for `borrow self` and `inout self` methods.

---

website/content/tutorial/06-structs-and-methods.md
```diff
@@ -0,0 +1,255 @@
...
+
+```text
+12
+48
+```
+
+Notice that the call site `r.scale(2)` does not spell out `inout`. The mode is
+part of the method's declaration, and calling a mutating method on a binding
+that is not `let mut` is an error. Free functions, by contrast, require the
+caller to write the mode at the call; you will see why in chapter 8.
```
steveklabnik: "You will see why in chapter 8" promises an explanation that chapter 8 never gives. Its "Method receivers, revisited" section (line 349) restates the rule and says nothing about *why* methods differ from free functions. Either give the reason here (the honest one today is "this is how the language currently is; see RUE-2067 for whether it should change") or drop the forward reference.

---

website/content/tutorial/08-ownership.md
```diff
@@ -0,0 +1,356 @@
...
+template = "tutorial/page.html"
++++
+
+# Ownership and Access Modes
+
+This is the chapter where Rue is most different from languages you may know.
+Every value has one owner. Passing a value gives it away, unless you say you
+are only lending it, and lending comes in two flavors: read-only `borrow` and
+read-write `inout`. All of it is checked at compile time, and all of it is
+written at the call site.
```
steveklabnik: "All of it is written at the call site" is contradicted by this chapter's own last section (line 349), which says method calls do not repeat the keyword. Suggest "written at the call site for free functions" here, so the reader is not told two different things forty lines apart.

---

website/content/tutorial/10-strings.md
```diff
@@ -0,0 +1,287 @@
...
+
+```text
+HELLO, RUE
+length: 10
+```
+
+## `str` and `StrBuf`
+
+A literal is a `str`. Giving it a `StrBuf` type, or using it where a `StrBuf`
+is expected, copies it into a fresh heap buffer:
```
steveklabnik: "Copies it into a fresh heap buffer" is not what happens. Per spec 3.10, a `StrBuf` made from a literal has capacity zero and points at the read-only literal bytes; it is promoted to the heap on the first mutation. Confirmed with the trunk compiler: `let owned: StrBuf = "heap text"; owned.capacity()` prints `0`. The comment on line 54 ("copied from the literal") has the same problem.

Suggest: "wraps the literal as a `StrBuf` without copying; the first `push` allocates." That is also a nicer story, since it means the `str`/`StrBuf` distinction costs nothing until you mutate.

---

website/content/tutorial/10-strings.md
```diff
@@ -0,0 +1,287 @@
...
+heap-owned string from the standard library. `println` accepts both, `+`
+produces a `StrBuf`, and reading input gives you a `StrBuf`.
+
+```rue run
+const std = @import("std");
+const StrBuf = std.strbuf.StrBuf;
+
+fn shout(borrow s: StrBuf) -> StrBuf {
+    let mut out = StrBuf.new();
+    for b in s.clone() {
```
steveklabnik, line 20: `for b in s.clone()` is the workaround for RUE-2052 (`for` over a `borrow StrBuf` is rejected), and the process doc records that, but the prose here does not. A reader will reasonably ask why the first string function in the chapter clones its argument to read it. Editorial rule 5 in `docs/process/tutorial.md` asks for these to be named in the text; a one-line aside ("today `for` cannot walk a borrowed `StrBuf` directly, so we iterate a copy") would do it, and it gives the next author the cue to remove it when the bug is fixed.

---

website/content/tutorial/10-strings.md
```diff
@@ -0,0 +1,287 @@
...
+
+```text
+1, 2, 3, 4, 5
+```
+
+## Bytes and characters
+
+A `StrBuf` is a sequence of bytes, and `for` visits them as `u8`. For ASCII
+text that is usually what you want; a byte compared against `48` through `57`
+is a digit, `32` is a space, and so on.
```
steveklabnik: Rue has byte literals (spec 2.1:26): `b'0'`, `b'9'`, `b' '`, `b'+'`. They are ordinary integer literals, so `b >= b'0' && b <= b'9'` works anywhere `48` and `57` do. I checked that they compile and run on trunk. The spec's own example for the feature is an `is_digit` written this way.

The tutorial never mentions them and instead teaches "48 through 57 is a digit, 32 is a space" here, then carries the magic numbers into `shout` (line 21), `is_digit` (chapter 14 line 100), `is_operator` (chapter 14 line 104), the `apply` branches (chapter 14 lines 73 to 79), and the module version in chapter 15. Introducing `b'c'` in this paragraph and using it from here on would make the calculator considerably easier to read, and it removes a whole class of "what is 47 again" from the project chapters.

---

website/content/tutorial/14-project-a-calculator.md
```diff
@@ -0,0 +1,275 @@
...
+        } else if op == 45 {
+            left - right
+        } else if op == 42 {
+            left * right
+        } else {
+            if right == 0 {
+                return StepResult.Err(EvalError.DivisionByZero);
+            }
+            left / right
+        };
```
steveklabnik: With byte literals (see my comment on chapter 10 line 121) this becomes `if op == b'+' { ... } else if op == b'-'`, and the paragraph on line 220 that decodes 43, 45, 42, and 47 for the reader can go. Since `op` is a `u8` with four known values, a `match op { b'+' => ..., b'-' => ..., b'*' => ..., _ => ... }` would also let the project use `match` on something other than an enum, which the tutorial has not shown since chapter 5.

---

website/content/tutorial/05-functions-and-control-flow.md
```diff
@@ -0,0 +1,331 @@
...
+
+Both branches must have the same type, and an `if` used as a value must have
+an `else`:
+
+```rue compile-fail E0206
+fn main() -> i32 {
+    let x = if true { 1 } else { false };
+    0
+}
+```
```
steveklabnik: The prose makes two claims (branches must agree, and a value `if` needs an `else`) but the example only demonstrates the first, and it is the only compile-fail fence in the chapter without the diagnostic text the reader will see. Either show the E0206 message like the other examples do, or split into two small fences so each claim has its error.

---

website/content/tutorial/12-modules-and-std.md
```diff
@@ -0,0 +1,199 @@
...
+`RUE_STD_PATH` points at), and everything in it is reached through the
+binding:
+
+```rue run
+const std = @import("std");
+
+fn main() -> i32 {
+    println("gcd: " + @to_string(std.math.gcd(i32, 84, 36)));
+    println("max: " + @to_string(std.math.max(3, 9)));
+    println("hex: " + std.fmt.to_hex(255));
```
steveklabnik, line 125: Nit: these two lines show `gcd` taking an explicit type argument and `max` inferring it, side by side, with no comment. The reader has just been told generics are functions from types to types, so they will wonder why one needs `i32` and the other does not. Either pick two functions with the same shape or add half a sentence. Not the tutorial's fault, but it is the tutorial's problem.

---

website/content/tutorial/16-whats-next.md
```diff
@@ -0,0 +1,76 @@
...
+- **A prelude.** `Option`, `StrBuf`, and `ArrayBuf` must be imported and
+  named. This is a deliberate stance, though the ceremony may shrink.
+- **Iteration over `ArrayBuf`.** `for` walks fixed arrays and strings. For a
+  buffer, index over `len()`.
+- **Slices of narrow elements.** `[T]` currently works only when `T` is 64
+  bits wide.
+- **Package management.** A program is one root file and its imports. There
+  is no package registry or dependency resolution.
+- **Concurrency.** None. Rue programs are single-threaded, which is also why
+  its exclusivity rules need no runtime checks.
```
steveklabnik: Nit: "which is also why its exclusivity rules need no runtime checks" implies that adding threads would force runtime checks. The exclusivity rules are static because loans are second-class and the compiler sees both ends of every one (chapter 8 says this well); Rust keeps the same property with threads. Suggest dropping the clause or saying "and the exclusivity rules are what would make a future concurrency story safe."

---

steveklabnik (commented):
Read all sixteen chapters and the process doc against the spec, and checked the claims I doubted with the trunk compiler. The structure is a large improvement: concept order is right, every fence being verified is the correct discipline, and the calculator project earns its two chapters. Chapters 8 and 11 in particular are the clearest explanations of the access model and the value/trap split anywhere in the repository.

The comments fall into three groups.

**Factually wrong today** (prose that the snippet checker cannot catch):

- Chapter 6 line 216: `Point.new(1, 2).manhattan()` does not compile (E0427, spec 6.4:27). A `borrow self` receiver must be a place.
- Chapter 10 line 45: a literal typed as `StrBuf` is not copied to the heap. Capacity is zero until the first mutation (spec 3.10; verified `capacity()` prints `0`).
- Chapter 1 line 35: "every caller must write `inout` too" is false for methods (spec 6.4:25). The tutorial then says the honest version in chapters 6, 8, and 15, so the chapters disagree with each other. I filed RUE-2067 for the language question this raises, since either the claim or the language should change.

**Framing, per the PR's "things to review"**: the "Where things stand" section says the tutorial marks which principles are guarantees, but nothing in the locality list is marked. The call-site claim and "nothing converts silently" (string-literal contextual typing) are the two that need a marker; determinism I could not verify either way.

**Missed feature**: Rue has byte literals (`b'0'`, spec 2.1:26) and the tutorial teaches raw ASCII numbers instead, then carries them through both project chapters. Using them would make the calculator much easier to read.

The rest are nits. No comments on the checker changes or the process doc, which look right to me.

---

steveklabnik:
Addressed all fourteen comments in 799ffe6. Verified each claim against the trunk compiler before changing the prose.

**Factually wrong today**
- Chapter 1 call-site claim: took option 2 (keep autoref, soften the prose). Free functions carry the mode at both ends; methods carry it in the signature and require a `let mut` receiver. The Python example is gone, and the paragraph points at RUE-2067. Chapters 6 and 8 now say the same thing, and chapter 6's dangling "you will see why" is replaced with the honest reason.
- Chapter 6 chaining: says chaining works for by-value methods returning `Self`, and that a `borrow self` or `inout self` method needs a place, so `Point.new(1, 2).manhattan()` is rejected. Confirmed E0427 and confirmed a by-value chain compiles.
- Chapter 10 `StrBuf` literal: "wraps the literal without copying; capacity is zero until the first mutation." Confirmed `capacity()` prints `0` before and `16` after a `push`.

**Framing**
- "Where things stand" now lists the three marked items explicitly: call-site modes (methods), numeric-only "no implicit conversions" with the string-literal rule named, and determinism as direction. For determinism I could only confirm that `std.rand` takes an explicit seed; the maps don't document iteration order, so the text says neither is promised yet.
- "Nothing converts silently" narrowed to numbers, with the `str`/`StrBuf` contextual rule named and a note that `println` is a built-in rather than an overload.
- "Jump out of your function" is now "return to your caller", with a pointer to traps.

**Byte literals**
- Chapter 10 introduces `b'0'`, `b' '`, `b'+'` where bytes are first discussed, and the opener's `shout` uses them (with a sentence pointing forward, plus the RUE-2052 clone aside you asked for).
- Chapters 14 and 15: `is_digit`, `is_operator`, `b - b'0'`, `b != b' '`, and `apply` is now a `match op { b'+' => ..., b'-' => ..., b'*' => ..., _ => division }`. The paragraph decoding 43/45/42/47 is gone. The test uses `m.apply(b'+')`.

**Nits**
- Chapter 5: the `if` example is split into two compile-fail fences, each with its diagnostic (branch mismatch, and `if` without `else` typed as `()`).
- Chapter 12: half a sentence on why `gcd` takes a type argument and `max` does not.
- Chapter 16: dropped the runtime-checks clause; now "the exclusivity rules are what would make a future concurrency story safe."

Checker: 88 snippets pass (70 run, 3 compiled, 15 compile-fail); tool tests pass.

---

steveklabnik:
The merge queue dequeued this PR on a failure that is not this PR's.

**Failing check:** `native (macos-arm64)` in queue run [33980197550](https://github.com/rue-language/rue/actions/runs/33980197550). `//:corpus-action-tests` hung for its full 10-minute budget with empty stdout and stderr; the other 14 tests in the job passed, and every other job was green.

**Why it is not this PR's:** that target runs `scripts/test-corpus-action.sh` over `scripts/corpus-action`, `scripts/corpus-timeout.py`, and `scripts/corpus-stamp-check`, none of which this PR touches (the diff is tutorial markdown, the snippet checker and its tests, `docs/process/tutorial.md`, a BUCK comment, and a comment in `examples/welcome.rue`). The same job passed on this PR's own head a few minutes earlier and on the queue run immediately before it, which is the trunk this run merged against. Locally, trunk merged into this branch passes the tutorial checker.

**Fix:** none exists yet. Filed RUE-2068 with the run, the artifact, and where to look (the test's sleep-loop harnesses and the portable timeout wrapper). Re-queuing once as the one re-run for an unrelated failure.

---

PR #3103, opened 2026-09-11
RUE-2174: make listing, presentation and the toolchain probe cancellable

steveklabnik:
The cancellation boundary audit from RUE-2128 (ADR-0085 phase 2), as its own slice ahead of the service.

The executable and test-image adapters already ran under a request's `CompilationCancellation`, but three paths a daemon request will take still minted their own uncancelable token: the rooted test inventory behind `rue test --list`, the presentation batch behind `--emit air` (through `rooted_cfg` and `rooted_codegen`), and the semantic probe inside toolchain acquisition, which only checked its supersession probe between rounds, so a long body-closure analysis ran to completion after the caller had gone away.

**Compiler.** Cancellable forms go through the same computations rather than new ones: `cancellable_test_inventory` keeps the listing's partial-inventory semantics (a broken test is still listed beside its diagnostics) and reports `Canceled` instead of an inventory or diagnostics; `present_many_with_cancellation` is now the one presentation implementation, with the differential oracle's injected semantic fault kept on the same path, and `unstable_present_many` wraps it with a token nobody can cancel; `rooted_or_toolchain_park_with_cancellation` adds `RootedParkOutcome::Canceled`, which attaches no park and settles nothing. The plain `rooted_or_toolchain_park` and `rooted_codegen` session wrappers had only test callers left and are gated to tests, so production has one path each.

**Driver.** Toolchain acquisition takes the request's cancellation: every checkpoint observes it and the probe runs under its token, reporting `SourceLoadError::Superseded` like the other checkpoints (never a source diagnostic, never the uncanceled-path panic). Watch switches to that path, since its supersession probe already was the token. The host exposes `cancellable_test_inventory`, `cancellable_present_many`, and `acquire_reached_toolchain_modules_cancellable`. The loader's own round-transaction tests keep a test-only superseding wrapper because they trip at counted checkpoints.

**Coverage.** A test-only tripwire in the rooted projections cancels a request deterministically just before the shared body-closure request or before the CFG batch, mirroring the existing codegen tripwire. Compiler tests cancel the probe, the listing, and the presentation before admission, during body-closure work, during CFG work, and inside the backend kernel, then verify the next request over the same session equals a fresh session's. Driver tests do the same over a retained `FilesystemCompilerHost` and check that the listing's broken-test diagnostics never leak into the executable request; a loader test covers the token path's no-partial-commit and recovery.

Verified locally: full `rue-compiler` unit suite (1249 passed), both driver unit suites, `scripts/rue cli watch`, and the clippy/fmt gates for `rue-compiler` and `rue`.

Fixes RUE-2174

### DIFF
--- a/crates/rue-compiler/src/session.rs
+++ b/crates/rue-compiler/src/session.rs
@@ -57,7 +57,11 @@ use import_discovery_owner::ImportDiscoveryOwner;
 pub use metrics::*;
 use program_artifacts::no_published_program;
 pub use rooted_artifacts::*;
-pub(crate) use rooted_projections::{TestClosureAnalysis, TestClosureAnalysisOutcome};
+#[cfg(test)]
+pub(crate) use rooted_projections::{RootedCancellationSite, set_rooted_cancellation_tripwire};
+pub(crate) use rooted_projections::{
+    TestClosureAnalysis, TestClosureAnalysisOutcome, TestInventoryOutcome,
+};
 #[cfg(test)]
 use rooted_projections::{stable_function_definition_root, stable_producer_definition_root};
 
--- a/crates/rue-compiler/src/session/import_discovery_owner/tests.rs
+++ b/crates/rue-compiler/src/session/import_discovery_owner/tests.rs
@@ -73,6 +73,7 @@ fn closed_continuation_for(
         RootedParkOutcome::Errors(errors) => {
             panic!("expected a trusted-toolchain park, got errors: {errors:?}")
         }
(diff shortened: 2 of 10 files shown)


steveklabnik:
The merge-queue run for this PR (CI run 34563015524) failed only in `release (linux-x64)`: the BuildBuddy remote cache answered `503 Service Unavailable` with a malformed compression flag while downloading the prebuilt `third-party:toml-0.8.23` artifact, so the release smoke build stopped before any test ran. Nothing in this diff touches third-party crates or the release configuration, and the queue runs for #3096 and #3098 failed identically a few minutes earlier and passed on re-queue. Re-queued once.

---

PR #3138, opened 2026-09-17
RUE-2212: array pointees enter inference structurally

steveklabnik:
## Problem

`@ptr_read` published its result, and `@ptr_write` constrained its value, as the pointer's interned pointee `Type`. Every other array position in constraint generation is the structural `InferType::Array` shape, and unification never compares an interned array against a structural one. So a whole-array pointee was rejected the moment it met an annotation or an already-typed array value (`expected [i64; 4], found [i64; 4]`), and a written array literal never took the pointee's element type (`expected [i64; 2], found [{integer}; 2]`). Together these made it impossible to `@ptr_write` or `@ptr_read` a whole array.

The issue's mechanism guess (distinct interned ids compared in sema) does not hold: the ids dedup and sema's check was never reached, because inference rejected first.

## Fix

Both sites now go through `type_to_infer`, which decomposes arrays exactly as annotations do. Element and length mismatches are still E0206 / E0901, and the scalar annotation check (RUE-244) is unchanged.

## Tests

Five spec cases in `runtime/pointers.toml` (9.1:10, 9.2:6b, 9.2:6d): read into an annotated binding, write from a typed value, write an array literal, a struct-element array round trip, and an element-type mismatch that stays rejected.

Verified locally: rue-air unit tests, the pointer/unchecked/raw spec, CLI and UI suites, the staged example corpus build, and the oracle differential.

Fixes RUE-2212
Fixes RUE-2213


steveklabnik:
Dequeued once by `rue_program digest sensitivity (linux-x64)` in merge-queue run 8950: the job exited 2 twenty-one seconds into its first `buck2 build` of `fixtures/rue-program:hello`, before any compile could have finished, so it was a buck2 invocation error rather than a build or test result. The same job passed on the next queue batch (run 8951), whose tree carried this exact change on top of trunk, and `fixtures/rue-program:hello` builds locally on trunk plus this branch. Re-queued without changes.

PR #3139, opened 2026-09-17
RUE-2207: a zero-sized frame root diverts its projected accesses

steveklabnik:
## Problem

The zero-sized diversions in place lowering keyed on the *selected* sub-object, but a dynamic index into a zero-length array selects a sized element (`a[i]` of `[i32; 0]` is one `i32` slot) out of a root that owns no slot at all. The frame arithmetic had nothing to anchor on: a local root underflowed the descending slot numbering (the RUE-605 ICE in `place_lower`), and a by-value parameter root asked the storage plan for a slot it never allocated (the `param_storage` ICE). Both reproduced at every optimization level, including the ordinary `for x in a` over an empty array.

## Fix

The root now decides too: a projected access out of a zero-slot local or by-value parameter goes to the canonical zero-sized address (`frame_root_is_zero_sized` in `resolved_access` and `lower_place_addr_plan_with_bounds`). Nothing ever reads or writes through it, because the level that makes the root zero-sized while the leaf is not is a zero-length array whose bounds check, already emitted by `resolved_offsets`, traps first. By-reference parameters and indirect bases carry a real pointer whatever the pointee's size and keep their addressing.

The `for` forms now run zero iterations and every dynamic read or write traps with the ordinary `index out of bounds` message, on both backends.

## Tests

- A two-backend unit test in `place_lower` builds both root shapes and asserts the bounds trap edge, the canonical address, and that no frame address is formed (x86-64 and AArch64).
- Six CLI cases in `zero_sized_place_address.toml`: the local and by-value-parameter `for` forms, the taken-branch traps for both roots, the untaken read/store/nested-store shapes (all `differential_opt` across -O0..-O3), and an AArch64 `--emit asm` case.

Verified locally: rue-codegen unit tests, the zero-sized and array CLI and spec cases, the staged example corpus build, and the oracle differential.

Fixes RUE-2207


steveklabnik:
Dequeued once by `release (linux-x64)` in merge-queue run 8956: the job failed with a Buck remote-execution materialization error while fetching a cached third-party artifact (`libthiserror` rlib, `BatchReadBlobs` h2 stream reset from the remote cache), before the link step ran, and every other job in the run passed. Nothing in this change touches the release build or that dependency. Re-queued without changes.

PR #3142, opened 2026-09-20
RUE-1882: propose ADR-0097, the mechanized formal core

DorianListens:
Proposes ADR-0097, the decision record for growing the Lean 4 mechanization of the core calculus from the 2026-08-28 spike into a proved §7 metatheory. This PR is the ADR only: it lands no Lean code. The spike itself lands under RUE-2226 once this is accepted, and the project is tracked in Linear as "Formal core mechanization".

## What the ADR decides

1. Adopt `docs/formal/lean/` as the mechanization home, with Buck managing the Lean toolchain as a third `toolchain_distribution` beside Rust and Zig (SHA-pinned, all four platforms, a zstd path added since Lean ships `.tar.zst` only, kept out of the CAS since the unpacked tree is 2.7 GB) and the build as a build-only Buck target with no test tier until the CI gate. `lean-toolchain` stays as lake's pin with a validator tying the two pins together.
2. Extend the RUE-305 "disagreement is a defect, no precedence" rule from three views to four.
3. Interpreter-style safety as the primary §7 theorem form, matching `rue-oracle`'s shape, with `eval`'s adequacy to §6's small-step reduction as an owed lemma and kernel-checked refusal witnesses per slice.
4. A sixth rubric step in the formal README: extend the mechanization, or file the gap. A language change is never blocked on a proof.
5. Experimental and non-blocking posture. Nothing in CI until the full-core safety proof exists and an independent review confirms it. The mechanism is the test-tier system: no test target exists before the gate, and RUE-2241 then wraps the build target in a test and picks its tier (`stress` on the nightly sweep, or a new experimental tier with its own lane and exclusion labels; `slow` is a pre-merge selector). The merge-queue gate is a later, separate decision.
6. Explainability as a tenet: reader's guide, generated cross-reference index, derivation/trace visualizer, expert validation surface, and a worked-example deliverable on every slice.
7. Cross-model-family adversarial review at each phase boundary, by Steve's Codex agents, as gating issues.

Also adds the ADR's row to the index in `docs/designs/README.md`.

## Review notes

- Status is `accepted` (2026-09-20, on maintainer approval), so the merged document is already ratified.
- Revised after review: decision 1 originally kept the toolchain out of Buck entirely; it now uses Buck for toolchain management and relies on tiers, not graph absence, to stay out of required CI.
- Revised again after two independent adversarial reviews (Opus and Fable): corrected the tier facts (`slow` is pre-merge, the health check blocks only never-succeeded schedules), the toolchain facts (17.5k files, zst archives, four platforms), the gate (checkpoint C, not RUE-2237), and added the adequacy lemma, refusal witnesses, and the §9 rulings (RUE-2283) as owed items. Two issues added: RUE-2282 (leaf/float inventory) and RUE-2283.
- Points most worth pushing on: decision 3 (theorem shape), decision 5's tier question, and whether decision 7's gating is too strict for the parallel lanes.

Closes RUE-1882


DorianListens:
Addressing the Codex finding relayed by Dorian ("require adequacy/non-vacuity at the CI promotion gate"). Agreed: as written, gate condition (a) was a theorem about `eval`, and an interpreter returning `outOfFuel` before a bad transition could satisfy it without establishing anything about §6's dynamics. Resolved in the latest commit by taking the first of the two options and narrowing the claim as well:

- **Decision 3** now states that a theorem about `eval` is a theorem about §6 only once the two are proved to agree, and makes adequacy (a mechanized §6 reduction relation, `eval` sound and complete modulo fuel with respect to it) a Phase C deliverable that checkpoint C requires, rather than a Phase E promise. Filed as RUE-2289, blocked by every other Phase C slice and blocking checkpoint C and the CI issue.
- **Decision 5** gains condition (b): the fuel monotonicity and no-masking lemmas (RUE-2233) and the adequacy lemma (RUE-2289) must be proved for the same fragment before promotion, and the sentence after it narrows what promotion certifies: §6's dynamics are safe on the by-value fragment, with loans and buffers outside the claim.
- The metatheory skeleton and RUE-207 no longer list adequacy as owed to Phase E; checkpoint C's brief (RUE-2251) checks the lemmas against §7's own phrasing.

PR #3169, opened 2026-09-21
Fix cycle for the 2026-09-20 hunt: escaped-pointer forwarding, foreign-symbol inlining, zero-length arrays, interfaces preview gaps, for binders, ELF common symbols, accessor and slice typing

steveklabnik:
Fixes for the bugs filed from the 2026-09-20 hunt on trunk `168e9ea`. One commit per issue; each fix was reproduced before the change, falsified where a causal story was involved, and re-verified by running the original repro on the integrated tree.

## Optimizer and inliner (crates/rue-cfg, crates/rue-compiler)

- RUE-2262: -O2/-O3 block-local store forwarding no longer forwards across a local whose address escaped through a callee's `inout` parameter. The pre-scan in `forward.rs` now collects an escape set (address-taken slots plus every local rooting a by-ref call argument) that never enters Rule 2's table; indirect `PlaceWrite`s and pointer-taking intrinsics are barriers; `slot_facts::classify_slot_writes` fails closed on an unresolvable by-ref root. CLI cases pin 42 at every level for all three repro shapes (oracle model gaps registered: the oracle does not model `@ptr_write`).
- RUE-2264: `AccessorCall` is now observable in `classify.rs` (unit-tested; unreachable today because accessors are spliced before `optimize()`).
- RUE-2271: the inline splice merge lets several source symbols share one foreign machine symbol when the calling conventions agree, so a symbol declared in two `extern "C"` blocks no longer ICEs at -O3 (spec 9.3:5). CLI cases on both backends.

## Codegen and layout (crates/rue-air call_abi, crates/rue-codegen)

- RUE-2256: `slot_layout_agrees` answers for a zero-length array's own zero extent (the `FfiTypePool` trait gains `ffi_array_len`), so `@raw` of `[i32; 0]` compiles instead of E9001; `[i32; 4]` still reports E9001 (RUE-1595) and is pinned.
- RUE-2243: both CFG lowerings divert a zero-slot pointer read before the compact-image branch, so `ArrayBuf([i32; 0])` push works; oracle model gaps registered for the new cases.

## Parser, CLI, linker

- RUE-2254: a struct pattern as a `for` binder is a syntax error (E0102) instead of an astgen panic.
- RUE-2242: under `--preview interfaces`, `comptime p: P` with `P` a struct is a comptime value parameter again; the bound reading applies only when the name is an interface. The spec case that encoded the old behaviour now uses a composed bound, and the E0306 explanation example follows.
- RUE-2261: `rue --help` describes the daemon's real support table; docs/development.md gains the `--module-manifest` row.
- RUE-2257: ELF `SHN_COMMON` symbols are tentative definitions allocated in the merged bss (largest extent wins, a real definition displaces them); `SHN_XINDEX` is refused with a clear parse error instead of reading as undefined. Linker unit tests with hand-built objects.

## Interfaces preview and sema (crates/rue-air sema, crates/rue-error)

- RUE-2258: freestanding conformance assertions on enum subjects are verified (E0302 at the assertion) instead of silently dropped.
- RUE-2269: new E0308 rejects `==`/`!=` on a value whose type is or contains an opaque bounded `T` (6.8:20).
- RUE-2268: new E0309 refuses `@raw`/`@raw_mut`/`@field_ptr`/`@ptr_offset`/`@ptr_read`/`@ptr_write` on an opaque bounded pointee; trusted std is exempt because `ArrayBuf`'s read path is skolem-instantiated (documented in 6.8:20).
- RUE-2260: a receiver method passed to a `fn(borrow T, ...)` callback parameter of a generic function reports the ADR-0096 E0216.
- RUE-2259: an accessor in an anonymous struct reports E0257 at the signature instead of a generic comptime failure.
- RUE-2223: `Self` resolves in struct field types, in type-valued associated-constant initializers, in enum payloads and inside anonymous structs, through the one existing type-substitution hook; 6.4:18's enumeration is updated to match.
- RUE-2248 (assignment half): storing a non-`str` value into a `str` variable is E0206 (`analyze_assign`), placed after the first-class-`str` checks so buffers keep E0495 and views keep E0496. The `let x: str = 5` half is covered by #3145.

## Accessors and slices (crates/rue-air inference, sema, types)

- RUE-2255: an accessor result enters inference structurally, so `-> borrow [i64; 3] { yield self.cells; }` type-checks; `!` unifies against a structural array.
- RUE-2263: the yield place's constant indices are bounds-checked at compile time (7.1:9).
- RUE-2270: slice-ness is decided by the outer bracket pair, so `[[i64; 2]]` is a slice with `.len()`, indexing, and coercion.

## Comptime types (crates/rue-air sema, crates/rue-compiler durable comptime)

- RUE-2266: `Str(N)` names a type in expression position, so it is a valid argument to `comptime T: type` parameters and type constructors (`Option(Str(4))`, `ArrayBuf(Str(4))`, `let T = Str(4);`).
- RUE-2267: a builtin nominal (`str`, `Str(N)`, builtin enums) is a durable type argument, so `Option(str)`, `Result(str, E)` and user constructors over `str` reduce.
- RUE-2265: a float literal bound to a `comptime x: f64`/`f32` parameter on the durable const path takes the parameter's width (3.12:7), so `const R: f64 = ident_f(16.0);` compiles.

## Bookkeeping

- The RUE-914 planted-miscompile preimage hash follows the `slot_facts.rs` edit (the patch still applies).
- `_CLI_TESTS_TIMEOUT_SECONDS` 5000 -> 6300 per the documented 25% headroom policy (derived deadline 5021s after the new cases).
- Pinned host-contract method count 80 -> 83 (three new comptime host hooks) and the semantic-nucleus registration identity pin, both updated after the tests named them.

## Verification

Full `./test.sh` runs on this branch are recorded in the PR comments. Two CLI cases fail only in the integration container and are unrelated to the diff: `cli.fs_remove_dir_all::remove_dir_all_permission_denied_is_not_partial_success` (the container runs as root, so the chmod-0 directory is still removable) and `cli.std_net_tcp::tcp_ipv6_loopback_round_trip` (no IPv6 loopback in the container).

RUE-2244 needs no change here: #3145 (RUE-2209) already makes the pointer-to-array annotation shapes E0206, and the "written through a const pointer" part of that report was wrong (`@ptr_write` rejects it with E0702).

Fixes RUE-2262
Fixes RUE-2264
Fixes RUE-2271
Fixes RUE-2256
Fixes RUE-2243
Fixes RUE-2254
Fixes RUE-2242
Fixes RUE-2261
Fixes RUE-2257
Fixes RUE-2258
Fixes RUE-2269
Fixes RUE-2268
Fixes RUE-2260
Fixes RUE-2259
Fixes RUE-2223
Fixes RUE-2255
Fixes RUE-2263
Fixes RUE-2270
Fixes RUE-2266
Fixes RUE-2267
Fixes RUE-2265
Part of RUE-2248


steveklabnik:
Verification record for head `4a6f6c9` (21 commits on trunk `168e9ea`), run in the integration container (4 cores, root, no IPv6 loopback):

- `./test.sh`: `Tests finished: Pass 237. Fail 1.` The one failure is `crates/rue:rue-driver-test :: daemon::tests::accepted_connections_have_a_hard_bound_and_a_retryable_rejection` ("the compiler service already has 33 connections"), a connection-bound timing test that ran while another worktree was building; it passed in the two earlier full runs on this branch and passes when re-run alone (`buck2 test //crates/rue:rue-driver-test` → Pass 1, Fail 0). The banner is therefore FAILED rather than PASSED on this host.
- Within the CLI corpus, `cli.fs_remove_dir_all::remove_dir_all_permission_denied_is_not_partial_success` and `cli.std_net_tcp::tcp_ipv6_loopback_round_trip` fail only because the container runs as root and has no IPv6 loopback; every other CLI case passes (3035 passed, 10 ignored).
- Targeted on the final tree: `//crates/rue-air:rue-air-test` 928 passed; `//crates/rue-compiler:rue-compiler-test` 1278 passed / 60 ignored; `//crates/rue-oracle-diff:*` (unit, corpus, -O2, -O3, spec, planted-miscompile) all Pass; `//crates/rue-spec:*` Pass; `//:planted-miscompile-isolation-validation`, `//:cli-timeout-policy-validation`, `//crates/rue-compiler:error-explanation-examples-test`, `//crates/rue-error:rue-error-test` Pass; `//examples:cli-staged-programs` builds; `scripts/rue fmt` clean.
- Every issue's original repro was re-run by hand on the integrated compiler at -O0 and -O3 (and compile-only on aarch64 where the fix touches codegen), plus cross-mechanism probes (accessor yielding `[Str(4); 2]`, `ArrayBuf(Str(4))` with `get_ref`, `Option(str)` inside a bounded body, a `[[Str(4); 1]]` slice, a comptime struct parameter next to an interface bound).

Not in this PR: RUE-2244 and RUE-2248 (annotation holes) are being worked on against #3145.

steveklabnik:
Verification record for head `dba7868` (the branch merged with trunk `2dcf78c`; 24 commits):

- GitHub CI on this head: all 30 check runs green, including `premerge (linux-x64)`, `native (linux-arm64)`, `native (macos-arm64)`, `release`, `asan`, `valgrind`, `compiler reproducibility`, the four CLI shards, `spec`, and all six `oracle-diff` variants (`CI success` at 07:09Z, run 35570629772).
- Merge resolution: trunk had already raised `_CLI_TESTS_TIMEOUT_SECONDS` to 6300 (kept); the host-contract method count is 85 (trunk's 82 plus the three hooks here); `builtin_nominal_identity` now routes its enum arm through trunk's `is_builtin_enum_id` so there is one builtin-enum predicate; the semantic-nucleus registration identity pin was re-derived after the merge.
- On the merged tree, locally: every issue's original repro re-run by hand (all 27 shapes, at -O0/-O3 where relevant, plus both FFI archive links); `//crates/rue-air:rue-air-test`, `//crates/rue-compiler:rue-compiler-test`, `//crates/rue-error:rue-error-test`, `//:planted-miscompile-isolation-validation` and `//:cli-timeout-policy-validation` pass. A local `./test.sh` on the merged tree was started twice and killed both times by the integration container being reclaimed between wakes; CI's full run above stands in for it.

PR #3179, opened 2026-09-21
RUE-2305: a returned array enters inference structurally

steveklabnik:
Fixes RUE-2305

## Problem

An explicit `return <fixed array>;` in an ordinary function was rejected with a type mismatch that printed the same type on both sides:

```rue
fn mk() -> [i64; 3] {
    let a: [i64; 3] = [1, 2, 3];
    return a;
}
```

```
error: [E0206]: type mismatch: expected [i64; 3], found [i64; 3]
```

and `return [1, 2];` in a `-> [i64; 2]` function never took the result's element type (`expected [i64; 2], found [{integer}; 2]`). The trailing-expression form of the same functions compiled.

## Mechanism

`InstData::Ret` in `crates/rue-air/src/inference/generate.rs` constrained the returned value against `InferType::Concrete(ctx.return_type)`, the interned `Type`. Every other array position in constraint generation is the structural `InferType::Array` shape, and `Unifier::unify_with` has no arm relating an interned array to a structural one, so the pair fell through to the array-versus-non-array catch-all. This is the same interned-versus-structural hole RUE-2212 closed for `@ptr_read`/`@ptr_write` pointees and RUE-2255 closed for an accessor's `yield`.

Confirmed before changing anything by temporarily forcing `type_to_infer` to publish arrays as `Concrete`: previously-working programs then fail on the same axis (a struct literal's `[i64; 2]` field reports `expected [i64; 2], found [{integer}; 2]`).

## Fix

The `Ret` arm publishes the declared result through `type_to_infer`, which decomposes arrays exactly as annotations do. That is the whole code change.

A `return` whose operand diverges or names an undefined variable meets the now-structural array through the `!`/recovery-type admission that RUE-2255 added to `unify_with` (now on trunk). I verified it is load-bearing here too: without it, both cases reported a second mismatch (`found !` / `found <error>`) after the real error. This PR originally carried its own copy of that arm; the rebase onto trunk dropped it in favour of trunk's.

Element and length mismatches are unchanged: a returned array of the wrong length is E0901 and of the wrong element type is E0206.

## Tests

- `crates/rue-spec/cases/expressions/return.toml`: whole-array binding, literal taking the result's element type, early return inside an `if`, nested arrays, a diverging operand, and the wrong-length and wrong-element-type rejections, cited to 4.9:4 (with 7.1:4 and 3.4:3 where relevant).
- `crates/rue-cli-tests/cases/array_return.toml`: a new `cli.array_return` section with two `differential_opt = true` cases, since the behaviour is runtime-visible.

## Verification

On the rebased head:

- `./buck2 test 'root//crates/rue-air:'`: passed.
- `scripts/rue spec 4.9`: 19 passed.
- `scripts/rue cli array_return`: 4 passed (both new cases at `-O0` through `-O3`).
- `./buck2 build //examples:cli-staged-programs`: succeeded.
- `//crates/rue-oracle-diff:oracle-diff-test` and `//crates/rue-oracle-diff:oracle-diff-spec-test`: passed on the pre-rebase head, whose corpus changes are identical.
- `scripts/rue fmt` run before committing.


steveklabnik:
Dequeued by the merge queue at 22:21 UTC on a failure that is not this PR's.

The queue run's only red job is `performance staleness (linux-x64)`, the repository-wide series gate: every platform's newest plotted point is `7c2727233cfe` (the RUE-2282 merge), 23 trunk commits behind with a threshold of 5, and older than the two-hour grace window. The gate says so itself: "This is a repository-wide condition and almost certainly not caused by this pull request. Every pull request fails while it lasts."

Cause: the `Performance collection` run for the current trunk head `c2e40c5f` ([run 35654101467](https://github.com/rue-language/rue/actions/runs/35654101467)) collected on all six lanes but its `publish` job failed on `actions/download-artifact` with a 403 from the artifact service, so no point for that commit was ever published. The measurements are retained as that run's artifacts, so re-running the failed `publish` job should publish them and advance the series; nothing in this PR can port that fix, and the session's GitHub integration is not permitted to re-run workflow jobs (403).

Once the series advances I will re-enqueue this PR. Any other PR queued meanwhile (#3145 is next) hits the same gate.

PR #3181, opened 2026-09-22
RUE-2321: carry edge deliveries in the typing judgment; loop and function rules in judgment form

DorianListens:
Rewrite the core calculus's divergence and loop statics into judgment form, so the next slice of the Lean mechanization (RUE-2326) has rules it can transcribe. Calculus only: `docs/formal/01-core-calculus.md`, no Lean, no spec change.

Closes RUE-2321

## What changes

- **§5.3.** The outgoing result `Ω` of every typing judgment carries a set `Δ` of *edge deliveries* `⟨κ, Σ⟩` (κ one of `ret`, `break`, `continue`, `diverge`, `panic`), each recording the ownership state in force where a reachable diverging edge fires. The old provenance set `δ` becomes the projection `δ(Δ)` that forgets the states, so every consumer that only asks *whether* a kind of edge is reachable (§5.5's join, §5.6's panic exemption) reads as before, and the two consumers that need the state *at* an edge, (Fn) and (Loop-Break), read `Δ` itself.
- **§5.7.** (Return-Value), (Break), (Loop-Div-Backedge), (Loop-Div) and (Loop-Break) are restated over `Δ`. `edge-observations`, the projection defined over the body's *derivation* rather than its judgment, is gone: (Loop-Break)'s back-edge set `B` and exit set `X` are read off the body's `Ω`, reachability is a consequence of the bottom rules typing nothing past a diverging subexpression, a loop removes its own `break`/`continue` deliveries from what it passes outward (nesting), and the `X = ∅` case delivers `⟨diverge, Σ⟩` at the loop's entry state. (Loop-Div-Backedge) gains the back-edge invariance premise the compiler already enforces on a `break`-less loop.
- **§5.8.** (Fn) reads its edge snapshots from `Ωf` instead of an `Edge(D_body)` derivation set, with the `ret` and `diverge` deliveries' states as the residual-linear check's `Ends`; (Call-Bottom) and (Panic) deliver `⟨diverge, Σ⟩` and `⟨panic, Σ⟩`.
- A "rewritten into judgment form" note in §5.7 says what the old wording left ambiguous: which edges count as reachable, what state a diverge edge carries into (Fn), and whether a `break`-less loop is subject to back-edge invariance.

## What does not change

One verdict, toward the compiler: the old (Loop-Div-Backedge) accepted a move of an outer binding in a `break`-less loop body; the new rule rejects it as the compiler does (E0205 "moved in a previous iteration"). Every other one of the nineteen probe programs recorded on RUE-2321 (the RUE-1615 and RUE-1614 shapes, moves and reinitialisations across the back edge, the break-edge join on affine and linear bindings, `return` and `@panic` inside a body, a live linear parameter at `loop { }`, `continue`, `@panic` in a body, an unreachable `break` beside a `return`, nested loops) agrees with the old text's reading and with the compiler. Rule labels are unchanged, so the Lean cross-reference index is unchanged. The parent issue's "`break e`" was a slip: the surface has no value-carrying `break` (E0502, `4.8:22`) and the core never had one.

## Validation

`scripts/validate-doc-links.py`; `scripts/validate-lean-xref-index.py` (index current); one Opus adversarial round on statement fidelity, internal consistency, and the spec citations. It found two blocking defects in the first draft, both fixed: the `X = ∅` case of (Loop-Break) delivered `diverge` even with no back edge, and no rule said the non-bottom rules union their subexpressions' `Δ` (now §5.3's threading convention). Six smaller fixes followed its should-fix list.

Requesting a maintainer read for statement fidelity before the Lean slice depends on it.


steveklabnik (commented):
Review by **Astra**, against `6864ecc69506936ff26c2f5d0612979745f34cd9`.

I recommend fixing the following before the Lean loop slice depends on these rules. Carrying exit kinds and ownership snapshots in the judgment is a good direction; the concerns are about statement fidelity and completing the conversion of the surrounding rules.

1. **The new threading convention excludes divergent results from rules that need to preserve them.** At lines 814–817, every bare-state premise is expanded to a continuing `Σᵢ;Δᵢ`. But `(Sub-Never)` at lines 1406–1408 still uses a bare state, whereas `return`, `break`, and diverging loops produce `⊥;Δ`. It therefore cannot derive the document's own `if c { 5 } else { return 0 }` example. `(Seq)` and `(Let)` have the same problem for their tails, independently of coercion: `(unit ; loop { unit })` cannot use `(Seq)` because its tail is bottom, cannot use `(Seq-Bottom)` because its prefix continues, and cannot use `(Strict-Bottom)` because sequence tails are explicitly excluded. Please state `(Sub-Never)` over an unchanged `Ω`, and the `(Seq)`/`(Let)` tail premises and conclusions over `Ω`, retaining deliveries from the evaluated prefix.

2. **The new back-edge equality rejects an intentionally supported safe idiom, so the compiler-agreement claim needs reconciliation.** The added premise at line 1255 requires the back-edge state to equal the initial entry state. This program is accepted by the compiler built at this PR's revision:

   ```rue
   struct Data { x: i32 }
   fn consume(d: Data) -> i32 { d.x }
   fn main() -> i32 {
       let mut d = Data { x: 0 };
       loop {
           d = Data { x: 1 };
           consume(d);
       }
   }
   ```

   Initially `d` is `Owned`; at every back edge it is `MovedOut`. Each subsequent iteration reinitializes it before use, so the compiler accepts it. This is deliberate: `crates/rue-air/src/sema/control_flow.rs:1105–1138` uses unequal states to trigger reanalysis from the back-edge state, rather than rejecting inequality itself. The existing `reassign_before_move_ok` case in `crates/rue-cli-tests/cases/loop_moves.toml:455–468` explicitly protects reassignment at the beginning of each iteration.

   The strict equality already appears in prose `3.8:79` and `(Loop-Break)`, so the broader specification/compiler disagreement is pre-existing. This PR newly extends it to breakless loops, however, and the before/after note's claim that only one verdict changes toward compiler agreement is incorrect. Please reconcile this distinction explicitly. I recommend preserving safe reinitialization-before-use behavior and expressing iteration safety through validation from the reachable back-edge state or an appropriate invariant, rather than tightening the compiler merely to satisfy equality.

One smaller clarification: lines 1037–1042 define the branch delivery set as the union of the arms' deliveries. Please explicitly include the condition/scrutinee's deliveries (`Δ₀ ∪ ⋃Δᵢ`) too. The general threading convention might be intended to supply this, but an early return inside a partially continuing condition must not disappear before `(Fn)` checks its ownership snapshot.

Validation: built the compiler at the reviewed revision; checked five focused compiler probes, including the accepted reinitialization example and the rejected early-return/linear-value cases; both `validate-doc-links.py` and `validate-lean-xref-index.py` passed.

DorianListens:
Thanks. All three points are resolved in 1773445f8, which is rebased onto current trunk.

**1. Divergent results through `(Sub-Never)`, `(Seq)` and `(Let)`.**
- `(Sub-Never)` is now stated over an unchanged `Ω`. A divergent `⊥;Δ` stays divergent, with its deliveries, when it is re-typed. So `if c { 5 } else { return 0 }` derives, and the join sees the `ret` edge.
- `(Seq)` and `(Let)` now require only the *prefix* to continue (`⊣ Σ1;Δ_1`). The tail is typed to any `Ω_2`, and the conclusion is `Ω_2 ⊕ Δ_1`, where `⊕` adds the prefix's deliveries and keeps the tail's continuing or divergent shape.
- So `(unit ; loop { unit })` derives by `(Seq)` with `Ω_2 = ⊥;{⟨diverge, Σ_h⟩}`, and a `break` in the prefix still reaches the loop.
- The threading paragraph now lists `(Seq)`, `(Let)` and `(Sub-Never)` among the rules written out over `Ω`, and `(Strict-Bottom)`'s note no longer sends sequence tails to the bottom rules.

**2. Back-edge equality.** Agreed: iteration safety is now validation from the reachable back-edge state, not equality. §5.7 defines a **loop-head state** `head(Σ, e) = Σ_h`: the entry state joined (§5.5) with every reachable back-edge state the body reaches when typed from `Σ_h`, a fixpoint computed by iterating from `Σ`. All three loop rules type their body once at `Σ_h`, and read the back edges `B` and the exits `X` off that typing.
- Your example is accepted: `d` is `MovedOut` at the head, `d = Data { x: 1 }` reinitializes it, and `consume(d)` leaves it `MovedOut` at the back edge again.
- `loop { eat(v0); }` is still rejected: `v0` is `MovedOut` at the head, so the body's `eat(v0)` has nothing to move (E0205).
- A linear outer value moved at a back edge but live at entry has no join, so the loop is ill-formed.

The before/after note now lists the verdict changes honestly: the breakless `eat(v0)` rejection; reassign-then-move accepted again in both loop forms (my first revision had extended the equality to breakless loops, as you said); and one change *away* from the compiler.

Reading the exits at `Σ_h` turned up a compiler soundness bug, **RUE-2354**: the compiler takes the post-loop state from its *first* pass only. A `break` on a later iteration, after an earlier iteration moved an outer binding, is missed:

```rue
loop { if stop(i) { break; } i = i + 1; d = D { x: 10 + i }; eat(d); }
eat(d)   // accepted; the value consumed on iteration 2 is dropped again
```

This prints `1 11 12 12`: a double drop. The calculus now refuses it, per `3.8:80`.

`3.8:79` still literally states the equality. Restating it as the loop-head state is **RUE-2355**, a wording PR for you.

**3. Branch deliveries.** §5.5 now says the delivery set is `Δ_0 ∪ ⋃ᵢ Δ_i`: the condition's or scrutinee's own deliveries, from its `⊣ Σ0;Δ_0`, plus every arm's. So an early `return` inside a still-continuing condition reaches `(Fn)`.

Also updated: the diverge delivery carries the loop-head state (`⟨diverge, Σ_h⟩`), and the traceability row. `validate-doc-links.py` and `validate-lean-xref-index.py` pass.

